Issue #84, opened 2014-07-11
Default -j and -p values are too high on systems with many cores


xqms:
From the documentation:

> By default catkin build will build up to N packages in parallel and pass -jN -lN to make where N is the number of cores in your computer. 

That means that `catkin build` will spawn up to 64 parallel gcc instances on an 8-core machine, which is enough to fill 16GB of RAM and cause the OOM killer to step in.

Is there a chance to enforce a global limit on the number of parallel make jobs? It seems GNU make has an interface for an external job server, maybe we could use that.

Otherwise, please consider lowering the default values for -p and -j. On a 4-core machine 16 processes might be okay, but 64 processes on an 8-core machine is too much.

v01d:
May I ask if catkin is using MAKEFLAGS? I have these defined as -j3 -l3 with my quad-core processor to control load, and it seems whenever I compile with catkin my computer is briefly overloaded. I tried finding the answer to this through all the comments but the discussion is a bit too technical .

xqms:
@v01d Take a look at PR #155 for the most recent solution.

Brief summary: -j M does not do what you expect it to, because `catkin_tools` starts up to N parallel make instances, which in turn spawn M gcc processes. The PR integrates a make job server into `catkin_tools`, which controls the **total** number of compiler processes.

wjwwood:
> May I ask if catkin is using MAKEFLAGS?

It does, here is the function which handles the make arguments: https://github.com/catkin/catkin_tools/blob/0afc6f9c21c933348e48a3b3f2da1da63d2a4865/catkin_tools/argument_parsing.py#L186-L227

jbohren:
@v01d `MAKEFLAGS` are respected by the jobserver in this PR: https://github.com/catkin/catkin_tools/pull/155

jbohren:
This can be closed now that #155 has been merged.

PR #85, opened 2014-07-12
build: add an internal GNU make job server, see #84

xqms:
This is a proof-of-concept for #84.

To test it, try `catkin build --jobserver-limit 4` to limit the number of build tool processes to 4.

You can see the number of currently running make jobs in the lower right corner (`Make: X`).

I know the implementation is pretty ugly, but maybe this is a starting point for discussion.


jbohren:
Rebased in #155

PR #95, opened 2014-09-24
build: Automatically build vanilla CMake packages in workspace

jbohren:
This should enable a user to drop a "vanilla" CMake project into a catkin workspace with the package name the same as the rosdep name. The catkin dependency resolution should then resolve it and build it appropriately in parallel and then install it to the develspace or install space. It's equivalent to putting a placeholder package.xml file into the directory.


jbohren:
Reopening this so it doesn't get lost.

mikepurvis:
Just re-reading this issue— I ended up doing something along these lines with [catkin_tools_python](https://github.com/mikepurvis/catkin_tools_python), including providing a [create_python_package_xmls](https://github.com/mikepurvis/catkin_tools_python/blob/master/catkin_tools_python/create_python_package_xmls.py) tool which spiders the workspace and drops `package.xml` files in where they are missing. So I'm supportive of the idea of an "annotate" tool, especially if other changes (eg, in `catkin_pkg`) can allow the annotations to be supplied out-of-source.

Now, all that said, for our purposes (building big deployment bundles), it turned out to be easier to just "bloom" the Python packages and patch in the required `package.xml` files using that process. This allows them to simply live in our rosdistro, where rosdistro_build_cache picks them up, and rosinstall_generator can drop them into any arbitrary workspace.

g-arjones:
Any updates on this?

A different solution could be adding support for out-of-source package.xml files. catkin_tools would have the concept of a "manifest space" where it would look for package.xml files for third-party non-ros, non-catkin packages (cmake or otherwise). Rosdep would have to be patched to use that as well.

@jbohren @wjwwood Thoughts?

jbohren-hbr:
@g-arjones I think I'm comfortable with build type detection. Are you interested in helping out with such an effort?

jbohren-hbr:
Something else related to build type detection is dependency extraction from the CMakeLists.txt, requirements.txt, or other dependency lists in a given package.

mikepurvis:
Closing this for now. As `catkin_tools` is on life support, I think users who need this functionality might be best served by moving to `colcon`.

jbohren:
> @mikepurvis life support

it had a good run

PR #133, opened 2015-01-16
Fixes linked build-run depend & logic cleanup

### LINKED ISSUE #132: catkin list --depends-on ignores packages with run-only dependancies
The summary says it all. I noticed this mostly because the [baxter_sdk package](https://github.com/RethinkRobotics/baxter/blob/master/baxter_sdk/package.xml) was not being listed as a package that depends on any of the other baxter packages.

To verify, create a new workspace by merging [this rosinstall file](https://raw.githubusercontent.com/RethinkRobotics/baxter/master/baxter_sdk.rosinstall), and then run  `catkin list --depends-on baxter_interface`. You won't see baxter_sdk on that list!

IanTheEngineer:
This PR resolves #132
- Changes depends_on variable to be a python set()
  This makes it clearer and easier to do logic operations
  on list contents
- Adds the logic required to fall into the dependancies
  if-statement if there is only a run dependancy
- Fixes bug where run dependancies were only listed if
  build dependancies also exist
- Minor error message color fix to be consistent with other
  catkin verbs

I have tested this on a few of my workspaces, in particular the one mentioned in #132, and run depends are listed, even if there are no build depends.

### DIFF
--- a/catkin_tools/verbs/catkin_list/cli.py
+++ b/catkin_tools/verbs/catkin_list/cli.py
@@ -33,28 +33,35 @@ def prepare_arguments(parser):
     add('--deps', '--dependencies', default=False, action='store_true',
         help="List dependencies of each package.")
     add('--depends-on', nargs='*',
-        help="One or more dependencies a package must have to be listed.")
+        help="List all packages that depend on supplied argument package(s).")
 
     return parser
 
 
 def main(opts):
     folders = opts.folders or [os.getcwd()]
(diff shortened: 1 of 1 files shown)


wjwwood:
Oops, sorry, I stepped on this a bit. Let me revert.

wjwwood:
Ok, let me review this.

wjwwood:
Cool, looks good to me. Thanks!

IanTheEngineer:
Thanks William!

PR #141, opened 2015-01-28
add status_rate option as we had in rosmake

k-okada:


wjwwood:
Thanks for the pull request, I'll have a look at it as soon as possible.

Also, it looks like there are some failing tests related to PEP8 compliance.

k-okada:
fixed style

--
◉ Kei Okada

catkin_tools/verbs/catkin_build/cli.py
```diff
@@ -215,6 +217,7 @@ def main(opts):
             quiet=not opts.verbose,
             interleave_output=opts.interleave_output,
             no_status=opts.no_status,
+            status_update_rate=opts.status_update_rate,
```
wjwwood: We should ensure that the given status update rate is positive and non-zero.

wjwwood: Also, we should put a limit on it based on the update loop rate, which is currently running at 10Hz:

https://github.com/catkin/catkin_tools/pull/141/files#diff-2482c854781a7d39252bf95c9201c631R521

Because as is, if the user puts `--status-rate=20` it will still only run at 10Hz.

catkin_tools/verbs/catkin_build/build.py
```diff
@@ -606,8 +610,10 @@ def set_error_state(error_state):
                         total_packages
                     ))
                     # Update status bar
-                    wide_log(msg, rhs=msg_rhs, end='\r')
-                    sys.stdout.flush()
+                    if not status_update_rate or time.time() - last_status_update_time > (1.0 / status_update_rate):
```
wjwwood: I would prefer that you check `status_update_rate is not False` since the default value is a boolean whereas the user's provided value is a float. Something like this with a comment describing the condition might be useful:

```python
# if default value (not limited) or enough time has passed, then print a status update
if status_update_rate is False or (time.time() - last_status_update_time) > (1.0 / status_update_rate):
```

catkin_tools/verbs/catkin_build/cli.py
```diff
@@ -105,6 +105,8 @@ def prepare_arguments(parser):
         help='Prevents ordering of command output when multiple commands are running at the same time.')
     add('--no-status', action='store_true', default=False,
         help='Suppresses status line, useful in situations where carriage return is not properly supported.')
+    add('--status-rate', action="store", dest="status_update_rate", default=False, type=float,
+        help='How fast to update the status bar in Hz.  Default: 5Hz.')
```
wjwwood: How is the default of `5Hz` enforced?

k-okada: oh, it just copied from rosmake, Default is 10

catkin_tools/verbs/catkin_build/cli.py
k-okada: check these case in 6ad53d2

catkin_tools/verbs/catkin_build/build.py
k-okada: fixed in c6c7351c0ae05984a1ddf4d8ebad9430fbe38cbb

k-okada:
passed the test

wjwwood:
I want to do the check differently, but I'll merge it manually with the appropriate changes.

wjwwood:
I merged your commits in 70bbe0b1909c11bebb19841612317a618a80d058 and touched them up in 6997625983004c93923d3ed871362264367cd922.

The option now allows you to set 0 to turn off limiting (as fast as normal), and there is no upper limit. I did this because it actually ended up working differently that I had first imagined. I also renamed the option to `--limit-status-rate`, but I left `--status-rate` as a valid alternative flag.

Anyways, give master a spin and see if it is ok with you.

k-okada:
it works find for me! thanks.

--
◉ Kei Okada

PR #147, opened 2015-03-03
build: making it so ephemeral list-type arguments are appended to existing list-type arguments

jbohren:
This allows you to add additional CMake arguments on top of the arguments in a given profile.

@adeguet1 does this do what you needed?

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -71,7 +71,7 @@ class Context(object):
         'space_suffix']
 
     @classmethod
-    def Load(cls, workspace_hint=None, profile=None, opts=None, strict=False):
+    def Load(cls, workspace_hint=None, profile=None, opts=None, strict=False, append=False):
         """Load a context from a given workspace and profile with optional modifications.
 
         This function will try to load a given context from the specified
@@ -93,6 +93,8 @@ class Context(object):
         :type opts: namespace
         :param strict: Causes this function to return None if a workspace isn't found
         :type strict: bool
+        :param append: Appends any list-type opts to existing opts
+        :type append: bool
 
         :returns: A potentially valid Context object constructed from the given arguments
         :rtype: Context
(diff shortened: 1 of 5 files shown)


catkin_tools/context.py
```diff
@@ -126,9 +128,12 @@ def Load(cls, workspace_hint=None, profile=None, opts=None, strict=False):
 
         # User-supplied args are used to update stored args
         # Only update context args with given opts which are not none
-        context_args.update(dict([
-            (k, v) for (k, v) in opts_vars.items()
-            if k in Context.KEYS and v is not None]))
+        for (k,v) in opts_vars.items():
+            if k in Context.KEYS and v is not None:
+                if append and type(context_args.get(k,None)) is list and type(v) is list:
```
wjwwood: nitpick: `k, None`

catkin_tools/context.py
```diff
@@ -71,7 +71,7 @@ class Context(object):
         'space_suffix']
 
     @classmethod
-    def Load(cls, workspace_hint=None, profile=None, opts=None, strict=False):
+    def Load(cls, workspace_hint=None, profile=None, opts=None, strict=False, append=False):
```
wjwwood: nitpick: I guess I missed this on my original review, but methods should always by lowercase: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

jbohren: Even class methods?

catkin_tools/context.py
jbohren: Also `k,v` above.

catkin_tools/context.py
jack-oquin: Yes, even class methods. :smile:

jbohren: I'll change that, then.

jbohren:
@wjwwood I'll update the `Load` / `Save` functions in another PR.

Issue #149, opened 2015-03-06
Whitelist Packages


aghazi:
I am use the whitelist command to select the packages I want to build. Is there something like this for catkin_tools?
``` catkin_make -DCATKIN_WHITELIST_PACKAGES="foo;bar" ```

**Comment by jbohren** [comment]
The simplest equivalent is:

```
catkin build too bar
```

I'm not certain what the behavior of this would be though:

```
catkin config --cmake-args -DCATKIN_WHITELIST_PACKAGES="foo;bar"
catkin build
```

aghazi:
I tried this. It does not work for me. Can I selected packages to be build in another way? Not just once, but whenever catkin build is called after configuring the list once.

jbohren:
@aghazi What do you think of the interface proposed in #152?

jbohren:
I'm closing this now that #152 has been merged. @aghazi feel free to reopen this if you want to keep discussing it.

PR #152, opened 2015-03-06
config/build: adding whitelist/blacklist behavior

jbohren:
`catkin_make` supports "whitelisting" packages with the `CATKIN_WHITELIST_PACKAGES` CMake variable. This PR adds these capabilities to `catkin config` command like so:

*Set packages on the whitelist:*
```
catkin config --whitelist foo bar
```
*Set packages on the blacklist:*
```
catkin config --blacklist foo bar
```

Example in #149, requested by @aghazi

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -63,7 +63,9 @@ class Context(object):
                    'isolate_install',
                    'cmake_args',
                    'make_args',
-                   'catkin_make_args']
+                   'catkin_make_args',
+                   'whitelist',
+                   'blacklist']
 
     KEYS = STORED_KEYS + [
         'workspace',
@@ -161,7 +163,9 @@ class Context(object):
         cmake_args=None,
         make_args=None,
         catkin_make_args=None,
-        space_suffix=None
+        space_suffix=None,
+        whitelist=None,
+        blacklist=None
     ):
         """Creates a new Context object, optionally initializing with parameters
 
@@ -193,6 +197,10 @@ class Context(object):
         :type catkin_make_args: list
         :param space_suffix: suffix for build, devel, and install spaces which are not explicitly set.
         :type space_suffix: str
+        :param whitelist: a list of packages to build by default
+        :type whitelist: list
+        :param blacklist: a list of packages to ignore by default
+        :type blacklist: list
         :raises: ValueError if workspace or source space does not exist
         """
         self.__locked = False
@@ -211,6 +219,10 @@ class Context(object):
         self.devel_space = Context.DEFAULT_DEVEL_SPACE + ss if ss or devel_space is None else devel_space
         self.install_space = Context.DEFAULT_INSTALL_SPACE + ss if ss or install_space is None else install_space
         self.destdir = os.environ['DESTDIR'] if 'DESTDIR' in os.environ else None
+
+        # Handle package whitelist/blacklist
+        self.whitelist = whitelist or []
+        self.blacklist = blacklist or []
 
         # Handle build options
         self.isolate_devel = isolate_devel
@@ -352,6 +364,10 @@ class Context(object):
                 clr("@{cf}Additional Make Args:@|        @{yf}{make_args}@|"),
                 clr("@{cf}Additional catkin Make Args:@| @{yf}{catkin_make_args}@|"),
             ],
+            [
+                clr("@{cf}Whitelisted Packages:@|        @{yf}{whitelisted_packages}@|"),
+                clr("@{cf}Blacklisted Packages:@|        @{yf}{blacklisted_packages}@|"),
+            ]
         ]
 
         # Construct string for extend value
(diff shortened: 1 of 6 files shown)


jbohren:
@wjwwood What do you think of this pattern?

wjwwood:
I'm fine with the approach, but I'm concerned about the visibility of this feature. For example, is the black and white listing expressed in the build summary? I'd have to try it out locally. I'm not going to have time to do that in the next few days. Maybe by the end of the week I can give a spin with some more detailed feedback.

jbohren:
It is shown in the build summary.

NikolausDemmel:
I imagine this to be very powerful in combination with profiles.

+1

jbohren:
@wjwwood The whitelist / blacklist info is printed here: 
https://github.com/jbohren-forks/catkin_tools/blob/feature-blacklist-whitelist/catkin_tools/context.py#L368-369

It's documented here:
https://github.com/jbohren-forks/catkin_tools/commit/df4950796d7b7b0267d1d06da3fc0c3e1d1b12bf#diff-2a8cdb2fb21c1e017c507e90f960ccbbR146

wjwwood:
@jbohren I was talking about the build summary created in this pull request: https://github.com/catkin/catkin_tools/pull/138

Where the build is summarized like this:

jbohren:
@wjwwood Ahh that one. No, it's not shown there. Should be easy to add a `Blacklisted` tag though. If there are whitelisted packages, should all the non-whitelisted ones show up as `Not Whitelisted`?

wjwwood:
Yeah, I don't know how that should look, I was just going to mess around with the output until something looked reasonable, but it will be a few days until I have time to do that.

jbohren:
Better idea: in addition to `Build Summary:` and `Failed Packages:` we could have the following additional categories:

- `Blacklisted Packages`
- `Non-Whitelisted Packages`

### Non-empty blacklist:

Blacklist: `pkg_c`, `pkg_d`, `pkg_e`

```
Build Summary:
 Not Built  pkg_a
 Successful pkg_b
Blacklisted Packages:
 Not Built pkg_c
 Not Built pkg_d
 Not Built pkg_e
Failed Packages:
 Failed    pkg_f
```

### Non-empty whitelist:

Whitelist: `pkg_a`, `pkg_b`

```
Build Summary:
 Not Built  pkg_a
 Successful pkg_b
Non-Whitelisted Packages:
 Not Built pkg_c
 Not Built pkg_d
 Not Built pkg_e
Failed Packages:
 Failed    pkg_f
```

### Non-empty blacklist and whitelist:

Blacklist: `pkg_e`
Whitelist: `pkg_a`, `pkg_b`

```
Build Summary:
 Not Built  pkg_a
 Successful pkg_b
Non-Whitelisted Packages:
 Not Built pkg_c
 Not Built pkg_d
Blacklisted Packages:
 Not Built pkg_e
Failed Packages:
 Failed    pkg_f
```

wjwwood:
That's basically what I was thinking.

davetcoleman:
This is awesome, I will try to test soon

jbohren:
@wjwwood Ok. They're now printed in the build summary.

wjwwood:
Other than the comments, looks good to me.

jbohren:
@wjwwood I think I addressed all of your comments.

wjwwood:
Looks good to me.

hgaiser:
I just pulled the master branch and tried to blacklist some packages but it seems completely ignored. The output `catkin config` does show they are blacklisted, but `catkin build` still tries to build them. In `catkin build` I also see it is noted they are blacklisted in the listing of the profile. I tried clearing all catkin_tools stuff, all build files, but it still occurs.

hgaiser:
Ah, found the issue ... my package that I wanted to blacklist was in a folder called common, so catkin is checking if `'common/package_name'` is in the blacklist, and I had blacklisted just `'package_name'`.

jbohren:
@hgaiser No, you were right first. Thanks for the report. It looks like the whitelist/blacklist was looking at the package path instead of the package name to filter. See #175 for a fix.

hgaiser:
In addition, it would be nice to support at least adding of blacklist / whitelist packages. I'm compiling a workspace and I find some package I can't compile because of library dependencies, I would like to add that to the blacklist.

jbohren:
> In addition, it would be nice to support at least adding of blacklist / whitelist packages to the current list.

Yeah, for both this and the other list-type configure options, it would be good to have a `catkin config --append` option.

hgaiser:
I was thinking to make `--blacklist` or `--whitelist` append by default since there already is a `--clear-blacklist` and `--clear-whitelist`. That way you retain the same functionality as with a `--append` option but with one less option.

jbohren:
@hgaiser Take a look at #179 we can discuss this over there.

PR #155, opened 2015-03-10
build: GNU Make jobserver to prevent resource glut (make_job_server rebased on master)

### LINKED ISSUE #84: Default -j and -p values are too high on systems with many cores

jbohren:
This is a rebase and extension of @xqms's jobserver prototype in PR #85 which was created to fix #84. Without the jobserver, `catkin build` will use too many resources on machines with multiple cores.

Behavior using this PR:
- `catkin build --no-jobserver` -- current behavior (maxes out all jobs for all packages unless using another jobserver like distcc)
- `catkin build` -- build using the jobserver and as many jobs as CPUs
- `catkin build --jobserver` -- explicit version of the line above
- `catkin build -jN`-- build using the jobserver and `N` jobs
- `MAKEFLAGS="-jN" catkin build`-- build using the jobserver and `N` jobs
- `catkin config --no-jobserver` -- disable the jobserver for future builds
- `catkin build -lV` -- don't create more than one job at a time if the system load is greater than `V`
- `catkin build --mem-limit P` -- don't create more than one job at a time if more than `P`% of the system memory is used

Outstanding issues:
- [x] The jobserver behavior isn't documented
- [x] The status line reads `?/0 Jobs` when the jobserver is disabled, it should display something friendlier
- [x] change `internal_make_jobserver` switches `use_internal_make_jobserver`
 - https://github.com/catkin/catkin_tools/pull/155/files#r26704312
 - https://github.com/catkin/catkin_tools/pull/155/files#r26705185
- [x] [document `configure_make_args`](https://github.com/catkin/catkin_tools/pull/155/files#r26703936)
- [x] clean up jobserver API
 - [checking support](https://github.com/catkin/catkin_tools/pull/155/files#r26704509)
- [x] make `psutil` an optional dependency, and only load it if the memory control options are used

Future work:
- The number of jobs (`-jN`) and the number of packages (`-pM`) are still decoupled. The jobserver will limit the number of `Make` instances, but maybe this is ok?
- The current internal jobserver implementation does not monitor load or memory usage. These would be good arguments to support.

### DIFF
--- a/catkin_tools/argument_parsing.py
+++ b/catkin_tools/argument_parsing.py
@@ -212,15 +212,17 @@ def extract_jobs_flags(mflags):
 
     :param mflags: string of space separated make arguments
     :type mflags: str
-    :returns: space separated list of make jobs flags
-    :rtype: str
+    :returns: list of make jobs flags
+    :rtype: list
     """
     regex = r'(?:^|\s)(-?(?:j|l)(?:\s*[0-9]+|\s|$))' + \
             r'|' + \
             r'(?:^|\s)((?:--)?(?:jobs|load-average)(?:(?:=|\s+)[0-9]+|(?:\s|$)))'
     matches = re.findall(regex, mflags) or []
     matches = [m[0] or m[1] for m in matches]
-    return ' '.join([m.strip() for m in matches]) if matches else None
+    filtered_flags = [m.strip() for m in matches] if matches else []
+
+    return filtered_flags
 
 
 def handle_make_arguments(
@@ -253,10 +255,13 @@ def handle_make_arguments(
             wide_log('Forcing "-j1" for running unit tests.')
             jobs_dict['jobs'] = 1
 
-    if 'jobs' in jobs_dict:
-        make_args.append('-j{0}'.format(jobs_dict['jobs']))
-    if 'load-average' in jobs_dict:
-        make_args.append('-l{0}'.format(jobs_dict['load-average']))
+    if len(jobs_dict) == 0:
+        make_args.extend(jobserver_arguments())
+    else:
+        if 'jobs' in jobs_dict:
+            make_args.append('-j{0}'.format(jobs_dict['jobs']))
+        if 'load-average' in jobs_dict:
+            make_args.append('-l{0}'.format(jobs_dict['load-average']))
 
     return make_args
 
@@ -285,17 +290,17 @@ def configure_make_args(make_args, use_internal_make_jobserver):
             'load-average': 1}
 
     # Get MAKEFLAGS from environment
-    makeflags_jobs_flags = extract_jobs_flags(os.environ.get('MAKEFLAGS', '')) or []
+    makeflags_jobs_flags = extract_jobs_flags(os.environ.get('MAKEFLAGS', ''))
     using_makeflags_jobs_flags = len(makeflags_jobs_flags) > 0
-    jobs_flags.update(extract_jobs_flags_values(''.join(makeflags_jobs_flags)))
+    jobs_flags.update(extract_jobs_flags_values(' '.join(makeflags_jobs_flags)))
 
     # Extract make jobs flags (these override MAKEFLAGS)
-    cli_jobs_flags = extract_jobs_flags(''.join(make_args)) or []
+    cli_jobs_flags = extract_jobs_flags(' '.join(make_args))
     using_cli_flags = len(cli_jobs_flags) > 0
-    jobs_flags.update(extract_jobs_flags_values(''.join(cli_jobs_flags)))
+    jobs_flags.update(extract_jobs_flags_values(' '.join(cli_jobs_flags)))
     if cli_jobs_flags:
         # Remove jobs flags from cli args if they're present
-        make_args = re.sub(cli_jobs_flags, '', ' '.join(make_args)).split()
+        make_args = re.sub(' '.join(cli_jobs_flags), '', ' '.join(make_args)).split()
 
     # Instantiate a jobserver
     if use_internal_make_jobserver:
@@ -305,7 +310,7 @@ def configure_make_args(make_args, use_internal_make_jobserver):
 
     # If the jobserver is supported
     if jobserver_supported():
-        jobs_args = jobserver_arguments()
+        jobs_args = []
     else:
         jobs_args = cli_jobs_flags.split()
 
@@ -331,11 +336,10 @@ def argument_preprocessor(args):
     args, cmake_args, make_args, catkin_make_args = extract_make_args(args)
 
     # Extract make jobs flags (these override MAKEFLAGS later on)
-    jobs_flags = extract_jobs_flags(' '.join(args)) or ''
-    jobs_args = jobs_flags.split()
-    if jobs_flags:
+    jobs_args = extract_jobs_flags(' '.join(args))
+    if jobs_args:
         # Remove jobs flags from cli args if they're present
-        args = re.sub(jobs_flags, '', ' '.join(args)).split()
+        args = re.sub(' '.join(jobs_args), '', ' '.join(args)).split()
 
     extras = {
         'cmake_args': cmake_args,
--- a/catkin_tools/argument_parsing.py
+++ b/catkin_tools/argument_parsing.py
@@ -58,8 +58,17 @@ def add_cmake_and_make_and_catkin_make_args(parser):
     """
 
     add = parser.add_argument
-    add('--parallel-jobs', '--parallel', '-p', default=None,
+    add('-p', '--parallel-packages', '--parallel-jobs', '--parallel', dest='parallel_jobs', default=None,
         help='Maximum number of packages which could be built in parallel (default is cpu count)')
+    add('-j', '--jobs', default=None,
+        help='Limit parallel job count through the internal GNU make job server. default is cpu count')
+
+    add = parser.add_mutually_exclusive_group().add_argument
+    add('--jobserver', dest='internal_make_jobserver', default=None, action='store_true',
+        help='Use the internal GNU Make job server which will limit the number '
+             'of Make jobs across all active packages.')
+    add('--no-jobserver', dest='internal_make_jobserver', default=None, action='store_false',
+        help='Disable the internal GNU Make job server, and use an external one (like distcc, for example).')
 
     add = parser.add_mutually_exclusive_group().add_argument
     add('--cmake-args', metavar='ARG', dest='cmake_args', nargs='+', required=False, type=str, default=None,
@@ -169,6 +178,33 @@ def extract_cmake_and_make_arguments(args):
     return args, cmake_args, make_args
 
 
+def extract_jobs_flags_values(mflags):
+    """Gets the values of tha make jobs flags
+
+    :param mflags: string of space separated make arguments
+    :type mflags: str
+    :returns: dictionary mapping jobs flags to jobs flags values
+    :rtype: dict
+    """
+
+    regex = r'(?:^|\s)(?:-?(j|l)(\s*[0-9]+|\s|$))' + \
+            r'|' + \
+            r'(?:^|\s)(?:(?:--)?(jobs|load-average)(?:(?:=|\s+)([0-9]+)|(?:\s|$)))'
+
+    jobs_dict = {}
+
+    matches = re.findall(regex, mflags) or []
+    for k, v, key, value in matches:
+        v = v.strip()
+        value = value.strip()
+        if k == 'j' or key == 'jobs':
+            jobs_dict['jobs'] = int(v or value) if (v or value) else ''
+        elif k == 'l' or key == 'load-average':
+            jobs_dict['load-average'] = float(v or value)
+
+    return jobs_dict
+
+
 def extract_jobs_flags(mflags):
     """Extracts make job flags from a list of other make flags, i.e. -j8 -l8
 
@@ -185,7 +221,9 @@ def extract_jobs_flags(mflags):
     return ' '.join([m.strip() for m in matches]) if matches else None
 
 
-def handle_make_arguments(input_make_args, force_single_threaded_when_running_tests=False):
+def handle_make_arguments(
+        input_make_args,
+        force_single_threaded_when_running_tests=False):
     """Special handling for make arguments.
 
     If force_single_threaded_when_running_tests is True, jobs flags are
@@ -203,34 +241,73 @@ def handle_make_arguments(input_make_args, force_single_threaded_when_running_te
     """
     make_args = list(input_make_args)
 
+    # Get the values for the jobs flags which may be in the make args
+    jobs_dict = extract_jobs_flags_values(' '.join(make_args))
+
     if force_single_threaded_when_running_tests:
         # force single threaded execution when running test since rostest does not support multiple parallel runs
         run_tests = [a for a in make_args if a.startswith('run_tests')]
         if run_tests:
             wide_log('Forcing "-j1" for running unit tests.')
-            make_args.append('-j1')
+            jobs_dict['jobs'] = 1
 
-    jobserver = MakeJobServer.get_instance()
+    if 'jobs' in jobs_dict:
+        make_args.append('-j{0}'.format(jobs_dict['jobs']))
+    if 'load-average' in jobs_dict:
+        make_args.append('-l{0}'.format(jobs_dict['load-average']))
 
-    # If no -j/--jobs/-l/--load-average flags are in make_args
-    if not extract_jobs_flags(' '.join(make_args)):
-        # If -j/--jobs/-l/--load-average are in MAKEFLAGS
-        if 'MAKEFLAGS' in os.environ and extract_jobs_flags(os.environ['MAKEFLAGS']):
-            # Do not extend make arguments, let MAKEFLAGS set things
-            pass
-        elif jobserver.supported:
-            make_args += jobserver.make_arguments()
-        else:
-            # Else extend the make_arguments to include some jobs flags
-            # Use the number of CPU cores
-            try:
-                jobs = cpu_count()
-                make_args.append('-j{0}'.format(jobs))
-                make_args.append('-l{0}'.format(jobs))
-            except NotImplementedError:
-                # If the number of cores cannot be determined, do not extend args
-                pass
     return make_args
+
+
+def configure_make_args(make_args, internal_make_jobserver):
+    """Initialize the internal GNU Make jobserver or configure it as a pass-through
+
+    :param make_args: arguments to be passed to GNU Make
+    :type make_args: list
+    :param internal_make_jobserver: if true, use the internal jobserver
+    :type make_args: bool
+    :rtype: tuple
+    """
+
+    # Configure default jobs options: use all CPUs in each package
+    try:
+        # NOTE: this will yeild greater than 100% CPU utilization
+        n_cpus = cpu_count()
+        jobs_flags = {
+            'jobs': n_cpus,
+            'load-average': n_cpus}
+    except NotImplementedError:
+        # If the number of cores cannot be determined, limit to one job
+        jobs_flags = {
+            'jobs': 1,
+            'load-average': 1}
+
+    # Get MAKEFLAGS from environment
+    makeflags_jobs_flags = extract_jobs_flags(os.environ.get('MAKEFLAGS', '')) or []
+    using_makeflags_jobs_flags = len(makeflags_jobs_flags) > 0
+    jobs_flags.update(extract_jobs_flags_values(''.join(makeflags_jobs_flags)))
+
+    # Extract make jobs flags (these override MAKEFLAGS)
+    cli_jobs_flags = extract_jobs_flags(''.join(make_args)) or []
+    using_cli_flags = len(cli_jobs_flags) > 0
+    jobs_flags.update(extract_jobs_flags_values(''.join(cli_jobs_flags)))
+    if cli_jobs_flags:
+        # Remove jobs flags from cli args if they're present
+        make_args = re.sub(cli_jobs_flags, '', ' '.join(make_args)).split()
+
+    # Instantiate a jobserver
+    jobserver = MakeJobServer.get_instance(
+        num_jobs=jobs_flags.get('jobs', None),
+        enable=internal_make_jobserver)
+
+    # If the jobserver is supported and enabled, override the jobs flags
+    using_jobserver = jobserver.supported and internal_make_jobserver
+    if using_jobserver:
+        jobs_args = jobserver.make_arguments()
+    else:
+        jobs_args = cli_jobs_flags.split()
+
+    return make_args + jobs_args, using_makeflags_jobs_flags, using_cli_flags, using_jobserver
 
 
 def argument_preprocessor(args):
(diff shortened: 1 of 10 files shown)


jbohren:
@xqms I've rebased this on master, it appears to work, and feel free to copy this snapshot to your fork.

jbohren:
@davetcoleman when you're feeling bored, give this branch a shot and see if you still run out of resources

jbohren:
Rebase of #85

davetcoleman:
I just pulled that branch and am testing now.

davetcoleman:
It kept my memory usage nice and stable - think whatever this is helped a lot!
screenshot from 2015-03-09 22 57 56

jbohren:
@xqms I modified the implementation so that it's possible to disable the jobserver with the `--ext-jobserver` argument. This way you can still use `distcc` like so:

```
CC="distcc gcc" CXX="distcc g++" catkin build -p$(distcc -j) -j$(distcc -j) --ext-jobserver
```

xqms:
@davetcoleman Nice to hear it helped you as well! `catkin_tools` is almost unusable without it on my workspace...

@jbohren okay, sounds good, though I think @wjwwood preferred an opt-in (like `--enable-jobserver`). I'll have time later today or tomorrow to take a look. If you guys want to go ahead changing/merging, don't wait for me ;-)

jbohren:
@xqms @wjwwood after seeing Dave's example workspace, I think that this should be the default behavior.

jbohren:
@xqms also is there a difference between `--jobserver-limit` and the normal `-j` jobs argument?

jbohren:
Something that might make sense is to add the jobserver options to the persistent config. This patch just augments the `--make-args` with the jobserver flags like `--jobserver-fds=3,4 -j` which makes it really visible, but also obfuscated.

Alternatively, it'd be better to have a high-level `Job Server: [internal / external]` option.

This would display like:

```
----------------------------------------------------------------------------
Extending:          [cached] /opt/ros/hydro
Workspace:                   /home/jbohren/ws/ascent
Source Space:       [exists] /home/jbohren/ws/ascent/src
Build Space:        [exists] /home/jbohren/ws/ascent/build
Devel Space:        [exists] /home/jbohren/ws/ascent/devel
Install Space:     [missing] /home/jbohren/ws/ascent/install
DESTDIR:                     None
----------------------------------------------------------------------------
Isolate Develspaces:         False
Install Packages:            False
Isolate Installs:            False
----------------------------------------------------------------------------
Additional CMake Args:       -DOBJREC_USE_CUDA=On -DOBJREC_CUDA_DEVICE_ID=1 
Make Job Server:             Internal
----------------------------------------------------------------------------
Workspace configuration appears valid.
----------------------------------------------------------------------------
```

xqms:
@jbohren `-j N` controls how many processes a `make` process may spawn. Since `catkin_tools` starts up to `-p X` parallel `make` processes this is useless for controlling the system load. `--jobserver-limit` specifies how many jobs (e.g. gcc processes) may be run in **total**, which is much more useful.

If the jobserver is enabled, it may make sense to replace `-j` and `-p` altogether with a single `-j` option controlling the jobserver limit...

Yes, I agree this thing should be integrated with the workspace config.

davetcoleman:
This is still running ok for me and catkin no longer crashes my computer, but it still uses a huge amount of memory sometimes while compiling a large workspace. Just now it got pretty close:

screenshot from 2015-03-11 16 34 37

jbohren:
Can you show the memory trace using catkin_make also?

wjwwood:
Thanks for everyone helping out on this one. I'll try to review it and try it out myself by the end of the week.

jack-oquin:
> If the jobserver is enabled, it may make sense to replace ``-j`` and ``-p`` altogether with a single ``-j`` option controlling the jobserver limit...

+1 That is what most of us naively *expected* ``-j`` to mean in the first place.

davetcoleman:
You mean the memory history? I'll try to record it next time I rebuild

jbohren:
Yeah.

jbohren:
This still needs a bit of work to do what people "expect" it to do. Each job calls `handle_make_arguments` a couple times and I don't think that's where `MAKEFLAGS` should be read. It needs to be read early on and act as a default configuration. Then that needs to initialize the jobserver.

jbohren:
@xqms @davetcoleman @wjwwood This should be good to review. I re-worked it to make sure the jobserver singleton gets initialized in one place, but this could still be cleaned up.

The behavior of this PR is as follows:
- `catkin build --no-jobserver` -- current behavior (maxes out all jobs for all packages unless using another jobserver like distcc)
- `catkin build` -- build using the jobserver and as many jobs as CPUs
- `catkin build --jobserver` -- explicit version of the line above
- `catkin build -jN`-- build using the jobserver and `N` jobs
- `MAKEFLAGS="-jN" catkin build`-- build using the jobserver and `N` jobs

catkin_tools/argument_parsing.py
```diff
@@ -167,6 +183,33 @@ def extract_cmake_and_make_arguments(args):
...
+def extract_jobs_flags_values(mflags):
+    """Gets the values of tha make jobs flags
+
+    :param mflags: string of space separated make arguments
+    :type mflags: str
+    :returns: dictionary mapping jobs flags to jobs flags values
+    :rtype: dict
+    """
+
+    regex = r'(?:^|\s)(?:-?(j|l)(\s*[0-9]+|\s|$))' + \
```
xqms: nit: you probably wanted to use JOBS_FLAGS_REGEX from above here, right?

xqms:
@jbohren I had a look at the code, seems good to me. I'll test this in my setup today and let you know if anything breaks.

catkin_tools/argument_parsing.py
jbohren: Ah yeah I meant to remove that in favor of just having two different patterns.

xqms:
I've been running this for 2 days now with high compilation workloads in my setup, everything works as expected.

catkin_tools/argument_parsing.py
wjwwood: Is this comment still unaddressed?

wjwwood:
lgtm, I'll use this while building up Indigo/Jade on my new laptop to try it out.

catkin_tools/argument_parsing.py
```diff
@@ -201,32 +241,75 @@ def handle_make_arguments(input_make_args, force_single_threaded_when_running_te
...
 
+def configure_make_args(make_args, internal_make_jobserver):
+    """Initialize the internal GNU Make jobserver or configure it as a pass-through
+
+    :param make_args: arguments to be passed to GNU Make
+    :type make_args: list
+    :param internal_make_jobserver: if true, use the internal jobserver
+    :type make_args: bool
+    :rtype: tuple
+    """
```
wjwwood: The return tuple of this function is complicated, I think it would be really helpful for others if what is in it is spelled out in the `:returns:` directive or in the body of the doc string.

catkin_tools/argument_parsing.py
```diff
@@ -201,32 +241,75 @@ def handle_make_arguments(input_make_args, force_single_threaded_when_running_te
...
+
     return make_args
 
 
+def configure_make_args(make_args, internal_make_jobserver):
+    """Initialize the internal GNU Make jobserver or configure it as a pass-through
+
+    :param make_args: arguments to be passed to GNU Make
+    :type make_args: list
+    :param internal_make_jobserver: if true, use the internal jobserver
```
wjwwood: `use_internal_make_jobserver` would be a more clear variable name.

catkin_tools/argument_parsing.py
```diff
@@ -201,32 +241,75 @@ def handle_make_arguments(input_make_args, force_single_threaded_when_running_te
...
+        # Remove jobs flags from cli args if they're present
+        make_args = re.sub(cli_jobs_flags, '', ' '.join(make_args)).split()
+
+    # Instantiate a jobserver
+    jobserver = MakeJobServer.get_instance(
+        jobs_flags.get('jobs', None),
+        internal_make_jobserver)
+
+    # If the jobserver is supported and enabled, override the jobs flags
+    using_jobserver = jobserver.supported and internal_make_jobserver
```
wjwwood: It's weird that you pass `internal_make_jobserver` to `get_instance` (and therefore the constructor of `MakeJobServer`), only to check `supported` AND `internal_make_jobserver` here. Is it not just sufficient to check `supported` here?

catkin_tools/context.py
```diff
@@ -194,6 +196,8 @@ def __init__(
         :type cmake_args: list
         :param make_args: extra make arguments to be passed to make for each package
         :type make_args: list
+        :param internal_make_jobserver: true if this configuration should use an internal make jobserv
+        :type internal_make_jobserver: bool
```
wjwwood: Again, throughout the code, `use_internal_make_jobsever` or something similar that indicates this is an option and not the actually job sever would be better.

catkin_tools/make_jobserver.py
```diff
@@ -0,0 +1,151 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import print_function
+
+from os import pipe, write, close, unlink, read
```
wjwwood: Please do not do multiple imports on one line.

catkin_tools/make_jobserver.py
```diff
@@ -0,0 +1,151 @@
...
+from multiprocessing import cpu_count
+from tempfile import mkstemp
+from subprocess import call, PIPE
```
wjwwood: Same here.

catkin_tools/make_jobserver.py
```diff
@@ -0,0 +1,151 @@
...
+'''
+
+
+class MakeJobServer:
+
+    """
+    This class implements a GNU make job server.
+
+    TODO:
+     - use os.getloadavg() to maintain server load
```
wjwwood: Is this TODO still relevant?

catkin_tools/make_jobserver.py
```diff
@@ -0,0 +1,151 @@
...
+
+        while True:
+            try:
+                token = read(self.pipe[0], 1)
+                return token
+            except OSError as e:
+                if e.errno == errno.EINTR:
+                    continue
+                else:
+                    raise
```
wjwwood: Simplify to:

```python
if e.errno != errno.EINTR:
    raise
```

catkin_tools/verbs/catkin_build/build.py
```diff
@@ -732,12 +733,24 @@ def set_error_state(error_state):
...
                     for job_msg_args in sorted(executing_jobs, key=lambda args: args['number']):
                         msg += clr("[{name} - {run_time}] ").format(**job_msg_args)
-                    msg_rhs = clr("[{0}/{1} Active | {2}/{3} Completed]").format(
-                        len(executing_jobs),
-                        len(executors),
-                        len(packages) if no_deps else len(completed_packages),
-                        total_packages
-                    )
+
+                    if MakeJobServer.get_instance().supported:
```
wjwwood: I guess you're assuming here that `get_instance()` was already called, otherwise it will call the constructor with no args. Could that be a problem? Should there be a function to get and instance only if it already exists, otherwise fail?

catkin_tools/verbs/catkin_build/build.py
```diff
@@ -732,12 +733,24 @@ def set_error_state(error_state):
...
+                        msg_rhs = clr("[{0}/{1} Jobs][{2}/{3} Packages][{4}/{5} Completed]").format(
+                            MakeJobServer.get_instance().num_running_jobs(),
+                            MakeJobServer.get_instance().num_jobs,
```
wjwwood: Is there any reason to continue to get the instance? Are you expecting it to asynchronously change? If not it would be better to store it and reuse it to reduce function calls (which do cost in Python).

catkin_tools/argument_parsing.py
jbohren: No, it's outdated. The referenced variable isn't in the PR any more.

catkin_tools/argument_parsing.py
jbohren: Agreed.

catkin_tools/argument_parsing.py
jbohren: Yep.

catkin_tools/argument_parsing.py
jbohren: Yeah, there's no reason to check both, this is just slop. I'd actually like it if `supported` didn't keep having to be checked.

catkin_tools/make_jobserver.py
jbohren: Yeah. Right now the jobserver only cares about the number of jobs, and not the system load average. It'd be nice if it could consider also the server load and the RAM footprint.

wjwwood: Ok, let's just remember to open an issue to track this if we merge this without addressing that.

jbohren: Yeah it's already mentioned in the "Future Work" section at the top of the PR.

catkin_tools/verbs/catkin_build/build.py
```diff
@@ -732,12 +733,24 @@ def set_error_state(error_state):
...
                         msg += clr("[{name} - {run_time}] ").format(**job_msg_args)
-                    msg_rhs = clr("[{0}/{1} Active | {2}/{3} Completed]").format(
-                        len(executing_jobs),
-                        len(executors),
-                        len(packages) if no_deps else len(completed_packages),
-                        total_packages
-                    )
+
+                    if MakeJobServer.get_instance().supported:
+                        msg_rhs = clr("[{0}/{1} Jobs][{2}/{3} Packages][{4}/{5} Completed]").format(
```
wjwwood: This breaks the colorization of the part of the status message. You'll need to provide a matching entry in this file:

https://github.com/catkin/catkin_tools/blob/0afc6f9c21c933348e48a3b3f2da1da63d2a4865/catkin_tools/verbs/catkin_build/color.py#L57-L58

jbohren:
@wjwwood I've cleaned up the API a bunch based on the discussion before

I also added the following
- basic load average monitoring
- an experimental memory usage monitoring

All rebased on latest master.

catkin_tools/make_jobserver.py
```diff
@@ -0,0 +1,233 @@
...
+        """
+        Obtain a job server token. Be sure to call release() to avoid
+        deadlocks.
+        """
+
+        while True:
+            # make sure we're observing load maximums
+            if self.max_load is not None:
+                try:
+                    max_load = 8.0
```
NikolausDemmel: This line is probably left over by accident?

catkin_tools/make_jobserver.py
```diff
@@ -0,0 +1,233 @@
...
+                except NotImplementedError:
+                    pass
+
+            # make sure we're observing memory maximum
+            if self.max_mem is not None:
+                mem_usage = psutil.phymem_usage()
+                if jobserver_running_jobs() > 0 and mem_usage.percent > self.max_mem:
+                    time.sleep(0.01)
+                    continue
+
```
NikolausDemmel: If both `self.max_load` and `self.max_mem` are set, you probably want to wait for both conditions in a single loop, right?

jbohren: Yeah, it's already in a single loop.

NikolausDemmel: My bad. My parsing was off. Sorry for the noise.

xqms:
I have just discovered an issue: Multiple invocations of `catkin build --save` duplicate the `--jobserver-fds=3,4 -j` arguments. Looks like this:

```
----------------------------------------------
Additional CMake Args:       -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_FLAGS=-Wall
Additional Make Args:        --jobserver-fds=3,4 -j --jobserver-fds=3,5 -j
----------------------------------------------
```

xqms:
Okay, to clarify: the `--jobserver-fds` argument is saved in the workspace config. It shouldn't be, since it is also generated by the "Internal Make Job Server" flag.

jbohren:
@xqms Yeah, also now that `Internal Make Job Server:    True` is in the summary, I don't think the jobserver flags should be added in the way that I have them added. I'll make it so it displays the summary like this:

```
----------------------------------------------
Additional CMake Args:       -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_FLAGS=-Wall
Internal Make Job Server:    True, Args: "--jobserver-fds=3,4 -j"
----------------------------------------------
```

wjwwood:
https://travis-ci.org/catkin/catkin_tools/jobs/56253398#L1432

jbohren:
@wjwwood Just updated. There was an issue previously where the make arg parsing wasn't properly extracting jobs args formatted like `-j N` with a space between the `-j` and the `N`.

@xqms This removes the jobserver args from "additional make args". I was considering adding the jobserver max jobs / max load / max mem to the context, but we'd also have to add those options to the saved configs, which just involves a lot of additional changes.

As it stands, I think that this is a pretty critical patch and if there are any other outstanding features, we should bump them to another PR.

It would be nice to add a strong test harness around this thing, but I'm not sure what the best approach would be for that. Mostly you'd want to test:

 - defaults
 - cli args
 - MAKEFLAGS args
 - disabling the jobserver

davetcoleman:
I've been using, and keeping up with the updates, of this branch everyday. Seems to work for me.

wjwwood:
You've not added the dependency on `psutil` correctly, you also have to update some other files:

- https://github.com/catkin/catkin_tools/blob/master/setup.py
- https://github.com/catkin/catkin_tools/blob/master/stdeb.cfg

Also the function you're using from `psutil` is deprecated:

https://code.google.com/p/psutil/wiki/Documentation#Memory

Please at least update the pull request to use `psutil.virtual_memory()`.

I'm not happy about adding a dependency on `psutil`, but if there isn't a reasonable way to do it without `psutil` then I guess we'll have to. One option would be to not import psutil unless a user tries to use the memory limits.

Having said all of that, I'm really not convinced the memory limits is going to be a reliable way to manage how much memory the whole compile is using. Unlike CPU usage, the memory usage of each `cc` in a make job can grow over time. Imagine:

- each of 6 `cc` process will approach 2 GB of memory usage over 30 seconds (this behavior is common with some of the `pcl` object files)
- your system has 4 GB of RAM
- you compiled with -j6
- and you've set the memory limit to 90%

You'll hand out all six job tokens (because for the first few seconds the memory usage is really low) and each of them will grow to use 2 GB and run your system out of memory anyways. However, it's possible that it won't work out like that and the memory limit will help, but I wouldn't call it reliable.

This is a tough problem and we worked with the Tango project to work around this problem in a few places. The most reliable way to fix this kind of "perfect storm of parallel make jobs" was to instrument the build so we could figure out which process, when run at the same time, busted the system. Then they would manually adjust the inter-target dependencies to prevent those object files from being built at the same time. @tfoote was the instrumenting code ever pushed somewhere public?

Based on that, I would say we should either remove the memory limit option or make it "experimental" and make the dependency on `psutil` optional. If a user is savvy enough to make use of the memory limit options then they are probably also ok with installing `psutil` when prompted.

jbohren:
> Having said all of that, I'm really not convinced the memory limits is going to be a reliable way to manage how much memory the whole compile is using. Unlike CPU usage, the memory usage of each cc in a make job can grow over time. Imagine:

Of course it's not going to be reliable. But if some crazy person (@davetcoleman *cough* *cough*) is running around with a computer with no swap, he can probably set it to something like 65% and be reasonably confident it's not going to brick his machine.  

> This is a tough problem and we worked with the Tango project to work around this problem in a few places. The most reliable way to fix this kind of "perfect storm of parallel make jobs" was to instrument the build so we could figure out which process, when run at the same time, busted the system. Then they would manually adjust the inter-target dependencies to prevent those object files from being built at the same time. @tfoote was the instrumenting code ever pushed somewhere public?

"Perfect is the enemy of the good" -- *some guy that got something done*

I'm not concerned enough about this problem to do any more about it than is in the experimental memory limit check.

> Based on that, I would say we should either remove the memory limit option or make it "experimental" and make the dependency on psutil optional. If a user is savvy enough to make use of the memory limit options then they are probably also ok with installing psutil when prompted.

It already is an unlisted "experimental" option. I'm happy putting a try-except around the psutil import and then throwing an error if someone tries using it without support. That being said, anyone with `rqt` installed will have `python-psutil`.

> You've not added the dependency on psutil correctly, you also have to update some other files:

> - https://github.com/catkin/catkin_tools/blob/master/setup.py
> - https://github.com/catkin/catkin_tools/blob/master/stdeb.cfg

I'm going to remove the dependency description completely (except for the travis config) because it's going to be an optional dependency.

>Also the function you're using from psutil is deprecated:

> https://code.google.com/p/psutil/wiki/Documentation#Memory

> Please at least update the pull request to use psutil.virtual_memory().

Yeah I know it's deprecated. Unfortunately the `python-psutil` package on 12.04 doesn't have this function yet. Also ROS Hydro depends on the `python-psutil` package, so ditching it in favor of the pypi doesn't sound like a good plan. It wouldn't be too hard to just check if the function is available and use the new one if the old one isn't available.

wjwwood:
> It already is an unlisted "experimental" option. I'm happy putting a try-except around the psutil import and then throwing an error if someone tries using it without support. That being said, anyone with rqt installed will have python-psutil.

I would appreciate making it optional. It's not going to be a problem for ROS users, but I put a lot of effort into minimizing the dependencies, so I'd like to keep it that way if possible.

> Yeah I know it's deprecated. Unfortunately the python-psutil package on 12.04 doesn't have this function yet. Also ROS Hydro depends on the python-psutil package, so ditching it in favor of the pypi doesn't sound like a good plan. It wouldn't be too hard to just check if the function is available and use the new one if the old one isn't available.

I remember this now, it was quite a problem for `rqt_top`. Unfortunately the `pypi` version is never an option because we cannot build a `.deb` that depends on something in `pypi`.

If you can get the optional import in place tonight or tomorrow morning I can merge this and get it in the next release. Tomorrow is our ROS bug fix party, so I plan to spend some time on this tracker and making a release.

Thanks for working on it and iterating with me.

tfoote:
@wjwwood re: collecting build parameters of specific invocations. The tools is available at https://github.com/osrf/watchprocess  It can generate a report of cpu and memory usage for each process using an instrumented program during a build.

davetcoleman:
Thanks for the crazy person shout out @jbohren I've taken your advice:
screenshot from 2015-04-02 12 07 20

jbohren:
@wjwwood I added the lazy import of psutil and the version checking. I've tested it with version 0.4.1 but no other versions.

wjwwood:
I merged this with some adjustments.

wjwwood:
Also, a rebase which is why it doesn't show up as "merged".

Issue #90, opened 2014-08-19
meta: things todo before recommending a switch from cmi


wjwwood:
This is a list of things we need to get done before we make the official switch to `catkin build` and move away from using `catkin_make_isolated`.
#### bugs
- [x] Installed libraries reference devel space (https://github.com/catkin/catkin_tools/issues/70)
  - [x] Related: https://github.com/catkin/catkin_tools/issues/69
- [x] catkin: should not fail if cmake is not available (https://github.com/catkin/catkin_tools/issues/61)
- [x] leaf_sources out of bounds error (https://github.com/catkin/catkin_tools/issues/45)
#### missing functionality or features
- [x] build: support for arch-specific library folders (https://github.com/catkin/catkin_tools/issues/154)
- [x] build: test for compatibility with DESTDIR (~~https://github.com/catkin/catkin_tools/issues/31~~) (closed by https://github.com/catkin/catkin_tools/pull/240)
- [ ] build: test for topological ordering with underlays (https://github.com/catkin/catkin_tools/issues/30)
- [ ] build: speed up setup.sh for non-merged devel spaces (https://github.com/catkin/catkin_tools/issues/28)
  - [x] This issue also applies to this in general: https://github.com/catkin/catkin_tools/issues/74
  - [x] Also this: https://github.com/catkin/catkin_tools/issues/23
- [x] Bash/Zsh completion (https://github.com/catkin/catkin_tools/issues/15)
- [x] Detect warnings like rosmake (https://github.com/catkin/catkin_tools/issues/1)
- [x] build: support invocation from within symlink folders (#91) (ros/catkin#638)

dirk-thomas:
Since it has been announced on *ros-users* and likely more users will start using it / looking into it I think the still missing features compared to `catkin_make(_isolated)` should be listed clearly in the documentation.

jbohren:
Yeah, I think that would be a helpful page to have. Dirk, do you have any time to start it?

dirk-thomas:
I don't think I am aware enough about the progress of `catkin_tools` to know what is supported right now and what isn't. Basically because I am rarely using the took.

I also don't have the time to check if the tool actually supports the features e.g. listed at the top of this ticket, sorry. Others might know this out of their heads?

PR #156, opened 2015-03-10
Add arch-specific library folders (#154)

### LINKED ISSUE #154: Catkin tools don't handle arch-specific library folders
Support for arch-specific library folders was added to catkin in https://github.com/ros/catkin/pull/624 but it does not seem to be present in catkin tools. This is needed to build gazebo with catkin tools.

scpeters:
Try to detect arch-specific library folders using
gcc and dpkg-architecture, then add to
`LD_LIBRARY_PATH` and `PKG_CONFIG_PATH`.
Closes #154.
Based on https://github.com/ros/catkin/pull/624

### DIFF
--- a/catkin_tools/verbs/catkin_build/job.py
+++ b/catkin_tools/verbs/catkin_build/job.py
@@ -15,6 +15,8 @@
 from __future__ import print_function
 
 import os
+import subprocess
+import sys
 import tempfile
 
 from catkin_tools.argument_parsing import handle_make_arguments
@@ -139,6 +141,32 @@ class CMakeJob(Job):
         Job.__init__(self, package, package_path, context, force_cmake)
         self.commands = self.get_commands()
 
+    def get_multiarch(self):
+        if not sys.platform.lower().startswith('linux'):
+            return ''
+        # this function returns the suffix for lib directories on supported systems or an empty string
+        # it uses two step approach to look for multiarch: first run gcc -print-multiarch and if
+        # failed try to run dpkg-architecture
+        error_thrown = False
+        try:
+            p = subprocess.Popen(
+                ['gcc', '-print-multiarch'],
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            out, err = p.communicate()
+        except (OSError, FileNotFoundError):
+            error_thrown = True
+        if error_thrown or p.returncode != 0:
+            try:
+                out, err = subprocess.Popen(
+                    ['dpkg-architecture', '-qDEB_HOST_MULTIARCH'],
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE).communicate()
+            except (OSError, FileNotFoundError):
+                return ''
+        # be sure to return empty string or a valid multiarch tuple
+        decoded = out.decode().strip()
+        assert(not decoded or decoded.count('-') == 2)
+        return decoded
+
     def get_commands(self):
         commands = []
         # Setup build variables
@@ -189,14 +217,17 @@ class CMakeJob(Job):
                 # Do not replace existing setup.sh if devel space is merged
                 return commands
         # Create the setup file other packages will source when depending on this package
+        arch = self.get_multiarch()
         subs = {}
         subs['cmake_prefix_path'] = install_target + ":"
         subs['ld_path'] = os.path.join(install_target, 'lib') + ":"
         pythonpath = os.path.join(install_target, get_python_install_dir())
         subs['pythonpath'] = pythonpath + ':'
-        subs['pkgcfg_path'] = os.path.join(install_target, 'lib', 'pkgconfig')
-        subs['pkgcfg_path'] += ":"
+        subs['pkgcfg_path'] = os.path.join(install_target, 'lib', 'pkgconfig') + ":"
         subs['path'] = os.path.join(install_target, 'bin') + ":"
+        if arch:
+            subs['ld_path'] += os.path.join(install_target, 'lib', arch) + ":"
+            subs['pkgcfg_path'] += os.path.join(install_target, 'lib', arch, 'pkgconfig') + ":"
         setup_file_directory = os.path.dirname(setup_file_path)
         if not os.path.exists(setup_file_directory):
             os.makedirs(setup_file_directory)


scpeters:
Which package do I need to install to run nosetests?

wjwwood:
`sudo apt-get install python-nose` on Ubuntu or `sudo -H pip install -U nose` on OS X.

scpeters:
Ok, I have that installed, but when I do `python setup.py nosetests`, it says `error: invalid command 'nosetests'`

scpeters:
That command was recommended in #90

scpeters:
oops indentation...

wjwwood:
What operating system?

scpeters:
Ubuntu trusty

wjwwood:
I can run `python setup.py nosetests` in the root of the `catkin_tools` directory on my VM of Trusty and on my Mac. I'm not sure why it doesn't work for you. Make sure you've installed nose for the same instance of python that you're using.

scpeters:
It was a venv problem; it's working now.

scpeters:
With python 2.7 on trusty, starting in catkin_tools source repo root:
~~~
$ virtualenv $HOME/local
$ . ~/local/bin/activate
$ python setup.py install
$ pip install argparse catkin-pkg distribute PyYAML
$ pip install nose coverage flake8 --upgrade
$ pip install mock
$ python setup.py nosetests -s
~~~
That mostly works, but some tests complain about not being able to find catkin. There's a few steps in the `.travis.yml`I skipped.

wjwwood:
Yeah, you'll need to install catkin and source the setup file for it.

scpeters:
python3 failure: any advice?
~~~
  File "/home/travis/build/catkin/catkin_tools/catkin_tools/verbs/catkin_build/job.py", line 159, in get_multiarch
    assert(not out.strip() or out.strip().count('-') == 2)
TypeError: Type str doesn't support the buffer API
~~~

wjwwood:
That error happens because the contents of `out` (I guess returned from calling out to the system) is a "bytes" array which is different from a `str` in Python3. You can decode `out` before you start doing normal operations on it if you like. Something like this:

```
>>> import subprocess
>>> out = subprocess.check_output('ls')
>>> out
b'foo\nbar'
>>> out.count('-')
  File "<stdin>", line 1, in <module>
TypeError: Type str doesn't support the buffer API
>>> out = out.decode()
>>> out
'foo\nbar'
>>> out.count('-')
1
```

scpeters:
I don't need to declare an explicit type for decode?

scpeters:
I did that in 3833087 after seeing stack overflow

wjwwood:
It defaults to `utf-8`.

scpeters:
I'll squash after code review unless otherwise directed.

catkin_tools/verbs/catkin_build/job.py
```diff
@@ -139,6 +141,25 @@ def __init__(self, package, package_path, context, force_cmake):
         Job.__init__(self, package, package_path, context, force_cmake)
         self.commands = self.get_commands()
 
+    def get_multiarch(self):
+        if not sys.platform.lower().startswith('linux'):
+            return ''
+        # this function returns the suffix for lib directories on supported systems or an empty string
+        # it uses two step approach to look for multiarch: first run gcc -print-multiarch and if
+        # failed try to run dpkg-architecture
+        p = subprocess.Popen(
```
wjwwood: These calls to `subprocess.Popen` can raise a `OSError` or a `FileNotFoundError` if the executable you gave does not exist. You may need to catch that. You can test it by changing `gcc` to `gcc_does_not_exist` or something like that.

wjwwood: It would be good to either check that the executable exists before calling the subprocess or by catching these types of errors to be more graceful when there is an error.

wjwwood:
+1 other than comments.

catkin_tools/verbs/catkin_build/job.py
scpeters: Confirmed that these raise errors. Will fix when I have a chance.

scpeters: I wrapped both Popen calls in try/except blocks. How does it look now?

catkin_tools/verbs/catkin_build/job.py
```diff
@@ -139,6 +141,32 @@ def __init__(self, package, package_path, context, force_cmake):
...
+            out, err = p.communicate()
+        except:
+            throws = True
+        if throws or p.returncode != 0:
+            try:
+                out, err = subprocess.Popen(
+                    ['dpkg-architecture', '-qDEB_HOST_MULTIARCH'],
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE).communicate()
+            except:
+                return ""
```
wjwwood: Nit pick: Can you either make this `return ''` or update the above `return ''` to also be `return ""`?

catkin_tools/verbs/catkin_build/job.py
```diff
@@ -139,6 +141,32 @@ def __init__(self, package, package_path, context, force_cmake):
         Job.__init__(self, package, package_path, context, force_cmake)
         self.commands = self.get_commands()
 
+    def get_multiarch(self):
+        if not sys.platform.lower().startswith('linux'):
+            return ''
+        # this function returns the suffix for lib directories on supported systems or an empty string
+        # it uses two step approach to look for multiarch: first run gcc -print-multiarch and if
+        # failed try to run dpkg-architecture
+        throws = False
```
wjwwood: nitpick: Should this be called `threw` or `error_thrown`?

wjwwood:
+1 After the two nitpicks, I'll merge.

catkin_tools/verbs/catkin_build/job.py
```diff
@@ -139,6 +141,32 @@ def __init__(self, package, package_path, context, force_cmake):
...
+        try:
+            p = subprocess.Popen(
+                ['gcc', '-print-multiarch'],
+                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            out, err = p.communicate()
+        except:
```
wjwwood: At first I was gonna let this pass, but it is bothering me...

It's not good practice to catch and silently pass on all exceptions. I know it's more work, but it would be better if you caught specific errors, something like this: `except (OSError, FileNotFoundError):`

scpeters:
I fixed the last three comments and then squashed

wjwwood:
Lgtm, I want to fixup something else, but I'll do that after the merge.

wjwwood:
Never mind, I was thinking `FileNotFoundError` wasn't available before Python 3.3, but it is in Python 2.7 apparently too.

scpeters:
Thanks for the review. I'll start recommending this to the gazebo developers once it is released.

PR #158, opened 2015-03-11
color: moving color args out of build verb and into main program args

jbohren:
There are now a bunch of subcommands in `catkin_tools` which generate color, so this moves the actual processing of these arguments into the main `catkin` command. Note that the arguments still exist in the `catkin build` verb, but they are hidden.

### DIFF
--- a/catkin_tools/verbs/catkin_build/cli.py
+++ b/catkin_tools/verbs/catkin_build/cli.py
@@ -24,13 +24,10 @@ from catkin_tools.argument_parsing import add_context_args
 from catkin_tools.argument_parsing import add_cmake_and_make_and_catkin_make_args
 
 from catkin_tools.common import format_time_delta
-from catkin_tools.common import is_tty
 from catkin_tools.common import log
 from catkin_tools.common import find_enclosing_package
 
 from catkin_tools.context import Context
-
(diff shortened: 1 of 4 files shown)


catkin_tools/verbs/catkin_build/cli.py
```diff
@@ -145,6 +136,10 @@ def prepare_arguments(parser):
     add('--no-summarize', '--no-summary', action='store_false', dest='summarize',
         help='explicitly disable the end of build summary')
 
+    # Deprecated args now handled by sysargs
```
wjwwood: "not handled by catkin command"?

wjwwood:
Other than the comment, +1

catkin_tools/verbs/catkin_build/cli.py
jbohren: Yeah I'll change that

NikolausDemmel:
Does catkin make use of `--no-color` during the alias expansion output? On the current master it does not.

NikolausDemmel:
I actually just tested this and this PR appears to achieve that.

jbohren:
@wjwwood Updates the requested comment change.

PR #163, opened 2015-03-16
common: suppressing warnings from catkin-pkg when just building

jbohren:
This depends on https://github.com/ros-infrastructure/catkin_pkg/pull/114

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -403,14 +403,14 @@ def wide_log(msg, **kwargs):
     wide_log_fn(msg, **kwargs)
 
 
-def find_enclosing_package(search_start_path=None, ws_path=None):
+def find_enclosing_package(search_start_path=None, ws_path=None, warnings=None):
     """Get the package containing the current directory."""
 
     search_start_path = search_start_path or os.getcwd()
     child_path = ''
 
     while True:
-        pkgs = find_packages(search_start_path)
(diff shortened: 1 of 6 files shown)


jbohren:
This is ready now that the latest `catkin_pkg` has been released.

wjwwood:
I don't think it is appropriate to suppress the warnings in all the cases, or at least it's not immediately obvious why each of these instances suppresses the warnings.

Can you expand on why you think it's appropriate to suppress the warnings in each of these cases? If you want to disable them globally then I think it should be a command line configuration and it should not suppress the warnings by default.

The warnings are there for a reason and at least in the case of the package naming conventions we want to expose users to the fact their package name doesn't conform as soon as possible because changing the name of a package is harder the longer you wait. If you know what you're doing then you can disable them, but I just don't see the need to suppress them by default.

jbohren:
> I don't think it is appropriate to suppress the warnings in all the cases, or at least it's not immediately obvious why each of these instances suppresses the warnings.

I've added justifications for each place warnings are suppressed. 

One of the irritating things was that often the same warning would be printed out multiple times. Obviously this indicates that `catkin build` is scraping the same workspaces too many times, but regardless it just didn't work well with the way the `catkin_tools` command line tools work.

I've also added a formatted list of warnings that will get printed at the bottom of `catkin list` unless you call it with the `--quiet` option.

wjwwood:
Ok, lgtm, thanks.

PR #168, opened 2015-03-22
completion: adding bash (zsh) completion

### LINKED ISSUE #15: Add bash/zsh tab completion

jbohren:
**NOTE: Depends on #190 See [cli-completion-full](https://github.com/jbohren-forks/catkin_tools/tree/cli-completion-full) for the integrated branch**

Completes:

- `catkin [--CATKIN-OPTS...] <tab>` completes to all builtin verbs
- `catkin [--CATKIN-OPTS...] build [--BUILD-OPTS...] <tab>` completes to package names given by `catkin list`
- `catkin [--CATKIN-OPTS...] build [--BUILD-OPTS...] -<tab>` completes to `--dry-run --this --no-deps --continue-on-failure --force-cmake --verbose --summarize --no-notify`
- `catkin create <tab>` completes to `catkin create` sub-commands
- `catkin create pkg <tab>` completes to `catkin create pkg` optional arguments
- `catkin profile <tab>` completes to `catkin profile` sub-commands
- `catkin profile [set rename remove] <tab>` completes to existing profiles

***Note:*** If you want to use them in zsh, you still need to source the bash file.

Fixes #15

### DIFF
--- a/completion/catkin_tools-completion.bash
+++ b/completion/catkin_tools-completion.bash
@@ -1,0 +1,109 @@
+# Copyright 2015 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# ZSH support
+if [[ -n ${ZSH_VERSION-} ]]; then
+  autoload -U +X bashcompinit && bashcompinit
+fi
+
+# TODO:
+# - parse --workspace and --profile options in order to complete outside of cwd
+# - autocomplete build options
+
+_catkin()
+{
+  local cur prev catkin_verbs catkin_opts
+
+  COMPREPLY=()
+
+  cur=${COMP_WORDS[COMP_CWORD]}
+  prev=${COMP_WORDS[COMP_CWORD-1]}
+
+  # complete to the following verbs
+  catkin_verbs="build clean config create init list profile"
+
+  # complete to verbs ifany of these are the previous word
+  catkin_opts="--force-color --no-color --test-colors"
+
+  # complete popular catkin build options
+  catkin_build_opts="--dry-run --this --no-deps --continue-on-failure --force-cmake --verbose --summarize --no-notify"
+
+  # complete popular catkin clean options
+  catkin_clean_opts="--all --build --devel --install --cmake-cache --setup-files --orphans"
+
+  # complete popular catkin config options
+  catkin_config_opts="--init --extend --no-extend --cmake-args --make-args --catkin-make-args --space-suffix"
+
+  # complete popular catkin create options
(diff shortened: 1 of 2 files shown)


NikolausDemmel:
Cool! I looking forward to the feature.

Basic usage works fine, but e.g. the following breaks:

    catkin build --no-deps[TAB]

with error:

    bash: compgen: --: invalid option
    compgen: usage: compgen [-abcdefgjksuv] [-o option]  [-A action] [-G globpat] [-W wordlist]  [-F function] [-C command] [-X filterpat] [-P prefix] [-S suffix] [word]

similar for something like:

    catkin --h[TAB]

P.S.: Is `sudo python setup.py develop` supposed to install the `catkin-completion.bash` file? It does not seem to do that over here.

jbohren:
> Basic usage works fine, but e.g. the following breaks:

Yeah, bash completion hacking is new for me, this is resolved in #f33e174

> P.S.: Is `sudo python setup.py develop` supposed to install the `catkin-completion.bash` file? It does not seem to do that over here.

No, I tried and it doesn't appear to install it, which makes sense since it's an external config file. For developing you can either add the following to your shell:

```
source /path/to/catkin_tools/completion/catkin-completion.bash
```

Alternatively you can link it to `/etc/bash_completion.d`:

```
cd /etc/bash_completion.d sudo ln -s /path/to/catkin_tools/completion/catkin-completion.bash
```

Note that if you want to use them in zsh, you still need to source the bash file.

NikolausDemmel:
Phantastic. It even completes the build optional arguments :+1:

adeguet1:
Can't wait to use it.

wjwwood:
I think that the installation code should be more sophisticated, I actually had already done that part:

https://github.com/catkin/catkin_tools/commit/49165cf555f0566da02a1ccda987fbaecf2885b8

Can you take a look at that commit and consider using all or part of it?

jbohren:
> I think that the installation code should be more sophisticated, I actually had already done that part:

Oh cool. Yeah I'll just update that with the added file.

wjwwood:
Ok, just make sure to try it out, I only tested that strategy a bit.

wjwwood:
Reviewing this I have a few questions:

- What's this file for? `catkin_tools-completion.sh`
- Have you tested installing this in a few different scenarios? Looking at my logic in the `setup.py` and poking at the value of `sys.prefix`, I'm not sure they're going to the right place.
- Should there be a `.zsh` which just sources the `.bash`? Even if the `.bash` has logic for when it is sourced by `.zsh` it seems like there should be a `.zsh` file.

jbohren:
> - What's this file for? catkin_tools-completion.sh

Left over from your 0c97f5f. I got rid of it.

> - Have you tested installing this in a few different scenarios? Looking at my logic in the setup.py and poking at the value of sys.prefix, I'm not sure they're going to the right place.

No I haven't tested the installation. I don't know if I have time to.

> - Should there be a .zsh which just sources the .bash? Even if the .bash has logic for when it is sourced by .zsh it seems like there should be a .zsh file.

I don't think that's really necessary. I'm not sure where it would go, anyway.

jbohren:
@wjwwood Separated the optimizations out into #190

jbohren:
See [cli-completion-full](https://github.com/jbohren-forks/catkin_tools/tree/cli-completion-full) for the integrated branch.

wjwwood:
Ok, I finally got around to testing this and fixing up my install logic on both Linux and OS X, in a variety of scenarios. I merged this manually after a rebase and adding this commit: f797eb6

jbohren:
> Ok, I finally got around to testing this and fixing up my install logic on both Linux and OS X, in a variety of scenarios. I merged this manually after a rebase and adding this commit: f797eb6

Cool!

PR #175, opened 2015-03-27
build: fixing blacklist/whitelist to use package names instead of paths

jbohren:
Fixes issue: https://github.com/catkin/catkin_tools/pull/152#issuecomment-86957722
Fixes confusion: https://github.com/catkin/catkin_tools/pull/152#issuecomment-88399324

### DIFF
--- a/catkin_tools/verbs/catkin_build/build.py
+++ b/catkin_tools/verbs/catkin_build/build.py
@@ -168,12 +168,6 @@ def determine_packages_to_be_built(packages, context):
     packages_to_be_built = []
     packages_to_be_built_deps = []
 
-    # Only use whitelist when no other packages are specified
-    if not packages:
-        # Check whitelist
-        if len(context.whitelist) > 0:
-            packages = [p[1].name for p in context.packages if (p[1].name in context.whitelist)]
-
     # Determine the packages to be built
     if packages:
         # First assert all of the packages given are in the workspace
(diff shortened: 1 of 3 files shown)


catkin_tools/verbs/catkin_build/build.py
```diff
@@ -195,13 +189,18 @@ def determine_packages_to_be_built(packages, context):
...
+            packages_to_be_built = ordered_packages
 
     # Filter packages with blacklist
     if len(context.blacklist) > 0:
-        packages_to_be_built = [p for p in packages_to_be_built if (p[0] not in context.blacklist or p[0] in packages)]
+        packages_to_be_built = [
+            p for p in packages_to_be_built if (p[1].name not in context.blacklist or p[1].name in packages)]
         packages_to_be_built_deps = [
-            p for p in packages_to_be_built_deps if (p[0] not in context.blacklist or p[0] in packages)]
+            p for p in packages_to_be_built_deps if (p[1].name not in context.blacklist or p[1].name in packages)]
```
wjwwood: It would probably be better to expand the tuple rather than access it by index. Something like:

```python
[(pth, pkg) for pth, pkg in packages_to_be_built_deps if pkg.name not in context.blacklist or pkg.name in packages]
```

wjwwood:
Other than the comment lgtm.

catkin_tools/verbs/catkin_build/build.py
jbohren: Agreed.

jbohren:
@wjwwood This is good to review.

PR #179, opened 2015-04-01
context: adding append/remove args option

### LINKED ISSUE #176: config: add `--append` option for list-type configuration options

jbohren:
- calling `catkin config [--append-args | -a] [ARGS]` will append elements to list args
- calling `catkin config [--remove-args | -r] [ARGS]` will remove elements from list args
- also fixing context load/save code style https://github.com/catkin/catkin_tools/pull/147#discussion_r26276789
- also fixing line-too-long code style

Fixes #176

### DIFF
--- a/docs/verbs/catkin_config.rst
+++ b/docs/verbs/catkin_config.rst
@@ -23,6 +23,26 @@ the root of the workspace. Doing so will not modify your workspace. The
 ``catkin`` command is context-sensitive, so it will determine which workspace
 contains the current working directory.
 
+Appending or Removing List-Type Arguments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Several configuration options are actually *lists* of values. Normally for
+these options, the given values will replace the current values in the
+configuration.
+
+If you would only like to modify, but not replace the value of a list-type
+option, you can use the ``-a`` / ``--append-args`` and ``-r`` /
+``--remove-args`` options to append or remove elements from these lists,
+respectively.
+
+List-type options include:
+
+ - ``--cmake-args``
+ - ``--make-args``
+ - ``--catkin-make-args``
+ - ``--whitelist``
+ - ``--blacklist``
+
 Installing Packages
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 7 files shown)


docs/verbs/catkin_config.rst
```diff
@@ -189,12 +189,20 @@ To clear the blacklist, you can use the ``--no-blacklist`` option:
 Note that you can still build packages on the blacklist and whitelist by
 passing their names to ``catkin build`` explicitly.
 
+Appending or Removing List-Type Arguments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Several configuration options are actually *lists* of objects. For such
+options, you can use the ``--append-args`` and ``--remove-args`` options to
+append or remove elements from these lists, respectively.
```
wjwwood: Should it (is it?) mentioned somewhere that the default behavior is to replace the list?

wjwwood:
Other than the comment +1

docs/verbs/catkin_config.rst
jbohren: Yeah, that should be made more clear.

hgaiser:
As mentioned in https://github.com/catkin/catkin_tools/pull/152 , I think it would be better to let the default behaviour be to add items to a list and have a --clear-[something] to clear the list. This maintains the same functionality with one less argument (and to be honest, is what I would expect it to do).

jbohren:
> As mentioned in #152 , I think it would be better to let the default behaviour be to add items to a list and have a --clear-[something] to clear the list. This maintains the same functionality with one less argument (and to be honest, is what I would expect it to do).

I agree. The `-clear-*list` arguments were misleading. They've been renamed to `--no-whitelist` and `--no-blacklist` so that they conform to the pattern used by other arguments like `--cmake-args` / `--no-cmake-args`.

With respect to making it append by default, making the `--whitelist` and `--blacklist` options behave differently from `--cmake-args` and `--make-args` can make the interface start getting really confusing. I think adding an extra `-a` to append isn't that onerous.

The real problem here is that "whitelist" can be a noun or a verb. The argument is meant to be a noun, i.e. you're setting *the whitelist* but it's equally reasonable at first glance to read it as if you're *whitelisting* a package. I also tried to clarify this in the documentation: https://github.com/catkin/catkin_tools/blame/master/docs/verbs/catkin_config.rst#L239-252

jbohren:
@wjwwood I've updated the documentation

wjwwood:
lgtm, thanks.

PR #185, opened 2015-04-03
optimization: caching resultspace environments

jbohren:
This can save something like 0.2s when starting up if the same resultspace gets loaded twice.

### DIFF
--- a/catkin_tools/resultspace.py
+++ b/catkin_tools/resultspace.py
@@ -9,6 +9,9 @@ from .utils import which
 CMAKE_EXEC = which('cmake')
 SORT_EXEC = which('sort')
 
+# Cache for result-space environments
+_resultspace_env_cache = {}
+
 
 def get_resultspace_environment(result_space_path, quiet=False):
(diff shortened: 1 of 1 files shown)


wjwwood:
Is there a reason and/or a way to get the uncached version? This is the pattern I usually follow:

```python
_foo_cache = {}

def get_foo_cached(arg):
    global _foo_cache
    if arg not in _foo_cache:
        _foo_cache[arg] = get_foo(arg)
    return _foo_cache[arg]


def get_foo(arg):
    ... Some implementation
    return foo

... Somewhere else:

foo = get_foo_cached(arg)
```

This way it is obvious from the calling code that you are getting a cached value, and it is possible to get an uncached result if desired. What's your opinion?

jbohren:
I don't really care either way. The only time you'd want to reload it is if
it changes while a single instance of catkin is running.

jbohren:
@wjwwood Take a look at 866e111, I just added an option to enable/disable caching.

wjwwood:
lgtm, thanks.

PR #186, opened 2015-04-03
context: use PWD environment variable (if available) to handle running from symlinked directories

### LINKED ISSUE #182: catkin build does not follow symlinks properly
I am working on OSX with ROS, but some packages can't compile there due to drivers that aren't available. For that reason, I am also using ubuntu in a virtual machine. To reduce duplication, I symbolically linked the `src` folder from OSX to the ubuntu guest machine. In that way, source code is the same in OSX and in ubuntu but both OS's have their own `build` and `devel` folder. Running `catkin build` in ubuntu somewhere inside the symlinked folder will try to erroneously use the build/devel from OSX, which is incompatible with the ubuntu version and will fail.

Long story short: running `catkin build` in `/home/hgaiser/catkin_ws/src<symlinked>/some_package` (where src is a symlink to `/media/psf/Home/catkin_ws/src`) is equal to `cd /media/psf/Home/catkin_ws && catkin build`, which doesn't work.

jbohren:
Fixes #182

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -48,6 +48,33 @@ class FakeLock(object):
         pass
 
 
+def getcwd(symlinks=True):
+    """Get the current working directory.
+
+    :param symlinks: If True, then get the path considering symlinks. If false,
+    resolve the path to the actual path.
+    :type symlinks: bool
+    :returns: the current working directory
+    :rtype: str
+    """
+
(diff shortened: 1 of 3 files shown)


hgaiser:
Seems to work, thanks!

wjwwood:
lgtm.

PR #190, opened 2015-04-07
optimization: Improve context loading and `list` verb

jbohren:
- Adding `--unformatted` tag to `catkin list` verb for use in cli
  completion
- Adding a lazy env-loading mode for `Context.load` when we only need
  the context for this workspace and not the loaded resultspace env
  variables like CMAKE_PREFIX_PATH. This is necessary to improve the
  runtime of `catkin list` and is also used to improve the runtime of
  `catkin clean`.
- Removing superfluous env-loading from `Context` `extend_path` accessor

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -75,8 +75,18 @@ class Context(object):
         'space_suffix']
 
     @classmethod
-    def load(cls, workspace_hint=None, profile=None, opts=None, strict=False, append=False, remove=False):
-        """Load a context from a given workspace and profile with optional modifications.
+    def load(
+        cls,
+        workspace_hint=None,
+        profile=None,
+        opts=None,
+        strict=False,
+        append=False,
+        remove=False,
+        load_env=True
+    ):
+        """Load a context from a given workspace and profile with optional
+        modifications.
 
         This function will try to load a given context from the specified
         workspace with the following resolution strategy:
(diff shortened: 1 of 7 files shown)


jbohren:
@wjwwood These improvements were split out of the tab-completion PR because they're useful elsewhere, do you have any time to review?

wjwwood:
I'll try to review it tonight.

catkin_tools/context.py
```diff
@@ -474,14 +497,12 @@ def extend_path(self):
...
-                get_resultspace_environment(value)
-            self.__extend_path = value
-        except IOError as exc:
-            raise ValueError("Unable to extend workspace from \"%s\": %s" % (value, exc.message))
+        if value is not None:
+            if not os.path.isabs(value):
+                value = os.path.join(self.workspace, value)
+            if not os.path.exists(value):
+                raise ValueError("Resultspace path '{0}' does not exist.".format(value))
+        self.__extend_path = value
```
wjwwood: Why is the call to `get_resultspace_environement()` no longer needed here? Is it deferred to somewhere else?

wjwwood:
Other than the comment +1.

catkin_tools/context.py
jbohren: Previously the setter was doing the heavy lifting of loading the extended environment, but we want to be explicit as to when we do this loading. This is now done in `Context.load_env()`.

PR #192, opened 2015-04-07
verbs: adding shell verbs 'cd' and 'source'

jbohren:
**NOTE: Depends on #165**

This is a bash/zsh wrapper function which, when sourced, enables the following:

- `catkin cd [ARGS...]` -- Change to package directory in source space where `ARGS` are any valid `catkin locate` arguments
- `catkin source` -- Source the develspace of the containing workspace
- `catkin -w /path/to/ws source` -- Source the develspace of the workspace `/path/to/ws source`

### DIFF
--- a/scripts/catkin_shell_verbs.bash
+++ b/scripts/catkin_shell_verbs.bash
@@ -1,0 +1,98 @@
+# Catkin Shell Verbs
+
+function __catkin_help_augmented() {
+  # Print out an augmented --help line
+
+  # Get catkin executable
+  CATKIN=$1
+
+  # Add our shell verbs to the main helpline
+  ORIGINAL_HELP_TEXT="$($CATKIN --help)"
+  AUGMENTED_HELP_TEXT=""
+
+  # Read each line of the help text, preserving whitespace
+  while IFS= read -r LINE; do
+    if [[ $LINE =~ "^\s+clean\s+" ]]; then
+      # Insert cd summary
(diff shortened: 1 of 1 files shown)


wjwwood:
This one needs to be updated with `catkin find` => `catkin locate`.

jbohren:
> This one needs to be updated with catkin find => catkin locate.

Yep.

jbohren:
I updated it to use `locate` instead of `find`.

wjwwood:
This is neither documented nor automatically sourced. So before merging this it needs to be documented.

@k-okada you mentioned it should be added to the bug squashing list, can you test and document the feature (both how to activate it and how to use it)?

k-okada:
ok, so where should I put documents?

--
◉ Kei Okada

wjwwood:
I'd add a "Extra catkin shell commands" section to the index:

https://github.com/catkin/catkin_tools/blob/master/docs/index.rst

Where you can describe that these are shell functions and that they can be optionally activated by `source`'ing the file in this pr. Then a bullet list of the new shell commands and either a short description of each or a link to another document in the docs folder which has more detail (similar to the list of built-in verbs).

k-okada:
@wjwwood  added docs to #244, 
this is not verb, but shell support and activated by 'source /opt/ros/indigo/etc/bash_completion.d/catkin_shell_verbs.bash', just to make sure this is not automatically loaded in `source /opt/ros/indigo/setup.bash'

wjwwood:
This was replaced by #244.

k-okada:
thank you!

--
◉ Kei Okada

PR #194, opened 2015-04-09
build: adding experimental installwatch-based log generation

jbohren:
This is an experiment in providing support for rolling-back the products of a package in the develspace.

This generates `PKGNAME_cmake_products.log` and `PKGNAME_make_products.log` in the `build_log` directory in the buildspace. The log is generated by running `installwatch` for the CMake and Make calls that `catkin build` makes.

These files contain all file-access operations by intercepting these specific system calls, and all file write calls are represented as `fopen` calls. This is not very portable, and I don't know if there's an easy equivalent for Windows. 

Just something to think about.

Things to do:
- [ ] add a package-specific option to `catkin clean` which cleans the results of a package and (optionally) cleans all packages which depend on that package
- [ ] add performance tests to gauge the impact of `installwatch` on building
- [ ] make it so `installwatch` is only used on `catkin` packages, for plain `cmake` packages, the `install_manifest.txt` can be used.
- [ ] add switch to turn `installwatch` on and off

### DIFF
--- a/catkin_tools/verbs/catkin_build/job.py
+++ b/catkin_tools/verbs/catkin_build/job.py
@@ -27,6 +27,7 @@ from .common import create_build_space
 from .common import generate_env_file
 from .common import get_python_install_dir
 
+INSTALLWATCH_EXEC = which('installwatch')
 MAKE_EXEC = which('make')
 CMAKE_EXEC = which('cmake')
 
@@ -298,6 +299,7 @@ class CatkinJob(Job):
         if not os.path.isfile(makefile_path) or self.force_cmake:
             commands.append(CMakeCommand(
                 env_cmd,
+                [INSTALLWATCH_EXEC, '-o', os.path.join(self.context.build_space_abs, 'build_logs', '%s_cmake_products.log' % self.package.name)] +
                 [
                     CMAKE_EXEC,
                     pkg_dir,
@@ -309,9 +311,13 @@ class CatkinJob(Job):
         else:
             commands.append(MakeCommand(env_cmd, [MAKE_EXEC, 'cmake_check_build_system'], build_space))
         # Make command
+        make_command = (
+            [INSTALLWATCH_EXEC, '-o', os.path.join(self.context.build_space_abs, 'build_logs', '%s_make_products.log' % self.package.name)] +
+            [MAKE_EXEC] +
+            handle_make_arguments(self.context.make_args + self.context.catkin_make_args))
         commands.append(MakeCommand(
             env_cmd,
-            [MAKE_EXEC] + handle_make_arguments(self.context.make_args + self.context.catkin_make_args),
+            make_command,
             build_space
         ))
         # Make install command, if installing


tfoote:
This is a nice bit of functionality, however I'm a little worried about this becoming scope creep for this tool.With this it goes from a build tool to a build tool as well as a sort of mini-package manager.

Installwatch is the tool behind checkinstall. And it seems like it might make sense to just use checkinstall instead of building the lower level tool into our build toolchain. Checkinstall builds the results into the standard format of deb packages which are rock solid and well understood by our users. They can just use their package interaction tools. Otherwise we end up needing to reproduce all the packaging, install, listing, introspection and uninstall semantics already done by dpkg. (It also works on Redhat based systems with rpms)

jbohren:
> This is a nice bit of functionality, however I'm a little worried about this becoming scope creep for this tool.With this it goes from a build tool to a build tool as well as a sort of mini-package manager.

I don't think this is scope creep at all. People expect to be able to build and clean parts of their workspace, and are frankly confused by the fact that they need to blow away their entire devel spaces just to completely clean the products of a single package.

*Suggesting to use `checkinstall` instead of `installwatch` misses the point that this is really just trying to monitor which files are put into a catkin devel space, and give users the ability to delete them in order to clean the build products.*

Fundamentally, this addresses a flaw in the design of the catkin merged devel space concept as a whole. Specifically, a user can't clean all the products of a single package without deleting his or her _entire_ devel space. `catkin_tools` manages to decouple the build process by building in isolated build spaces for each package, and the isolated devel space approach aims to solve the problem of cleaning build products. Unfortunately, the isolated devel space design is very heavy since it requires so many setup files to be sourced.

Until isolated devel spaces work properly, people end up wasting a _lot_ of time having to delete their entire build and devel spaces every time they remove or switch a package in their workspace. At this point, `installwatch` appears to be a very simple and lightweight way of logging what each package puts into the develspace. With this addition we could prevent some of the most common buildtool usage failures where people think one thing is in their devel space, but really it's something else. If you guys 

In my opinion, the only two issues I could see preventing this approach are (1) cross-platform support and (2) the potential overhead of intercepting the system calls. For (1) there may be other ways in which this could be implemented in a cross-platform way, and for (2) it simply requires testing to see what the impact is.

wjwwood:
I think the use case @jbohren is describing is within the scope of the tool, but there are a couple technical challenges that we'll have to address.

I take the point that this pull request is an experiment to see if this can be accomplished, but separately we need to decide what we would like the tool to do. The first step that I think we should take is to define what expectation we have for the `catkin clean <package name>` operation. These are just some things that I think we should consider off the top of my head:

- Does it remove from install space?
- If yes, does it remove from both install space and devel space if I have built into both at some point?
- Does it work with non-catkin packages?
- Does it try to change the environment on the next source, or does the user need to make a fresh terminal and source the new `setup.*sh` after re-building?
- For each of these questions what will the user assume and what must we implement? What can we do to mitigate places where the user's expectations do not align with the required behavior?

I haven't looked at the implementation of this in detail, but I have some thoughts about alternatives should this approach not work out.

You could follow the approach of Homebrew, building each package into a devel space, then symbolically linking the contents of each package's devel space into a common folder. Homebrew installs all applications into separate folders under `/usr/local/Cellar/<package>/<version>` and then symlinks a selection of those into `/usr/local` which is on the system paths. Since you have to collect the file for symlinking, you could also store them. The process for removing a package them becomes unlinking the files you created, deleting the package's own devel and build spaces.

Another, similar, approach is something we are trying out in ROS2 right now, which is to replace the devel space all together with a symlink version of the install target. Basically, right now in catkin we change cmake settings to get generated files to fall into the devel space and then we have some special files which live in the devel space and reference the source space (like the `__init__.py` files created when your package has both messages and Python code). This is a big change, and I'm not sure we'll ever be able to roll it out into catkin since it would be so disruptive, but it has a few big advantages:

- It prevents users from having to understand the difference between a devel space and the install space (a much more common concept in software packaging).
- It removes the need to have special hacky files which combine resources from devel and source space (like the `__init__.py` files I mentioned).
- It still allows you to build the package and then edit Python files and resources like launch files without rebuilding.
- It encourages users and our tools like `catkin_simple` to install their files before using them (which makes the transition to a releasable package easier).
- It prevents package developers from having to deal with small differences in layout and behavior between the devel space and the install space.
- It generates a CMake `install_manifest.txt` which can be used for an uninstall target.

One down side is that it requires that you install the package before using it. This has two affects, it prevents you from doing the `catkin_make` style one cmake and one make invocation for many packages, and it adds some overhead per package on build. There may be solutions to these issues with this strategy, but we haven't found them yet.

I mention what we're trying for ROS 2 not to suggest that we should try and adopt that in ROS 1 or that we should "just wait" for ROS 2, but rather to give you potential ideas for this problem.

jbohren:
> You could follow the approach of Homebrew, building each package into a devel space, then symbolically linking the contents of each package's devel space into a common folder. Homebrew installs all applications into separate folders under /usr/local/Cellar/<package>/<version> and then symlinks a selection of those into /usr/local which is on the system paths. Since you have to collect the file for symlinking, you could also store them. The process for removing a package them becomes unlinking the files you created, deleting the package's own devel and build spaces.

See #196 for most of the files, `make clean` in the build directory will remove all the *build* products from the develspace. So it would be sufficient to just run the CMake *configure* step with a "fake" devel space, and then symlink those files. Let's continue talking over in #196.

Issue #188, opened 2015-04-04
build: -j and -p defaults


NikolausDemmel:
With the jobserver implementation merged, I suggest to change the default of the `-p` argument from "cpu count" to "value of `-j` argument". 

If you pass neither, both will be "cpu count" as they are currently, making for a reasonable default.

I would assert, that most of the time when passing `-j` explicitly, you would also want to pass `-p` with the same value. By changing the default as suggested, `-j` would be more of DWIM.

One could suggest to make it the other way round, i.e. have `-j` default ot the value of `-p`, but IMHO there is a valid use case for wanting to `-p1` and still have `-j` default to cpu count. Also, `-j` is more of a standard for people who know `make` and `catkin_make`.

Thoughts?

jack-oquin:
Makes sense to me, as a naive user with little understanding of these details.

jbohren:
+1

wjwwood:
This is effectively what is happening since the start of a build is blocked on a `with job_server` context:

https://github.com/catkin/catkin_tools/blob/master/catkin_tools/verbs/catkin_build/executor.py#L127

I thought this might be unintended, because it seems like it would also prevent `cmake` from running if there are no job tokens available, however, it's still not entirely clear to me how the job server enter and exit are supposed to work along with the `make` subprocesses. Ultimately, I decided to merge based on the acceptance testing I did.

The observable result is that the jobs which are listed as "building" (taking up one of the `-p` slots) can actually be doing nothing and waiting on the job server to give them time with a job server token. So maybe @jbohren can shed some light on why the `job_server` context is there and what the intended behavior is supposed to be.

With all of that being said I think @NikolausDemmel's suggestion actually makes sense and I would take a pull request which changed this (along with any documentation that's needed to convey this behavior).

wjwwood:
@NikolausDemmel given that this is run on a 4 core machine:

```
% catkin build --no-jobserver -j1
```

What should be the default value of `-p`?

jbohren:
I was thinking that if e want to implement a more sophisticated policy that
the `p` value should, in general, be less than or equal to the `j` value.

NikolausDemmel:
> I was thinking that if e want to implement a more sophisticated policy that
the `p` value should, in general, be less than or equal to the `j` value.

Do you mean simply setting `p = min(p, j)` (if the jobserver is in use)? Seems reasonable as a policy and we don't have to change default values.

wjwwood:
[...]

Issue #78, opened 2014-07-02
build: Generate .catkin file when using merged develspace and catkin in "parallel-safe" mode


jbohren:
If catkin gets a "parallel-safe" mode, as described in ros/catkin/issues/675, then we could use that to generate a `.catkin` develspace marker file from the information generated by the process described in ros/catkin/issues/674.

jonbinney:
This is still a problem, and in my opion is a pretty big problem with catkin tools right now. Sometimes after running `catkin build` and sourcing `devel/setup.bash`, some packages are not in my `ROS_PACKAGE_PATH`. This has been happening to me and my coworkers for months, and I only yesterday figured out that it was due to the concurrency problems in catkin tools.

Here's a test script that will generate a workspace with a bunch of packages and try to elicit the problem: https://gist.github.com/jonbinney/0a09e57db3676e0bd29f
That scripts compiles a workspace with 20 packages 100 times. On my computer, it caused a corrupted `devel/.catkin` (and thus a corrupted ROS_PACKAGE_PATH) 12/100 times.

Note that while having lots of packages and having a carefully crafted cmakelists.txt in the test script causes the bug to happen more often, the bug exists for any workspace. In practice I run into this bug several times a week during normal ROS development.

wjwwood:
@jonbinney did you ever run this with the `--isolated-devel` option? My guess is that this issue would not be present in that case. But obviously the isolated devel space is not very nice to use.

These are the solutions as I see them:

- Isolated devel space with symlinked "merged" devel space
 - This is @jbohren's approach in #196
 - It does an isolated devel space for every package, but symlinks unique files, and merges non-unique ones (like `.catkin`), see: https://github.com/catkin/catkin_tools/pull/196 (the description about linked devel space)
- Change catkin to use something like ament's package resource
 - We solve this problem more generically in ament by avoiding packages from installing files which may collide (avoiding situations like `.catkin` all together)
 - This would be hard since we'd need to get it in catkin and then release and make sure no other existing tools are affected adversely.
- Have `catkin_tools` look at the `.catkin` file before building packages, build packages, then update the `.catkin` file to make sure it is correct.
 - This is similar to what @jonbinney's test script is doing, but rather than just asserting that it is correct or not, it would also just update it if needed.

[...]

Issue #66, opened 2014-06-10
build: ui stuck in "calculating new jobs..."


NikolausDemmel:
Sometimes it seems the build command ui gets stuck with the line

    [build] Calculating new jobs...

for a long time. It seems like the command is actually still building packages in the background, just not updating the display. I have a hard time reproducing this reliably, so I'm not sure.

jbohren:
The calculating jobs step isn't there now that #249 is merged.

PR #196, opened 2015-04-13
[major] executor/clean: Adding per-package cleaning, linked develspaces, and a new execution pipeline

jbohren:
***NOTE: This has been superseded by #247, #249, #276, and #293.***

## Overview

This PR involves a dramatic refactor of the core execution pipeline for building and cleaning packages in order to implement two high-priority features:

* Per-package / partial-workspace cleaning (reversible build actions)
  * Requires build "linked" develspaces, which are isolated per-package develspaces whose contents are symbolically-linked into a merged develspace
  * Requires using the `catkin build` execution pipeline outside of the `build` verb
* (#1) Automatic `stderr`-based warning and error detection
  * Requires separating `stdout` from `stderr` during execution of each job

Additionally, addresses the following outstanding issues:

* (#2) More general use of the GNU-Make compatible job server in the new execution pipeline
* (#78) The linked develspace construction avoids potential race conditions in the generation of the .catkin file
* (#28) (#45) (#65) The linked develspace provides the benefits of isolated devel spaces without the runtime complexity of generating (#28) and maintaining (#45) setup files, and partial cleaning (#65)
* (#188) By default, the `-p` option is set to whatever the `-j` option is
* (#59) The `catkin` tool should always exit cleanly when ctrl-c'd due to the new use of asyncio
* (#87) Supports workspaces containing only vanilla `cmake` packages
* (#63) Provides a brief build summary with important notices at the end of the build
* (#66) The pending jobs are now calculated in a completely different way and are unlikely to deadlock 
* (#213) Running integration tests on Linux and OS X
* (#208) Fixes jobserver support on Python3

It also provides the following:
* More transparency into what the `catkin_tools` executor is doing via named job "stages"
* Enables stages of jobs which are implemented in pure python and don't require subprocesses
* Clearer designation of errors and warnings when they are printed to the console
* In order to re-generate warnings a `--pre-clean` option has been added to the `build` verb to `make clean` in the build directory before running `make`
* In order to build unbuilt packages, an `--unbuilt` option has been added to the `build` verb to build any packages in the workspace which have not been built
* More introspection into the build status by displaying cmake package build percentages in status line

It also resulted in some unexpected benefits:
* Build times can be sped up dramatically for no-op packages since the env.sh files aren't needed for merged or linked develspaces, for example:
 * A workspace with 188 pre-built packages completes in ~17s instead of ~55s (3.2x speedup)

It also revealed the following problems:
* (#208) The jobserver doesn't appear to work with Python3.4

## Per-Package Cleaning and the Linked Devel Space

This PR provides support for cleaning individual packages with the following syntax:

```
catkin clean [-a | -b | -d | -i] [-f] [--orphans] [--deps] [--deinit] [[PKG] PKG ...]
```

Where `-b` means buildspace, `-d` means develspace, `-i` means installspace, and `-a` means all three. Other examples include:

### `catkin clean` Examples

* `catkin clean` / `catkin clean -a`
 * Clean the entire build, devel, and install spaces (currently supported)
* `catkin clean -af` / `catkin clean --all --force`
 * Clean the entire build, devel, and install spaces without asking for verififcation
* `catkin clean -b`
 * Clean just the build space (currently supported)
* `catkin clean pkg_a pkg_b`
 * Clean `pkg_a` and `pkg_b` from the build, devel, and install spaces
* `catkin clean --deps pkg_a`
 * Clean `pkg_a` as well as any packages which depend on `pkg_a`.
* `catkin clean --orphans`
 * Clean all packages which have been built but are no longer in the source space.
* `catkin clean --deinit`
 * De-initialize workspace (delete `.catkin_tools`) directory

### Approach

Doing this right is hard, but it's an important feature. I think this sort of per-package `clean` functionality is one of the the final big features preventing `catkin_tools` from being complete. In complex workspaces with hundreds of packages, debugging can be a huge challenge, and this is only made more challenging by the inability to reliably remove all of the products of a specific packages from a workspace.

#### CMake Packages
Vanilla CMake packages are handled trivially since their `install` target is used to write files to `devel` and `install` spaces. This target creates an `install_manifest.txt` which lists all the generated files, so to clean CMake packages, we can simply remove these files.

The normal location for `install_manifest.txt`, however, is in the package's build space. This file could get overwritten if the user builds both a devel and install variant, so we copy this file into the appropriate result space for safe keeping. Also, the `install_manifest.txt` file does not specify directories which are created by `cmake`, so the CMake clean job will remove any directories which are made empty through the removal of files listed in the `install_manifest.txt` file.

#### Catkin Packages
Catkin packages prove more challenging for a few reasons. For Catkin packages, there are two different steps which generate files in the `devel` space: the CMake configure step, and the Make build step. It's possible to clean the files generated in the build step with the standard `clean` target, but this still leaves numerous files generated by the configure step. Catkin packages also each generate some of the same files such as the setup shell scripts and each write to `.catkin` file which lists all the source packages in the workspace. This all makes for a big mess to clean up.

The approach used in this PR is to build each package into an isolated devel space, and then symbolically link the contents into the merged devel space. This prototype behavior is enabled by passing the `--link-devel` option to `catkin build`. Since `catkin_tools` does the linking, it can generate a `devel_manifest.txt` similar to CMake's `install_manifest.txt` which logs all of the files that a given package contributes to the devel space. 

If a file that was listed in the previous link step is no longer found in the isolated devel space, it is removed from the merged devel space.

The above handles tracking the files generated by each package, but doesn't handle collisions very well. In order to handle collisions, the symlink generation step counts the number of times a collision occurs for each file. This information is stored in a file called `devel_collisions.txt`. 

In summary:

* The isolated devel spaces are: `DEVEL_PREFIX/.catkin_tools/PKG_NAME/linked_devel`
* The devel manifests are stored in: `DEVEL_PREFIX/.catkin_tools/PKG_NAME/devel_manifest.txt`
* The collisions file is stored in: `DEVEL_PREFIX/.catkin_tools/devel_collisions.txt`
* All files from each isolated devel space are linked into: `DEVEL_PREFIX/`

The clean step for a given package reads its `devel_manifest.txt` and for each file that was generated it either removes the file or decrements the count in the `devel_collisions.txt` file.

#### Setup File Generation
One challenge with this approach, as mentioned above, is that catkin packages normally generate several files in the root of the devel space used for "sourcing" that devel space. These include:

* `.catkin`
* `.rosinstall`
* `_setup_util.py`
* `env.sh`
* `setup.sh`
* `setup.bash`
* `setup.zsh`

With the exclusion of `.catkin`, these files are all generated from [catkin_generate_environment.cmake](https://github.com/ros/catkin/blob/indigo-devel/cmake/catkin_generate_environment.cmake) during the CMake configure step of a catkin package, and several of them include information computed by existing Catkin CMake code. Since some of these reference the absolute path of the develspace, they can't simply be copied from the isolated devel spaces of the constituent packages.

The simplest solution to this problem is to have a "ghost" bootstrap package called `catkin_tools_bootstrap` which gets generated the first time a given resultspace is built. This guarantees that the setup files are generated correctly by Catkin, itself.

A convenient side-effect of the `catkin_tools_bootstrinstallap` package is that it also enables `catkin build` to generate setup files for a develspace consisting only of vanilla CMake packages.

For the `.catkin` file, we make `catkin_tools` responsible for managing its contents. This is for two reasons: 
 1. There have been observed problems building Catkin projects in parallel and writing to this file
 2. If we clean a package, we want to remove its source space from this file

Since `catkin_tools` will be modifying this file in parallel from a single process, we can manage exclusive access to it and prevent collisions. Additionally, we can remove packages from it when they are removed from the workspace.

#### Additional Notes
The current implementation works in Unix-based systems, but could be extended to Windows with [one of these methods](http://stackoverflow.com/questions/1447575/symlinks-on-windows).

### Executor

This PR completely re-implements the execution pipeline using `osrf_pycommon` and the `trollius` library (a Python2 and Python3 compatible version of Python3.4's `asyncio`. This *yields* an execution pipeline which can be used by `catkin clean` (and other potential verbs) and enables us to extract and display `stdout` and `stderr` separately. As such, it solves #1.

#### Execution Jobs and Stages
The new execution pipeline operates on a task model where a given "task" (build, clean, etc) can be decomposed into an acyclic "job" dependency graph, where each job is a sequence of "stages". Each job can be executed in parallel once its dependencies have been completed, and each stage within a job must be executed serially.

A [`Job`](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/execution/jobs.py#L20-L40) is a simple container with a few non-mutating utility functions. It is designated with a unique Job ID (`jid`) and has a list of the `jid`s of the jobs on which it depends. In `catkin_build`, these `jid`s are normally package names.

A `Stage` is some atomic operation, which contains a non-unqie label (such as "cmake", "make", or "install"). There are currently two types of stages: [`CmdStage`](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/execution/stages.py#L20-L56) and [`FunStage`](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/execution/stages.py#L59-L78). The `CmdStage` describes a system command. This type is similar to the current stage implementation, except it also facilitates using asynchronous protocols for capturing `stdout` and `stderr`. The `FunStage`, on the other hand, describes a blocking Python function. This second stage type enables us to affect the filesystem in arbitrary ways without spawning subprocesses or relying on CMake's shell utilities.

The construction of `Job` and `Stage` instances should have no side-effects, and are designed to be lazily evaluated only once their dependencies have been completed. Once a collection of `Job` objects have been created, they are passed directly to the executor.

#### Executing Jobs Asynchronously

The [executor](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/execution/executor.py#L108-L112) is an `asyncio` coroutine which executes each job asynchronously. Jobs follow a well-defined lifecycle. At any time, a given job is on one of the following lists:

 * *pending* - Jobs which are not ready to be executed (unfinished deps)
 * *queue* - Jobs which are ready to be executed
 * *active* - Jobs which are currently executing
 * *abandoned* - Jobs which were meant to be executed, but will not be attempted

By default, if one of the jobs fails, it will abandon all other active, queued, and pending jobs. This behavior can be controlled with the `continue_on_failure` and `continue_without_deps` options:

 * `continue_on_failure` - Don't abandon all jobs if one job fails.
 * `continue_without_deps` - Don't abandon the dependencies of failed jobs.

Jobs are activated subject to the availability of tokens from the job server. This happens whether or not the jobs erver is being used to manage `make` jobs since the job server treats the executor just like any other job client. Once a job is activated, each of its stages is executed in order, and if one stage fails, the job fails.

#### Getting Feedback from Execution
Like the current executor, the new execution pipeline uses a `Queue` to asynchronously communicate with the console thread. The executor writes [`ExecutionEvent`](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/execution/events.py) objects into the event queue to be processed by any kind of output controller.

The default console controller has several options for output filtering and formatting:

* `show_stage_events` False
* `show_buffered_stdout` False
* `show_buffered_stderr` True
* `show_live_stdout` False
* `show_live_stderr` False
* `show_active_status` True
* `show_full_summary` False
* `active_status_rate` 20.0

#### Creating `catkin build` and `catkin clean` Jobs
Jobs for building and cleaning jobs are created by factory functions defined for [catkin](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/jobs/catkin.py) and [cmake](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/jobs/cmake.py) packages. This makes the job objects task-agnostic, even if they require different types of information, as seen in the case of [`catkin build`](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/verbs/catkin_build/build.py#L299-L318) and [`catkin clean`](https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/verbs/catkin_clean/clean.py#L148-L168).

### Error Display

This PR dramatically simplifies the error output display in order to make errors clearer and less verbose. Since we have more controls to build single packages, it doesn't seem like it's necessary to have the "command to reproduce" and the numerous paths to different logfiles visible by default. On small screens, the actual error is usually even lost up above in the scrollback.

#### Current
current_warnings

#### This PR
cb_new_warnings

### Cleaning Individual Packages

This PR adds jobs for cleaning all files generated by individual catkin and CMake packages in the *build* and *devel* spaces. The "clean" jobs are different from the "build" jobs since they do not technically need to know the source path to packages. In fact, it's desirable to make them only depend on the information in the `devel_manifest.txt` files to be cleaned. This way we can support full orphan removal if a package's source is removed from the workspace.

### TODO

- [x] Properly filter output from "python" stages
- [x] Move `devel_manifest.txt` files into the develspace so they can be cleaned without the build directories
- [x] Move the `install_manifest.txt` files into the develspace so they can be cleaned without the build directories
- [x] Move linked-devel isolated devel spaces out of package build directories
- [x] Support complete orphan removal
- [x] Refactor `catkin clean` pipeline to not require source packages to exist in the source space
- [x] Make `--link-devel` the default behavior and add an option to build with an "unlinked" merged devel space like the current behavior 
- [x] Add load average and memory usage warnings to status bar
- [x] Add file logging back in
- [x] Add notifications back in
- [x] Add `--unbuilt` option to `catkin build` to build any packages which have not yet been built
- [x] Add notice to source the setup file if new packages are built or if packages are removed
- [ ] Add documentation for new additions
- [ ] Make "isolated" and "merged" devel space configs advanced (hidden) options
- [x] Fix completion time display for packages which might not run continuously due to job occupancy  / scheduling (i.e. `cmake` stage runs and completes, then job is idle until the jobserver allows the `make` stage to complete)
- [x] Fix queued package display to include "idle" jobs which are waiting for job tokens after being started (related to the above bullet)
- [ ] Make `--unbuilt` cause any packages which depend on the unbuilt packages to be cleaned completely and rebuilt
- [x] https://github.com/jbohren-forks/catkin_tools/issues/11 Remove all unused imports
- [ ] https://github.com/jbohren-forks/catkin_tools/issues/10 Cover with unit tests
- [x] https://github.com/jbohren-forks/catkin_tools/issues/8 Fix logging streaming behavior
- [x] https://github.com/jbohren-forks/catkin_tools/issues/7 Fix elapsed time display
- [x] https://github.com/jbohren-forks/catkin_tools/issues/6 Clarify load/mem high usage
- [x] https://github.com/jbohren-forks/catkin_tools/issues/5 Fix clean warning check
- [x] https://github.com/jbohren-forks/catkin_tools/issues/4 Fix extra line breaks in captured input
- [x] https://github.com/jbohren-forks/catkin_tools/issues/14 Merge changes from upstream


wjwwood:
> This moves Executor and the functions and classes which the Executor uses into a jobs module. This enables catkin clean (and other potential verbs) to take advantage of it.

I would have preferred doing this separately. It makes it really difficult to review things when you've mixed several large, somewhat unrelated, changes together. In the mean time you could have just imported what you needed from the `Executor` stuff while it remained in the `build` verb.

> I think this sort of per-package clean functionality is the final big feature preventing catkin_tools from being complete.

I must have missed that road map :stuck_out_tongue:.

jbohren:
> I would have preferred doing this separately. It makes it really difficult to review things when you've mixed several large, somewhat unrelated, changes together. In the mean time you could have just imported what you needed from the Executor stuff while it remained in the build verb.

The changes aren't hard to separate, and I can do that when this is ready for an in-depth review. The only reason why they're all together is that it was the quickest way to figure out what needed to move in order to make the clean functionality happen.

>>I think this sort of per-package clean functionality is the final big feature preventing catkin_tools from being complete.

>I must have missed that road map :stuck_out_tongue:.

Meaning this is the last big feature that would enable me to spend as little time compiling code in my day-to-day work as I did with `rosbuild` :metal:

wjwwood:
> The changes aren't hard to separate, and I can do that when this is ready for an in-depth review. The only reason why they're all together is that it was the quickest way to figure out what needed to move in order to make the clean functionality happen.

That makes sense; I get that it is easier to develop like this. Like you said, we can split it up when it's time for review.

jbohren:
@wjwwood So I added a couple functions to support symbolically linking isolated devel files into a merged devel space and creating a `devel_manifest.txt` similar to CMake's `install_manifest.txt`. I also created an experimental `--link-devel` argument to `catkin configure` , which right now simply builds into isolated develspaces contained in each package's build directory. I still need to incorporate the functions which actually create the symlinks.

What needs to be done now to support this in practice, is generating a merged devel space from a collection of isolated devel spaces. Besides copying all of the non-conflicting files, there are four files which are present in each devel space which would need to be properly merged or modified. Tell me if you think this is a correct assessment, and if you think it's reasonable for `catkin_tools` to be generating these files in the merged devel space or if `catkin_tools` should be leaning on Catkin for doing it.

* `.catkin`
 * Contains path to single package source space
 * Should paths to all source spaces
* `etc/catkin/profile.d/05.catkin-test-results.sh`
 * Contains path to single package's build space 
 * Should contain path to the workspace's build space (or something?)
 * Related to #198
* `.rosinstall`
 * Contains path to single package devel space
 * Should contain path to the merged devel space
* `setup.sh`
 * Contains path to single package devel space
 * Should contain path to the merged devel space

jbohren:
@wjwwood Check out the current design details in the updated PR description. I think it's actually converging to something that will work well in practice.

davetcoleman:
+1 i've always wanted this feature. i haven't looked at the actual code, though

wjwwood:
Thanks for working on this @jbohren, it's on my list to review, and I'll get to it asap.

scpeters:
One caveat about using `install_manifest.txt` for vanilla cmake packages: this lists only installed files, not folders, which are automatically created as needed by the installer. Deleting the files in the `install_manifest.txt` won't be a complete clean if there are otherwise empty folders left behind. We noticed this behavior when implementing a [make uninstall target in gazebo](https://bitbucket.org/osrf/gazebo/issue/258/make-uninstall-only-removes-files) but haven't tried to fix it.

It's not a deal-breaker, and some people might not even notice, but it's worth keeping in mind.

jbohren:
> One caveat about using install_manifest.txt for vanilla cmake packages: this lists only installed files, not folders, which are automatically created as needed by the installer. Deleting the files in the install_manifest.txt won't be a complete clean if there are otherwise empty folders left behind. We noticed this behavior when implementing a make uninstall target in gazebo but haven't tried to fix it.

This PR does in fact clean empty folders:
https://github.com/jbohren-forks/catkin_tools/blob/clean-pkgs/catkin_tools/jobs/cmake.py#L248-L274

scpeters:
Nice!

jbohren:
@wjwwood I've integrated my experimental asyncio hacking into this branch. This still includes the per-package clean support and symlinked devel support. I still need to flesh out the rest of the logging, remove the old execution stuff, update the cli interface, and update both the PR design details and the catkin_tools documentation, but the tests are passing now.

adolfo-rt:
This enhancement is great, but having `catkin clean --orphans` is just awesome.

jbohren:
@wjwwood This is ready for review. It still needs unit tests, but it has all the features that I think make the partial workspace cleaning / rebuilding complete.

wjwwood:
@jbohren Thanks for working on it. We've been busy with the DRC, but we're back from that now and I'll try to spend some time of catkin tools this week. It might be the later part of the week depending on follow ups from the DRC. Thanks for your patience and you work on these features.

jbohren:
> @jbohren Thanks for working on it. We've been busy with the DRC, but we're back from that now and I'll try to spend some time of catkin tools this week. It might be the later part of the week depending on follow ups from the DRC. Thanks for your patience and you work on these features.

Yeah, one of the biggest things is the "linked" develspace, which I feel like should just be the default configuration. It takes the best aspects of the merged and isolated devel setups since it builds everything isolated, and then merges them all into a single develspace with symlinks. This means you don't have issues with setup files like isolated devel spaces, but you can still separate the products package-by-package.

wjwwood:
I've been reviewing this piece by piece over the weekend. It's gonna take some time for me to get to all of it. But the changes are substantial and I think I owe them a thorough review. I just wanted to let you know I was working on it.

jbohren:
> I've been reviewing this piece by piece over the weekend. It's gonna take some time for me to get to all of it. But the changes are substantial and I think I owe them a thorough review. I just wanted to let you know I was working on it.

Cool. All I've been doing is adding / cleaning up unit tests.

wjwwood:
@jbohren I fixed a bug on the master branch and now this doesn't merge. It should be easy to rebase since you basically overwrite the section I changed in master, but it does need to be rebased.

catkin_tools/verbs/catkin_build/build.py
```diff
@@ -21,16 +21,12 @@
 import time
 import yaml
```
wjwwood: `import operator` is no longer used (a few lines above).

catkin_tools/execution/stages.py
```diff
@@ -0,0 +1,83 @@
...
+            'env': env,
+            'shell': shell,
+            # Emulate tty for cli colors
+            'emulate_tty': emulate_tty,
+            # Capture stderr and stdout separately
+            'stderr_to_stdout': stderr_to_stdout
+        }
+
+
+class FunStage(Stage):
```
wjwwood: Can you rename this `FunctionStage` or `FuncStage`, `FunStage` is confusing.

wjwwood:
There are tonnes of unused imports throughout the code. If I get a chance I'll make a pull request to remove them.

wjwwood:
I've also noticed that the output from warnings (maybe stderr, maybe all captured output) has extra line breaks in them. For example:

```
C
M
ake Warning (dev):
  Policy C
MP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.
  Use the cmake_policy command to
  set the policy and suppress this warning.
  MACOSX_RPATH is not specified
for the following targets:
   dart-core
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Whereas I would expect it to look like this:

```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake --help-policy CMP0042" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.
  MACOSX_RPATH is not specified for the following targets:
   dart-core
This warning is for project developers.  Use -Wno-dev to suppress it.
```

wjwwood:
With `catkin clean --all`, if you put something other than `y` for the answer it just exist with no notice. I would expect that it either:

- Tells you what you put and says it is doing nothing.
- Tells you what you put and offers y/n again.

I think that only `n`, `N`, or `<enter>` should exit directly. And regardless it should tell you that it is not deleting anything.

Example:

```
% catkin clean --all
y[clean] Warning: This will completely remove the existing build, devel, and install spaces for this profile:
[clean]   Build Space:   /Users/william/gazebo_ws/build
[clean]   Devel Space:   /Users/william/gazebo_ws/devel
[clean] Note: Use `--force` to skip this check.

[clean] Are you sure you want to completely remove the directories listed above? [yN]: y
```

That first `y` was me pushing it before it got a chance to ask the question. But based on the output after I answered the question I would have guessed it went ahead and deleted stuff.

wjwwood:
How is the user supposed to understand what `[Load too high]` means? Is there any documentation on that part of UI? Wouldn't `[High load]`/`[High memory usage]` be more appropriate, with documentation that says "When these messages are seen, the tool is limiting the start of new make jobs to prevent overloading the system."?

wjwwood:
This question is a matter of preference (maybe it should be configuration), but why did you move the `[x/y complete] [x/y jobs] [x queued]` block of the status line away from the right hand side? I rather liked it that way. It's equally space efficient and makes it easier to read IMO. Like I said it's a preference, but I'd like to know why you thought this was better.

wjwwood:
Another note, the time being displayed next to each package is the time of the current step (cmake, make, make install). It would be my preference to show the total time. Again, this is a preference which we should probably make configurable.

wjwwood:
I need to do some more investigating, but it seems like log files are not created until the stage of the build is completed. This makes it impossible to tail the log files. Depend on how its implemented (still looking at that) it could also mean that all of the console output is being buffered in memory before being written out to file all at once. I had explicitly avoided that in the current master branch to avoid using lots of memory during a build. It has the side benefit of allowing the user to tail a log file as it is being filled if they want to introspect a build but don't want to do `-vi` or similar.

wjwwood:
While testing it out I also got this traceback:

```
  File "/usr/local/bin/catkin", line 9, in <module>
    load_entry_point('catkin-tools==0.3.0', 'console_scripts', 'catkin')()
  File "/Library/Python/2.7/site-packages/catkin_tools/commands/catkin.py", line 219, in main
  File "/Library/Python/2.7/site-packages/catkin_tools/verbs/catkin_build/cli.py", line 337, in main
    summarize_build=opts.summarize  # Can be True, False, or None
  File "/Library/Python/2.7/site-packages/catkin_tools/verbs/catkin_build/build.py", line 334, in build_isolated_workspace
    context, pkg, pkg_path, context.devel_space_abs)
  File "/Library/Python/2.7/site-packages/catkin_tools/jobs/catkin.py", line 604, in create_catkin_tools_bootstrap_job
    cwd=build_space))
TypeError: __init__() got an unexpected keyword argument 'command'
```

Looking into it now.

catkin_tools/execution/stages.py
jbohren: Renamed `FunStage` to `FunctionStage` and `CmdStage` to `CommandStage` for consistency in 0f61d73

catkin_tools/verbs/catkin_build/build.py
jbohren: Removed in 0f61d73

jbohren:
@wjwwood I've created issues to track all of your comments over in my fork https://github.com/jbohren-forks/catkin_tools/milestones/Merge%20clean-pkgs

jbohren:
@wjwwood Are there any more outstanding issues with the current PR? Before this gets finalized, I'd like to discuss the merits of isolated / merged / linked develspaces. Other than that, all it needs is a few more unit tests and a bunch of new documentation.

wjwwood:
I haven't finished my review so I don't know if there are other things to be done; I'll try to spend some time on it tomorrow.

jbohren:
> I haven't finished my review so I don't know if there are other things to be done; I'll try to spend some time on it tomorrow.

Alright. Also, since this relies on `osrf_pycommon` could you include https://github.com/osrf/osrf_pycommon/pull/10 as part of that review? It turns out some changes in python's treatment of file descriptors is what was breaking the jobserver on Python 3.4.

jbohren:
@wjwwood Any other comments on this PR?

wjwwood:
I don't have any more comments just yet, but I'm still not done reviewing it. The change set is enormous, and so without splitting them up it will take a long time for me to review it to the point of "ok to to merge". I'll probably catch up on it this weekend, maybe Sunday.

jbohren:
> I don't have any more comments just yet, but I'm still not done reviewing it. The change set is enormous, and so without splitting them up it will take a long time for me to review it to the point of "ok to to merge".

Yeah, I'm not expecting that this is ready to be merged yet. The biggest change in this that I'd like to get some feedback on, though, is the linked develspace behavior / design. My only concern is that you're spending time reviewing line-by-line changes before accepting the high-level design decisions.

wjwwood:
Well I'm doing three things as I have time:

- Use the tool, which is where the UX comments I had came from.
- Review code, I only do this when I dig in to understand why something else is happening.
- Review the design, I've not had time to sit down and go through those yet.

I'll try to focus on the high level design stuff in my next session.

jbohren:
> Well I'm doing three things as I have time:
> I'll try to focus on the high level design stuff in my next session.

Sounds good!

wjwwood:
More notes (partially to myself):

The command to reproduce a failing package should not include the `--jobserver=3,4` arguments and does not pass the appropriate `-j` argument, for example if I do `-j1` then the reproduce command just has `-j` instead.

wjwwood:
More notes from starting to use this again (I'm going to do the high level review tonight hopefully):

- I have to have this patch in order to build anything:

```
diff --git a/catkin_tools/jobs/catkin.py b/catkin_tools/jobs/catkin.py
index e2a1c14..1ac180f 100644
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -530,7 +529,7 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
     if context.install:
         stages.append(CommandStage(
             'install',
-            command=env_prefix + [MAKE_EXEC, 'install'],
+            env_prefix + [MAKE_EXEC, 'install'],
             cwd=build_space))
 
     return Job(

```

- I've had it hang on me a few times and I've had to do `ctrl-\` to get out of it:

```
Finished  <<< rosparam                      [ 4.8 seconds ]
Starting  >>> rospack
_________________________________________________________________________________________________________________________________________________________________________________
Warnings   << rosservice:install /tmp/jade_catkin_tools/build/_logs/rosservice/build.install.000.log
+ '[' -n '' ']'
+ cd /tmp/jade_catkin_tools/src/ros_comm/rosservice
+ /usr/bin/env PYTHONPATH=/tmp/jade_catkin_tools/install/lib/python2.7/site-packages:/tmp/jade_catkin_tools/build/rosservice/lib/python2.7/site-packages: CATKIN_BINARY_DIR=/tmp/jade_catkin_tools/build/rosservice /usr/bin/python /tmp/jade_catkin_tools/src/ros_comm/rosservice/setup.py build --build-base /tmp/jade_catkin_tools/build/rosservice install --prefix=/tmp/jade_catkin_tools/install --install-scripts=/tmp/jade_catkin_tools/install/bin
.................................................................................................................................................................................
Finished  <<< rosservice                    [ 4.3 seconds ]
Starting  >>> gencpp






^C^C^C^C^\[1]    22679 quit       catkin build
```

Every package build outputs a warning:

```
_________________________________________________________________________________________________________________________________________________________________________________
Warnings   << rosservice:install /tmp/jade_catkin_tools/build/_logs/rosservice/build.install.000.log
+ '[' -n '' ']'
+ cd /tmp/jade_catkin_tools/src/ros_comm/rosservice
+ /usr/bin/env PYTHONPATH=/tmp/jade_catkin_tools/install/lib/python2.7/site-packages:/tmp/jade_catkin_tools/build/rosservice/lib/python2.7/site-packages: CATKIN_BINARY_DIR=/tmp/jade_catkin_tools/build/rosservice /usr/bin/python /tmp/jade_catkin_tools/src/ros_comm/rosservice/setup.py build --build-base /tmp/jade_catkin_tools/build/rosservice install --prefix=/tmp/jade_catkin_tools/install --install-scripts=/tmp/jade_catkin_tools/install/bin
.................................................................................................................................................................................
```

I've ticked that here: https://github.com/ros/catkin/pull/755

wjwwood:
When building `ros_comm` for jade I get this:

```
  File "/tmp/jade_catkin_tools/install/share/genmsg/cmake/../../../lib/genmsg/genmsg_check_deps.py", line 41, in <module>
    from genmsg import EXT_MSG, EXT_SRV, MsgContext
ImportError: No module named genmsg
make[2]: *** [CMakeFiles/_std_msgs_generate_messages_check_deps_UInt8] Error 1
make[1]: *** [CMakeFiles/_std_msgs_generate_messages_check_deps_UInt8.dir/all] Error 2
make: *** [all] Error 2
```

I am using `catkin config --install` and I'm on OS X, but I do not get this problem with `catkin_make_isolated --install`. I can also confirm that the `genmsg` Python package is installed, but I haven't looked into the `PYTHONPATH` at that point in the build yet.

wjwwood:
There seems to be a race condition with all jobs finishing and the printing of the build summary, which results in interleaved notices and the summary:

```
Starting  >>> roslaunch
_________________________________________________________________________________________________________________________________________________________________________________
Errors     << std_srvs:make /tmp/jade_catkin_tools/build/_logs/std_srvs/build.make.000.log
  File "/tmp/jade_catkin_tools/install/share/genmsg/cmake/../../../lib/genmsg/genmsg_check_deps.py", line 41, in <module>
    from genmsg import EXT_MSG, EXT_SRV, MsgContext
ImportError: No module named genmsg
make[2]: *** [CMakeFiles/_std_srvs_generate_messages_check_deps_Trigger] Error 1
make[1]: *** [CMakeFiles/_std_srvs_generate_messages_check_deps_Trigger.dir/all] Error 2
make: *** [all] Error 2
.................................................................................................................................................................................
Failed     << std_srvs:make                 [ Exited with code 2 ]
Failed    <<< std_srvs                      [ 3.8 seconds ]
Abandoned <<< rosconsole                    [ Unrelated job failed ]
Abandoned <<< roslz4                        [ Unrelated job failed ]
Abandoned <<< rosbag_storage                [ Unrelated job failed ]
Abandoned <<< rostest                       [ Unrelated job failed ]
Abandoned <<< rosgraph_msgs                 [ Unrelated job failed ]
Abandoned <<< roscpp                        [ Unrelated job failed ]
Abandoned <<< rosout                        [ Unrelated job failed ]
Abandoned <<< message_filters               [ Unrelated job failed ]
Abandoned <<< rosnode                       [ Unrelated job failed ]
Abandoned <<< rostopic                      [ Unrelated job failed ]
Abandoned <<< roswtf                        [ Unrelated job failed ]
Abandoned <<< topic_tools                   [ Unrelated job failed ]
Abandoned <<< rosbag                        [ Unrelated job failed ]
_________________________________________________________________________________________________________________________________________________________________________________
Errors     << std_msgs:make /tmp/jade_catkin_tools/build/_logs/std_msgs/build.make.000.log
  File "/tmp/jade_catkin_tools/install/share/genmsg/cmake/../../../lib/genmsg/genmsg_check_deps.py", line 41, in <module>
    from genmsg import EXT_MSG, EXT_SRV, MsgContext
ImportError: No module named genmsg
make[2]: *** [CMakeFiles/_std_msgs_generate_messages_check_deps_Bool] Error 1
make[1]: *** [CMakeFiles/_std_msgs_generate_messages_check_deps_Bool.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
  File "/tmp/jade_catkin_tools/install/share/genmsg/cmake/../../../lib/genmsg/genmsg_check_deps.py", line 41, in <module>
    from genmsg import EXT_MSG, EXT_SRV, MsgContext
ImportError: No module named genmsg
make[2]: *** [CMakeFiles/_std_msgs_generate_messages_check_deps_UInt8] Error 1
make[1]: *** [CMakeFiles/_std_msgs_generate_messages_check_deps_UInt8.dir/all] Error 2
make: *** [all] Error 2
.................................................................................................................................................................................
Failed     << std_msgs:make                 [ Exited with code 2 ]
Failed    <<< std_msgs                      [ 4.2 seconds ]
_________________________________________________________________________________________________________________________________________________________________________________
Warnings   << roslib:cmake /tmp/jade_catkin_tools/build/_logs/roslib/build.cmake.000.log
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   roslib

This warning is for project developers.  Use -Wno-dev to suppress it.

.................................................................................................................................................................................
_________________________________________________________________________________________________________________________________________________________________________________
Warnings   << roslaunch:install /tmp/jade_catkin_tools/build/_logs/roslaunch/build.install.000.log
+ '[' -n '' ']'
+ cd /tmp/jade_catkin_tools/src/ros_comm/roslaunch
+ /usr/bin/env PYTHONPATH=/tmp/jade_catkin_tools/install/lib/python2.7/site-packages:/tmp/jade_catkin_tools/build/roslaunch/lib/python2.7/site-packages: CATKIN_BINARY_DIR=/tmp/jade_catkin_tools/build/roslaunch /usr/bin/python /tmp/jade_catkin_tools/src/ros_comm/roslaunch/setup.py build --build-base /tmp/jade_catkin_tools/build/roslaunch install --prefix=/tmp/jade_catkin_tools/install --install-scripts=/tmp/jade_catkin_tools/install/bin
.................................................................................................................................................................................
Finished  <<< roslaunch                     [ 3.8 seconds ]
_________________________________________________________________________________________________________________________________________________________________________________
Warnings   << roslib:install /tmp/jade_catkin_tools/build/_logs/roslib/build.install.000.log
+ '[' -n '' ']'
+ cd /tmp/jade_catkin_tools/src/ros/roslib
+ /usr/bin/env PYTHONPATH=/tmp/jade_catkin_tools/install/lib/python2.7/site-packages:/tmp/jade_catkin_tools/build/roslib/lib/python2.7/site-packages: CATKIN_BINARY_DIR=/tmp/jade_catkin_tools/build/roslib /usr/bin/python /tmp/jade_catkin_tools/src/ros/roslib/setup.py build --build-base /tmp/jade_catkin_tools/build/roslib install --prefix=/tmp/jade_catkin_tools/install --install-scripts=/tmp/jade_catkin_tools/install/bin
.................................................................................................................................................................................
Finished  <<< roslib                        [ 6.6 seconds ]
[build] Runtime: 21.8 seconds total.
[build] Summary: 32 of 47 jobs completed.
[build]   Warnings:  22 completed jobs produced warnings.
[build]   Abandoned: 13 jobs were abandoned.
[build]   Failed:    2 jobs failed.
[build] Note: Workspace packages have changed, please re-source setup files to use them.
```

Also the use of the term `Abandoned` isn't very obvious to me here. Maybe `skipped` or `not attempted` would be more appropriate.

wjwwood:
Another note for myself: When the build fails it just reports that a certain number completes with warnings, but does not mention the failure (in the notification).

wjwwood:
Another point for me to look into:

Building `ros_comm` for jade I get this error:

```
Errors     << std_srvs:make /Users/william/jade/build/_logs/std_srvs/build.make.000.log
  File "/Users/william/jade/install/share/genmsg/cmake/../../../lib/genmsg/genmsg_check_deps.py", line 41, in <module>
    from genmsg import EXT_MSG, EXT_SRV, MsgContext
ImportError: No module named genmsg
  File "/Users/william/jade/install/share/genmsg/cmake/../../../lib/genmsg/genmsg_check_deps.py", line 41, in <module>
    from genmsg import EXT_MSG, EXT_SRV, MsgContext
ImportError: No module named genmsg
make[2]: *** [CMakeFiles/_std_srvs_generate_messages_check_deps_Trigger] Error 1
make[1]: *** [CMakeFiles/_std_srvs_generate_messages_check_deps_Trigger.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[2]: *** [CMakeFiles/_std_srvs_generate_messages_check_deps_Empty] Error 1
make[1]: *** [CMakeFiles/_std_srvs_generate_messages_check_deps_Empty.dir/all] Error 2
make: *** [all] Error 2
```

But it does not happen with `catkin_make_isolated`.

NikolausDemmel:
I feel like with more and more people using catkin tools, this getting this PR in would be another big leap forward. I understand both Johnathan and William have little time for catkin tools these days. I might have a bit of spare time at my hands the coming days. What would be the best way to support the efforts in the PR at this stage?

P.S.: Does anyone know a better tool for code-reviewing large change-sets such as in this PR? github is really a bit limited here...

scpeters:
@jbohren initially said he would split this into smaller, atomic pull requests; I think that is the best way forward. Maybe he could advise on how things could be split?

jbohren:
> What would be the best way to support the efforts in the PR at this stage?

@NikolausDemmel, I think the best thing to do would be to try to resolve @wjwwood's specific bugs, and to flesh out the rest of the unit tests so that there's more coverage of the existing and new features.

> @jbohren initially said he would split this into smaller, atomic pull requests; I think that is the best way forward. Maybe he could advise on how things could be split?

@scpeters, I don't think I was ever going to split this into smaller PRs, and I don't think refactoring it for that purpose would be a good use of time. This PR changes a ton of things, and I'm not even sure if @wjwwood has approved all of the _design_ decisions let alone the implementation details.

scpeters:
@jbohren the [first comment on this thread](https://github.com/catkin/catkin_tools/pull/196#issuecomment-92466595) was by @wjwwood, a portion of which says:

~~~
I would have preferred doing this separately.
It makes it really difficult to review things
when you've mixed several large, somewhat unrelated, changes together.
~~~

to which [you replied](https://github.com/catkin/catkin_tools/pull/196#issuecomment-92472494):

~~~
The changes aren't hard to separate,
and I can do that when this is ready for an in-depth review.
The only reason why they're all together is that
it was the quickest way to figure out what needed
to move in order to make the clean functionality happen.
~~~

I interpreted that to mean that this PR would be split into smaller ones.

It may not be ready for refactoring, but before merging I think it is absolutely worth it to refactor into smaller pull requests, since it is much harder to identify defects in large blocks of code based on my experience reviewing large pull requests in Gazebo.

wjwwood:
Splitting the pr up would certainly help, but @jbohren has a point in that I haven't commented on all of the design level decisions (separate from code review, automated test review, and acceptance testing).

I've looked at the smylinked devel space overview a few times now, and I think that it makes sense. The part I always get caught up in is that I want to test out some of the assertions on how it should behave and check some things (acceptance testing) and I get hung up on other issues (some mine and some with the pr). I haven't had time to dig into the Executor changes because they are more intrusive and the important design changes are coupled to the code changes. This is what is hard for me to review because I need a large block of time to keep it all in my mind for an extended period of time to verify that it is sound. I'd be prepared to assume @jbohren made the right decisions in the changes to the executor pattern, but in my acceptance testing I've run into a few suspicious behaviors which I've tried to quickly jot down in comments here so I can follow up on them later.

I'll try to find some time to work on this for a long enough contiguous block of time to follow through on this pr, but realistically it might not happen until mid-late December.

jbohren:
> I interpreted that to mean that this PR would be split into smaller ones.

@scpeters Yeah those comments do imply that, but at that point it was a much smaller PR. At this point, I don't think it's worth it to refactor into a bunch of atomic features.

jeffeb3:
This PR has some great stuff in it. The thing I'm most interested in is the warnings/stderr parsing. I've installed this PR, and I'm testing/working on it.

I installed using pip, and it completed successfully, but the first run was missing osrf_pycommon. pip didn't know how to find osrf_pycommon, and it's not in my 14.04 version of ubuntu, so I had to install that too. So there's a problem there.

I understand this is a huge PR, with many different aspects to the problem, but the thing that really makes my life easier is the fact that warnings (from cmake, make, and gcc) are all very clear, and I don't have to see everything from -v. I get so much help from warnings, this is great. I'll help the only way I can, really, which is to test it, and report or maybe fix minor bugs. This this is a great update.

scpeters:
@jbohren: in my opinion, if the PR is now bigger, that makes it even more desirable to split it up into smaller chunks to improve the likelihood of finding bugs during the review process.

I'm not trying to be argumentative; I'm just explaining why I asked if you could comment on how the PR might be split up.

davetcoleman:
@jbohren I am testing this pull request because I want to be able to remove individual packages without rebuilding my whole work space. After switching to your branch and running ``sudo python setup.py install`` catkin no longer works & I have an error. I am on ROS Jade and I get:

```
$ catkin build
  File "/usr/local/bin/catkin", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 2749, in <module>
    working_set = WorkingSet._build_master()
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 444, in _build_master
    ws.require(__requires__)
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 725, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 628, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: trollius
```

Thoughts? Thanks!

wjwwood:
@davetcoleman you must install Trollius (it is a Python2.7 compatible implementation of the Python3's asyncio). This should get you going:

```
$ sudo -H python2.7 -m pip install -U trollius
```

davetcoleman:
Thanks. Now I get:

> ImportError: No module named osrf_pycommon.process_utils

It looks like osrf_pycommon is only available [from soruce](https://osrf-pycommon.readthedocs.org/en/latest/)? I tried using a similar command to the one your just provided.

wjwwood:
Yeah, you'll have to install that from source for now: https://osrf-pycommon.readthedocs.org/en/latest/#installing-from-source

jeffeb3:
Yes, its only from source AFAIK. Use pip from github:
http://stackoverflow.com/questions/8247605/configuring-so-that-pip-install-can-work-from-github
With the address pointing to the osrf_pycommon repo:
https://github.com/osrf/osrf_pycommon

jeffeb3:
Or that will work too.

catkin_tools/jobs/catkin.py
```diff
@@ -0,0 +1,668 @@
...
+    # Create an environment file
+    env_file_path = get_env_file_path(package, context)
+    stages.append(FunctionStage(
+        'envgen',
+        create_env_file,
+        package=package,
+        context=context,
+        env_file_path=env_file_path))
+
+    # Only use it for building if the develspace is isolated
```
xqms: @jbohren Can you explain the motivation here? This breaks building from scratch for me, e.g.
```
# Clean workspace (just src folder)
. /opt/ros/indigo/setup.bash
catkin build
```

This fails for packages including message headers built in other packages, since `devel/include` will not be included in the header search path. Indeed, how should a package know about the shared devel space if not through the env file?

After the first build fails, I can source the then-available `devel/setup.bash` and a second build will succeed.

If I use `env_prefix = [env_file_path]` unconditionally here, the `devel/include` is added to `CPATH` and everything works on the first build.

NikolausDemmel:
For everyones reference: I have some minor tweaks to this PR here: https://github.com/jbohren-forks/catkin_tools/pull/17

catkin_tools/jobs/catkin.py
jbohren: @xqms Yeah, this is a bug in the current PR. The original intention was to avoid re-sourcing the same setup files, but I left this in by mistake.

xqms: Hm, you report a great speedup when avoiding re-sourcing the `env.sh` files above. Can we somehow keep that speedup? Maybe source one env.sh at the beginning of the build?

jbohren: So we can definitely avoid sourcing it with every command by only sourcing it once per package and saving the environment. We can also go a step further and bypass the environment files by sourcing the workspace setup file directl. The ultimate would be to only source it once per build and then re-source it only when the catkin env hooks change.

NikolausDemmel:
@jbohren, in the `Context` we now have `isolate_devel` and `link_devel`. It is unclear to me how they should affect each other and which one should take precedence. In the options we have `--isolate-devel` and `--merge-devel`, setting `isolate_devel` to `True` and `False`, respectively and `--link-devel` setting `link_devel` to `True` (which is also the default and therefore cannot currently be disabled). Would it make sense to rather have a single variable with 3 different possible values `link`, `merge`, `isolate` instead of these two booleans?

jbohren:
@NikolausDemmel Yeah, the interface should definitely be more enum-like. Currently there is no "precedence" they are all just mutually exclusive.

NikolausDemmel:
@jbohren @wjwwood any preferences on how enum style variables should be implemented? Using the Python 3.4 enum (there are pip-installable backports for earlier versions), or just something like a string?

catkin_tools/jobs/catkin.py
jbohren: @xqms Take a look at #248 (which is #196 without linked devel or per-package cleaning). It does "smart" env loading and caching based on the env hooks.

jbohren:
I'm going to close this now that all of the contributions here have either already been merged or are represented in #250 and #293.

PR #204, opened 2015-05-01
Catch KeyError when SHELL is not set

### LINKED ISSUE #203: catkin build fails if SHELL is not set
I was building some catkin packages using docker, which uses root by default. For some reason, `SHELL` was not set in the environment, which was causing `catkin build` to fail:
~~~
  File "/usr/bin/catkin", line 9, in <module>
    load_entry_point('catkin-tools==0.3.0', 'console_scripts', 'catkin')()
  File "/usr/lib/python2.7/dist-packages/catkin_tools/commands/catkin.py", line 219, in main
  File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/cli.py", line 229, in main
    ctx = Context.load(opts.workspace, opts.profile, opts, append=True)
  File "/usr/lib/python2.7/dist-packages/catkin_tools/context.py", line 168, in load
    ctx.load_env()
  File "/usr/lib/python2.7/dist-packages/catkin_tools/context.py", line 300, in load_env
    sticky_env = get_resultspace_environment(self.devel_space_abs, quiet=True)
  File "/usr/lib/python2.7/dist-packages/catkin_tools/resultspace.py", line 56, in get_resultspace_environment
    shell_path = os.environ['SHELL']
  File "/usr/lib/python2.7/UserDict.py", line 23, in __getitem__
    raise KeyError(key)
KeyError: 'SHELL'
~~~
The error can easily be reproduced with bash by calling `unset SHELL` before running `catkin build`.

There should be some error checking to see if the `SHELL` variable exists before trying to use it.

scpeters:
Fixes #203.

### DIFF
--- a/catkin_tools/resultspace.py
+++ b/catkin_tools/resultspace.py
@@ -53,7 +53,7 @@ def get_resultspace_environment(result_space_path, quiet=False, cached=True):
         )
 
     # Determine the shell to use to source the setup file
-    shell_path = os.environ['SHELL']
+    shell_path = os.environ.get('SHELL', '')
     (_, shell_name) = os.path.split(shell_path)
 
     # Use fallback shell if using a non-standard shell


catkin_tools/resultspace.py
```diff
@@ -53,7 +53,10 @@ def get_resultspace_environment(result_space_path, quiet=False, cached=True):
         )
 
     # Determine the shell to use to source the setup file
-    shell_path = os.environ['SHELL']
+    try:
+        shell_path = os.environ['SHELL']
+    except KeyError:
+        shell_path = ''
```
wjwwood: I think it would be better to use `get`:

```python
shell_path = os.environ.get('SHELL', '')
```

wjwwood:
Fix looks good, minor implementation comment, and then +1

catkin_tools/resultspace.py
scpeters: fixed and squashed

wjwwood:
Thanks.

PR #206, opened 2015-05-26
augmented bash completion, fixed installed file permissions

rhaschke:


setup.py
```diff
@@ -50,11 +52,23 @@
...
     ['completion/catkin_tools-completion.bash']))
 
+
+class PermissiveInstall(install):
+    def run(self):
+        install.run(self)
+        if os.name == 'posix':
+            for file in self.get_outputs():
+                # all installed files should be readable for anybody
+                mode = ((os.stat(file)[ST_MODE]) | 0444) & 07777
```
dirk-thomas: This line is not valid Python 3 syntax. You have to write octal number in the following style: `0o1234`

wjwwood:
lgtm, but could you expand on what this didn't work for you before?

rhaschke:
* The frequently used config options --whitelist, --blacklist, --install were not yet supported by bash completion.
* installing with `sudo python setup.py` left the installed files with read permission for root only on my Ubuntu 14.04

Both issues were fixed.

wjwwood:
Interesting, ok thanks.

PR #209, opened 2015-06-03
allow completion for several packages after --whitelist or --blacklist

rhaschke:
This is an extension to #206 allowing bash completion not only for the first but for any number of packages after --whitelist or --blacklist.

### DIFF
--- a/completion/catkin_tools-completion.bash
+++ b/completion/catkin_tools-completion.bash
@@ -17,6 +17,17 @@ if [[ -n ${ZSH_VERSION-} ]]; then
   autoload -U +X bashcompinit && bashcompinit
 fi
 
+_last_option()
+{
+  # search backwards for the last given option
+  for (( i=${COMP_CWORD} ; i > 0 ; i-- )) ; do
+    if [[ ${COMP_WORDS[i]} == -* ]]; then
+      echo ${COMP_WORDS[i]}
+      return
+    fi
+  done
+}
+
 # TODO:
 # - parse --workspace and --profile options in order to complete outside of cwd
 # - autocomplete build options
@@ -61,7 +72,7 @@ _catkin()
         COMPREPLY=($(compgen -W "$(catkin --no-color list --unformatted --quiet)" -- ${cur}))
       fi
     elif [[ "${COMP_WORDS[@]}" == *" config"* ]] ; then
-      if [[ "--whitelist --blacklist" == *${prev}* && ${cur} != -* ]] ; then
+      if [[ ${cur} != -* && "--whitelist --blacklist" == *$(_last_option)* ]] ; then
         COMPREPLY=($(compgen -W "$(catkin --no-color list --unformatted --quiet)" -- ${cur}))
       else
         COMPREPLY=($(compgen -W "${catkin_config_opts}" -- ${cur}))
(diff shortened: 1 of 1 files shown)


jbohren:
@rhaschke Is it possible to do this without having to search backwards through the arglist?

rhaschke:
If you have any idea to do it without backwards search, I'm curious to learn about it.
IMHO we need to search backwards for the latest option string. Otherwise we cannot know the option context. Note, that the search is only performed in case of `catkin config`. 
Using "forward search", e.g. checking the third element of the argument list to be --whitelist or --blacklist, will not scale to an arbitrary number of additional options.

jbohren:
Yeah, I know what you mean. Well, I like this enhancement, but haven't tried it out yet. In the future, it'd be cool to use that `_last_option` function for more places in the autocompletion code.

rhaschke:
Yes, that's why I decided for a generic _last_option() function.

rhaschke:
I rewrote the auto-completion script to parse options from catkin's help output itself.
This way, completion is always up-to-date and complete. The code follows the one in `/etc/bash_completion` that defines a generic `_longopt` function to do help-based completion.
Also, the `_init_completion` function comes from there, allowing standard completion for shell variables and redirection.

jbohren:
> I rewrote the auto-completion script to parse options from catkin's help output itself.

Cool! How's the speed when doing that? Would it be worth pre-compiling any of the completion options?

rhaschke:
I didn't notice any performance issues. Try yourself ;-)
Thinking about caching or precompiling options, I'm actually not sure how to achieve that. You would need to store them locally on the file system, because bash completion is triggered on the fly without any memory about the past, isn't it?

wjwwood:
Would it be better to have a mode where the `catkin` command dumps a more readable version of the args, for example I think this is how this works:

https://argcomplete.readthedocs.org/en/latest/

rhaschke:
Interesting link. It seems to provide pretty much everything needed.
However, this requires (again) a rewrite of the code (and the additional python package to be installed). As I am happy with the proposed patch, I will not put more effort into this for now.

jbohren:
@rhaschke Thanks for these contributions!

I'll test it out and @wjwwood and I will pick out the parts that meet all of our needs.

wjwwood:
Several of these commits have been picked to master, I'm gonna close this for now. Please open a new pr if there are still changes that need to be made.

NikolausDemmel:
@wjwwood: Maybe I'm missing something, but I only see the original #206 in master, not the extension to complete multiple packages.

wjwwood:
@NikolausDemmel oh you might be right.

jbohren:
@rhaschke The current state of this PR doesn't work on `zsh` due to the use of `_init_completion`

rhaschke:
If there is a conflict of catkin_tools-completion.bash with zsh, then we should have a separate completion script for zsh, shouldn't we? As I don't use zsh, I cannot contribute here.

jbohren:
Right now the completion works in both shells, `zsh` does have some more advanced completion options, but right now it's easier to maintain a single completion script.

jbohren:
@rhaschke I added these commits to #362, which also adds a zsh-specific completion script.

PR #214, opened 2015-06-26
adding check to prevent tool cross talk

wjwwood:
See: https://github.com/ros/catkin/issues/742

### DIFF
--- a/catkin_tools/verbs/catkin_build/cli.py
+++ b/catkin_tools/verbs/catkin_build/cli.py
@@ -143,7 +143,7 @@ the --save-config argument. To see the current config, use the
         help='Adds a build summary to the end of a build; defaults to on with --continue-on-failure, off otherwise')
     add('--no-summarize', '--no-summary', action='store_false', dest='summarize',
         help='explicitly disable the end of build summary')
-    add('--override-build-tool-check',
+    add('--override-build-tool-check', action='store_true', default=False,
         help='use to override failure due to using differnt build tools on the same workspace.')
 
     # Deprecated args now handled by main catkin command
--- a/catkin_tools/verbs/catkin_build/cli.py
+++ b/catkin_tools/verbs/catkin_build/cli.py
@@ -15,11 +15,12 @@
 from __future__ import print_function
 
 import argparse
-import os
 import sys
 import time
 
 from catkin_pkg.package import InvalidPackage
+from catkin_pkg.tool_detection import get_previous_tool_used_on_the_space
+from catkin_pkg.tool_detection import mark_space_as_built_by
 
 from catkin_tools.argument_parsing import add_context_args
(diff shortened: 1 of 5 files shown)


tfoote:
+1

The unit tests will need to be rerun once catkin_pkg is released with the new version.

jbohren:
Should this be done here or in `build_isolated_workspace()` in `build.py`?

wjwwood:
I wanted to have the check occur before the `dry_run` function, now I realize that the `mark_space_as_built_by` functions should also be after the `dry_run` function.

jbohren:
Ok then we should probably move the other check here too.

dirk-thomas:
This has been addressed in `catkin_pkg` and `catkin` which have both been released since then. I have closed the related tickets in these repos. Once `catkin_tools` merges and releases this check it would actually work.

catkin_tools/verbs/catkin_build/cli.py
```diff
@@ -142,6 +143,8 @@ def prepare_arguments(parser):
         help='Adds a build summary to the end of a build; defaults to on with --continue-on-failure, off otherwise')
     add('--no-summarize', '--no-summary', action='store_false', dest='summarize',
         help='explicitly disable the end of build summary')
+    add('--override-build-tool-check',
```
NikolausDemmel: shouldn't there be `, action='store_true', default=False,`?

NikolausDemmel:
> Ok then we should probably move the other check here too.

which check exactly?

> I wanted to have the check occur before the dry_run function, now I realize that the mark_space_as_built_by functions should also be after the dry_run function.

Why does it need to be moved after `dry_run`?

NikolausDemmel:
Could one of the owners retrigger the tests, please?

catkin_tools/verbs/catkin_build/cli.py
wjwwood: Yes, 7b7ffc0.

wjwwood:
> Why does it need to be moved after dry_run?

Because otherwise running `catkin build -n` on an empty workspace will result in the creation of a build and devel folder marked as catkin build's but with nothing else in them.

So now this is the case:

- Run catkin_make, succeeds
- Run catkin build, fails
- Run catkin build -n, fails
- Run catkin build -n --override-build-tool-check, warns but doesn't build
- Run catkin build -n --override-build-tool-check, _still_ warns but doesn't build
- Run catkin build --override-build-tool-check, warns but builds
- Run catkin build --override-build-tool-check, no warning and builds
- rm -rf build devel
- Run catkin build -n, no build or devel folder created
- Run catkin build , builds and therefore build and devel folder created

I've tested this with both `catkin_make` and `catkin_make_isolated` locally. Works for me. I also removed the extra checks which I think was what @jbohren was talking about, but he can comment if that wasn't the case.

PR #226, opened 2015-08-06
Using "/bin/bash" as the default if SHELL is unset.

pjreed:
This line was recently modified so that is does not throw an exception if the SHELL environment variable is not set, but it does so by using an empty string as the default value.  This means that later code that tries to use "shell_path" to execute the shell will fail (because you can't execute an empty string).

This fixes it to use /bin/bash as the default shell if SHELL is unset, which matches the behavior a few lines later that uses "bash" as the default shell_name.

### DIFF
--- a/catkin_tools/resultspace.py
+++ b/catkin_tools/resultspace.py
@@ -53,12 +53,12 @@ def get_resultspace_environment(result_space_path, quiet=False, cached=True):
         )
 
     # Determine the shell to use to source the setup file
-    shell_path = os.environ.get('SHELL', '/bin/bash')
+    shell_path = os.environ.get('SHELL', '/bin/sh')
     (_, shell_name) = os.path.split(shell_path)
 
     # Use fallback shell if using a non-standard shell
-    if shell_name not in ['bash', 'zsh']:
-        shell_name = 'bash'
+    if shell_name not in ['bash', 'sh', 'zsh']:
+        shell_name = 'sh'
 
     # Check to make sure result_space_path contains the appropriate setup file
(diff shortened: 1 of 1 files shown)


jack-oquin:
I have no strong feeling about this, but I wonder if ``/bin/sh`` might be a better choice, being more generic and present on any system that *has* a shell.

pjreed:
I don't have a strong feeling either, I just picked bash because, a few lines below, "bash" is being used as shell_name if it's not already "bash" or "zsh".

catkin_tools/resultspace.py
```diff
@@ -79,7 +79,7 @@ def get_resultspace_environment(result_space_path, quiet=False, cached=True):
         return {}
 
     # Construct a command list which sources the setup file and prints the env to stdout
-    norc_flags = {'bash': '--norc', 'zsh': '-f'}
+    norc_flags = {'bash': '--norc', 'sh': '', 'zsh': '-f'}
```
wjwwood: I guess it's ok to assume that `sh` doesn't source an rc file, though I'm not sure that's true since `sh` can be different things on different OS's. @jbohren what do you think about this?

NikolausDemmel: Maybe we should stick with `/bin/bash` as a default then, as initially suggested. If the need for other shells arises, they can be explicitly added? Otherwise this seems like fishing in the dark without a use case.

mikepurvis:
Whoops, didn't see that #239 is a dupe of this. Consider this a request for one or the other to be merged!

wjwwood:
Closing in favor of https://github.com/catkin/catkin_tools/pull/239

PR #231, opened 2015-11-10
Completion for some opts for catkin build

wkentaro:


wjwwood:
The travis failure is unrelated, and I fixed it in 00531f6 (a new version of flake8 changed the api)

PR #239, opened 2015-12-08
Fall back to /bin/bash for the default shell path.

mikepurvis:
Fixes an issue when building in an environment without SHELL specified (such as inside debuild).

### DIFF
--- a/catkin_tools/resultspace.py
+++ b/catkin_tools/resultspace.py
@@ -53,7 +53,7 @@ def get_resultspace_environment(result_space_path, quiet=False, cached=True):
         )
 
     # Determine the shell to use to source the setup file
-    shell_path = os.environ.get('SHELL', '')
+    shell_path = os.environ.get('SHELL', '/bin/bash')
     (_, shell_name) = os.path.split(shell_path)
 
     # Use fallback shell if using a non-standard shell


NikolausDemmel:
+1

I believe this is a reasonable default until a different choice can be motivated with a use case in #226.

catkin_tools/resultspace.py
```diff
@@ -59,6 +59,7 @@ def get_resultspace_environment(result_space_path, quiet=False, cached=True):
     # Use fallback shell if using a non-standard shell
     if shell_name not in ['bash', 'zsh']:
         shell_name = 'bash'
+        shell_path = "/bin/bash"
```
NikolausDemmel: Actually, specifying the default not here, but above in `shell_path = os.environ.get('SHELL', '/bin/bash')` would allow to set `SHELL` to something custom not ending in `/bash` and still have it work (with the `--norc` flag as if it were bash). With this PR, the value in `SHELL` gets overwritten if it does not end in `/bash` or `/zsh`.

NikolausDemmel: This is what the other PR did initially with https://github.com/pjreed/catkin_tools/commit/4e730b33ccaeb912f023dbb6e536b4851dbc18bf, but there were some concerns about `sh` being maybe a better default. As I said I would stick with `bash` until it is clear how `sh` should behave.

mikepurvis:
Right, makes sense— changed.

@pjreed @wjwwood @jack-oquin You guys good to see this go in?

pjreed:
I'm fine with it.

wjwwood:
I'm going to merge this and then add a check that `/bin/bash` exists and throw a meaningful error if it does not.

wjwwood:
https://github.com/catkin/catkin_tools/pull/243

PR #238, opened 2015-12-07
Properly respect DESTDIR in the build env.

mikepurvis:
This mostly fixes #235. The following now all works:

```
rosinstall_generator ros_base --rosdistro indigo --deps --tar > ros_base.rosinstall
rosinstall src ros_base.rosinstall -j8
catkin config --install --install-space /opt/foo
DESTDIR=$(pwd)/tmp catkin build
rm -rf build devel
sudo mv tmp/opt/foo/ /opt
source /opt/foo/setup.bash
roscore
```

I've also verified that `rospack find` and friends work, and binaries build and can link libraries, including in an overlaid space. Of course, because the space itself isn't relocatable, it _does not_ work in the DESTDIR location— you have to copy it to the install space for it to be functional.

So what's still broken?

```
grep -r test_ws /opt/foo
_setup_util.py:        CMAKE_PREFIX_PATH = '/opt/foo;/home/administrator/test_ws/tmp/opt/foo'.split(';')
share/*/cmake/*Config.cmake:    foreach(path /opt/foo/lib;/opt/foo/lib;/home/administrator/test_ws/tmp/opt/foo/lib)
```

Both of these could be fixed by a post-build find-and-replace, but because the space is functional without that, I'd vote to merge as is, and let that be dealt with later as required— by a flag, perhaps, or even externally by those needing DESTDIR for packaging or whatever.

Issue #235, opened 2015-12-01
DESTDIR issue with building dependencies


mikepurvis:
Successful bootstrap:

```
rosinstall_generator bond --rosdistro indigo --deps --tar > bond.rosinstall
rosinstall src bond.rosinstall
catkin config --install
catkin build
cat install/setup.sh | grep SETUP_DIR  # /full/path/to/ws/install
```

Unsuccessful:

```
rosinstall_generator bond --rosdistro indigo --deps --tar > bond.rosinstall
rosinstall src bond.rosinstall
catkin config --install --install-space /opt/foo
DESTDIR=$(pwd)/tmp catkin build
cat tmp/opt/foo/setup.sh | grep SETUP_DIR  # /opt/foo
```

The error is:

```
/home/administrator/test_ws/build/cpp_common/build_env.sh /usr/bin/cmake /home/administrator/test_ws/src/roscpp_core/cpp_common -DCATKIN_DEVEL_PREFIX=/home/administrator/test_ws/devel -DCMAKE_INSTALL_PREFIX=/opt/foo' in '/home/administrator/test_ws/build/cpp_common'
-- The C compiler identification is GNU 4.8.4
-- The CXX compiler identification is GNU 4.8.4
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Boost version: 1.54.0
-- Found the following Boost libraries:
--   system
--   thread
CMake Error at CMakeLists.txt:6 (find_package):
  By not providing "Findcatkin.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "catkin", but
  CMake did not find one.

  Could not find a package configuration file provided by "catkin" with any
  of the following names:

    catkinConfig.cmake
    catkin-config.cmake

  Add the installation prefix of "catkin" to CMAKE_PREFIX_PATH or set
  "catkin_DIR" to a directory containing one of the above files.  If "catkin"
  provides a separate development package or SDK, be sure it has been
  installed.
```

Problem appears to be in the logic which creates the per-package `build_env.sh` script— in the case where DESTDIR is specified, the appropriate source lines are not properly generated.

[...]

PR #240, opened 2015-12-08
Properly respect DESTDIR in the build env.

mikepurvis:
The method is to explicitly inject _CATKIN_SETUP_DIR into the generated build_env.sh environment wrapper, allowing the build to proceed even when the installspace at build-time is not in its final runtime location.

I'm open to suggestions for what tests could validate the behaviour here— the main scenario where it comes into play is when building two packages, one of which `find_package`s the other. With DESTDIR specified, the build_env of the second package will have `CMAKE_PREFIX_PATH` set incorrectly, and the `find_package` call will fail. Perhaps there's room for a new type of test to be added which covers interactions between multiple packages in a workspace?

This is a follow-on to #238, addressing #235.

### DIFF
--- a/catkin_tools/verbs/catkin_build/job.py
+++ b/catkin_tools/verbs/catkin_build/job.py
@@ -109,7 +109,7 @@ class Job(object):
 def create_env_file(package, context):
     # Exporting _CATKIN_SETUP_DIR supports DESTDIR, where the absolute location of the installspace
     # at build time does not match the intended final location of it.
-    source_snippet = '_CATKIN_SETUP_DIR={space} . "{source_path}"'
+    source_snippet = '_CATKIN_SETUP_DIR={space_path} . "{source_path}"'
     sources = []
     # If installing to isolated folders or not installing, but devel spaces are not merged
     if (context.install and context.isolate_install) or (not context.install and context.isolate_devel):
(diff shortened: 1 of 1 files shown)


mikepurvis:
Tried a few other build configurations, corrected an issue with building against isolated develspaces.

catkin_tools/verbs/catkin_build/job.py
```diff
@@ -107,24 +107,38 @@ def next(self):
...
-        source_path = os.path.join(
-            context.install_space_abs if context.install else context.devel_space_abs,
-            'setup.sh')
-        sources = [source_snippet.format(source_path=source_path)] if os.path.exists(source_path) else []
+        if context.install:
+            if context.destdir:
+                # Intentionally not using os.path.join in this instance, as it's expected that
+                # when DESTDIR is specified, the install space is an absolute path, so a straight
+                # concatenation is more suitable than os.path.join.
+                space_path = context.destdir + context.install_space_abs
```
wjwwood: A better explanation (in my opinion) is that since the install space is absolute (and it always is, not just when used in conjunction with `DESTDIR`), `os.path.join` will not return the concatenation by rather just the install space. It's not that this is more suitable than `os.path.join` (which to me implies `os.path.join` would also work) but instead that `os.path.join` doesn't work at all and therefore this is necessary. E.g.:

```
>>> import os
>>> a = '/tmp/build/'
>>> b = '/usr'
>>> os.path.join(a, b)
'/usr'
>>> a = '/tmp/build'
>>> os.path.join(a, b)
'/usr'
```

wjwwood:
Other than a comment about a comment (so meta) this lgtm. It seems to line up well as the equivalent to https://github.com/ros/catkin/commit/7282a4bb832610e16d1a63c8336be41d781cb0e6 which made `catkin_make_isolated` support `DESTDIR`.

catkin_tools/verbs/catkin_build/job.py
mikepurvis: Yeah, that was the reason, and you're right it should say so. I just didn't want it to become a novel. I can try again, or switch it to [do what catkin does](https://github.com/ros/catkin/commit/7282a4bb832610e16d1a63c8336be41d781cb0e6#diff-616a0c8f980657f6b84468d0ce4abc77R377). What's your preference?

wjwwood: I think this is fine. I might update the comment on merge.

wjwwood:
Since you're using `catkin build` in conjunction with `DESTDIR`, and I just ran this pr through some common uses of `catkin build` and didn't see any problems, I'm going to merge this now and check off the `DESTDIR` item from #90.

catkin_tools/verbs/catkin_build/job.py
wjwwood: I updated it in 72c5f3e, let me know if you think that's incorrect.

mikepurvis:
Sweet, thanks for the quick review.

PR #244, opened 2015-12-10
add documentation for #192

k-okada:
this includes #192

- verbs: adding shell verbs 'cd' and 'source'

### DIFF
--- a/catkin_tools/commands/catkin.py
+++ b/catkin_tools/commands/catkin.py
@@ -101,6 +101,8 @@ def main(sysargs=None):
         help='Forces catkin to output in color, even when the terminal does not appear to support it.')
     add('--no-color', action='store_true', default=False,
         help='Forces catkin to not use color in the output, regardless of the detect terminal type.')
+    add('--locate-extra-shell-verbs', action='store_true',
+        help='Returns the full path of extra shell verbs settings files')
 
     # Generate a list of verbs available
     verbs = list_verbs()
@@ -177,6 +179,13 @@ def main(sysargs=None):
                     used_aliases.append(arg)
                     del verb_aliases[arg]
                 break
+
+    # Check for --locate-extra-shell-verbs
+    for arg in sysargs:
+        if arg == '--locate-extra-shell-verbs':
+            import inspect
+            print(os.path.abspath(os.path.join(inspect.stack()[-1][1],'../../','share/catkin_tools/catkin_shell_verbs.bash')))
+            sys.exit(0)
 
     # Determine the verb, splitting arguments into pre and post verb
     verb = None
--- a/docs/advanced/catkin_shell_verbs.rst
+++ b/docs/advanced/catkin_shell_verbs.rst
@@ -1,0 +1,36 @@
+Shell support in ``catkin`` command
+===================================
+
+When you source the following file, you can use bash/zsh shell functions which provide added utility.
+
+.. code-block:: shell
+
+    . /opt/ros/indigo/etc/bash_completion.d/catkin_shell_verbs.{bash,zsh}
+
+Provided verbs are:
+
+- ``cd`` -- Change to package directory in source space
+- ``source`` -- Source the develspace or installspace of the containing workspace
+
+Full Command-Line Interface
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Change to package directory in source space with `cd` verb.
+
+.. code-block:: text
+
+    usage: catkin cd [ARGS...]
+
+    ARGS are any valid catkin locate arguments
+
+The `source` verb sources the develspace or installspace of the containing workspace
+
+.. code-block:: text
+
+    usage: catkin source [-w /path/to/ws]
+
+    Sources setup.sh in the workspace.
+
+    optional arguments:
+      -w [/path/to/ws] Source setup.sh from given workspace.
+
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -19,6 +19,7 @@ Catkin Command Line Tools
    verbs/catkin_list
    verbs/catkin_locate
    verbs/catkin_profile
+   Advanced: Shell support <advanced/catkin_shell_verbs>
    Advanced: Verb Aliasing <advanced/verb_customization>
    Advanced: Contributing Verbs <development/extending_the_catkin_command>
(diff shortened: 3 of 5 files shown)


docs/index.rst
```diff
@@ -67,6 +68,12 @@ Each of the following verbs is built-in to the ``catkin`` command and has its ow
 - :doc:`list -- Find and list information about catkin packages in a workspace <verbs/catkin_list>`
 - :doc:`profile -- Manage different named configuration profiles <verbs/catkin_profile>`
 
+Shell support in ``catkin`` command
+-----------------------------------
+
+A bash/zsh user can use wrapper wrapper function, please read :doc:`development/catkin_shell_verbs`.
```
wjwwood: "wrapper wrapper" looks like a typo.

docs/advanced/catkin_shell_verbs.rst
```diff
@@ -0,0 +1,36 @@
+Shell support in ``catkin`` command
+===================================
+
+When you source following file, you can use bash/zsh wrapper function for shell support.
```
wjwwood: "When you ``source`` the following file, you can use bash/zsh shell functions which provide added utility."?

docs/advanced/catkin_shell_verbs.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Change to package directory in source space with `cd` verb.
+
+.. code-block:: text
+
+    usage: catkin cd [ARGS...]
+
+    ARGS are any valid catkin locate arguments
+
+The `source` verb sources the develspace of the containing workspace
```
wjwwood: Is it only the devel space or also the install space when appropriate?

wjwwood:
Thanks for working on this @k-okada.

k-okada:
thanks for review, updated docs

docs/advanced/catkin_shell_verbs.rst
```diff
@@ -0,0 +1,36 @@
+Shell support in ``catkin`` command
+===================================
+
+When you source the following file, you can use bash/zsh shell functions which provide added utility.
+
+.. code-block:: shell
+
+    . /opt/ros/indigo/etc/bash_completion.d/catkin_shell_verbs.{bash,zsh}
```
wjwwood: Is that the right directory?

k-okada: Sorry, it is not correct ,and I think it is not installed, how can I configure setup.py to install ./scripts/catkin_shell_verbs.bash

wjwwood: I'm not sure what the right thing to do here is, maybe @jbohren can advise?

You could do something cute like this:

```
$ source `catkin --locate-extra-shell-verbs`
```

And then install the shell scripts that need to be sourced in with the Python code and provide an option to expose their final location.

k-okada: ok, how about this 25b8c4fc3fd3cd0107f97bbd45bf0dfb775c226b, I'm not sure
how to get installed location ...
http://stackoverflow.com/questions/50499/how-do-i-get-the-path-and-name-of-the-file-that-is-currently-executing

--
◉ Kei Okada

NikolausDemmel: Possibly [`resource_filename`](https://pythonhosted.org/setuptools/pkg_resources.html#resource-extraction) is what you are looking for.

k-okada: thanks, in this case, we should install `catkin_shell_verbs.bash` under `lib/python2.7/dist-packages/catkin_tools`. Is it correct? now I install this file under `share/catkin_tools`.

catkin_tools/commands/catkin.py
```diff
@@ -178,6 +180,13 @@ def main(sysargs=None):
                     del verb_aliases[arg]
                 break
 
+    # Check for --locate-extra-shell-verbs
+    for arg in sysargs:
+        if arg == '--locate-extra-shell-verbs':
+            import inspect
+            print(os.path.abspath(os.path.join(inspect.stack()[-1][1],'../../','share/catkin_tools/catkin_shell_verbs.bash')))
```
wjwwood: This looks vestigial.

docs/advanced/catkin_shell_verbs.rst
wjwwood: I think `share/catkin_tools` is appropriate as long as you can find it easily enough at runtime. `zsh

wjwwood:
Looks like the CI is not passing, but other than a few comments it is looking good to me.

docs/advanced/catkin_shell_verbs.rst
NikolausDemmel: Do you know a reliable way to determine where the correct `share/catkin_tools` is to be found?

k-okada: what I found so far is follows:
```
 os.path.abspath(os.path.join(os.path.dirname(inspect.getfile(catkin_tools)), '../../../../share/catkin_tools'))
```
```
os.path.abspath(os.path.join(imp.find_module('catkin_tools')[1], '../../../../share/catkin'))
```
if we do not have other solution, we have to accept this ugly expresson, or put `catkin_shell_verbs.bash` under `dist-packages`

jbohren: @k-okada What about this: http://peak.telecommunity.com/DevCenter/setuptools#non-package-data-files

k-okada:
```
In [2]: resource_filename(Requirement.parse("catkin_tools"), "catkin_shell_verbs.bash")
Out[2]: '/home/k-okada/catkin_ws/ws_catkin/install/lib/python2.7/dist-packages/catkin_shell_verbs.bash'

In [3]: 
Do you really want to exit ([y]/n)? y
k-okada@kokada-t440s:~/catkin_ws/ws_catkin$ find install  -name catkin_shell_verbs.bash
install/share/catkin_tools/catkin_shell_verbs.bash
```
x-9

wjwwood: I'd say just put it in `dist-packages` along side the code.

wjwwood:
I fixed this up and merged it, thanks @k-okada!

PR #246, opened 2015-12-13
Use red icon on fail

k-okada:
this patch uses red colored notification icon for failure cases, 'Build Finished' and 'Build Failed' looks very similar, at least for me...

hoge

catkin_tool_icon_success


jbohren:
+1

PR #248, opened 2015-12-15
New asyncio-based execution engine (PR#196 1/3)

jbohren:
This PR is the first set of refactored partial changes from PR #196 relating only to the execution engine and I/O handling, without the "linked" develspace support nor improved `catkin clean` support. It also includes the changes from #247. This is meant to be a "functional" PR and it is expected that it might need further small revisions before being merged. This also adds a Travis CI configuration that builds on Linux as well as OS X. 

#### `catkin_tools/common.py`

The changes in `catkin_tools.common` include the following:
* Removal of the no-longer needed `FakeLock`
* Addition of `get_recursive_build_dependants_in_workspace` function which gives the packages that *depend on* a given package in the workspace. (This is used in the future)
* Replacement of `os.Popen` with `subprocess.Popen`
* Fixing bug where `slice_to_printed_length` fails if the length is larger than the length of the lookup array.
* Fixing uncaught IOError when the program receives a `sigint` during a `wide_log` message
* Adds a `mkdir_p` function which recursively makes directories

#### `catkin_tools/resultspace.py`

The main change to `catkin_tools.resultspace` is the addition of environment and env hook checksum caching. This enables faster builds by using the `resultspace` mechanism instead of the `build_env.sh` mechanism. The `resultspace` loading can also cache the environment, and updates the cache when a resultspace's Catkin env_hooks change in any way.

It also adds a `strict` flag which controls whether or not to check for a `.catkin` file before loading the environment.

#### `catkin_tools/argument_parsing.py`

The changes in `catkin_tools.argument_parsing` mainly support the new job server, which is initialized even when there isn't any support for the GNU Make jobserver, to allow for future extensions to the jobserver in other contexts (distcc, ninja, etc).

### Execution Engine

The execution engine consists of the main asyncio-based `Executor`, `Job`s, and job `Stage`s described in #196. Jobs for different catkin build types are defined via setuptools `entry_points`. This improves job parallelism, but also provides a mechanism for capturing `stdout` and `stderr` independently. As such, warnings and errors are now detected and printed in a much clearer manner.

This PR improves on the CMake and Catkin jobs defined in #196 by forgoing using the `build_env.sh` script, and instead using the `catkin_tools.resultspace`-based environment loading. This not only speeds up builds, but it also fixes a bug that prevented single-pass workspace building as described in [this thread](https://github.com/catkin/catkin_tools/pull/196#discussion-diff-46946416).

More details on the design of the execution engine are forthcoming via a documentation PR.

PR #247, opened 2015-12-15
[PR196 0/3] Cleanup of integrated test harness, other small fixes

jbohren:
This PR is a prerequisite to merging PR #196. The main changes in this PR include unifying the integrated test harnesses and adding numerous test cases. This PR also involves a few small additions, bugfixes, and cleanup. This PR is a prerequisite to the first segment of PR#196 (new execution pipeline).

The individual changes to the different components of `catkin_tools` are broken up into six atomic commits. Note that commit 8e2df65 only appears to be large due to re-indenting into a try-catch block. In the future, `catkin_tools/commands/catkin.py` could use some additional cleanup and refactoring.

Issue #229, opened 2015-10-21
Unhelpful error message with circular dependency


DLu:
If you have a circular dependency, catkin_make outputs a decent error message. 

    CMake Error at /home/dlu/test_workspace/build/catkin_generated/order_packages.cmake:7 (message):
      Circular dependency in subset of packages:

      docbrown, marty

However, catkin build results in this error:

    Found '2' packages in 0.0 seconds. 
    [build] Runtime: 0.0 seconds 
    Traceback (most recent call last):
      File "/usr/bin/catkin", line 9, in <module>
        load_entry_point('catkin-tools==0.3.0', 'console_scripts', 'catkin')()
      File "/usr/lib/python2.7/dist-packages/catkin_tools/commands/catkin.py", line 219, in main
        sys.exit(args.main(args) or 0)
      File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/cli.py", line 296, in main
        summarize_build=opts.summarize  # Can be True, False, or None
      File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/build.py", line 601, in     build_isolated_workspace
        packages_to_be_built = topological_order_packages(dict(packages_to_be_built))
      File "/usr/lib/python2.7/dist-packages/catkin_pkg/topological_order.py", line 138, in topological_order_packages
        if package.name in decorators_by_name:
    AttributeError: 'str' object has no attribute 'name'

DLu:
http://gonzo.probablydavid.com/stuff/test_workspace.tar.gz

wjwwood:
Looks like the error is coming up from https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/topological_order.py#L117 which is in `catkin_pkg`. It might be that we have to check for circular dependencies before passing to that function or there is a regression in that function. Not sure.

I don't have time right now to dig into it, but I can try to have a look sometime next week.

IanTheEngineer:
Good find, David. Here's a reproduction of the bug, visualized:
![](http://33.media.tumblr.com/07329ebbf15aa37f8dedf296a0f5586b/tumblr_n1xczxDoQe1r5dm0so1_500.gif)

jbohren:
This is fixed in #249.

DLu:
+1

kopp:
[...]

Issue #213, opened 2015-06-22
testing: Enable travis CI Testing on OS X


jbohren:
A recent bug has shown up on OS X, but not on Linux, but it appears that python testing is still unsupported for OS X on Travis: https://github.com/travis-ci/travis-ci/issues/2312

jbohren:
But there is a workaround: https://travis-ci.org/jbohren-forks/catkin_tools/builds/67902954

jbohren:
Operational: https://travis-ci.org/catkin/catkin_tools/builds/67988718

PR #249, opened 2015-12-15
[PR196 1/3] New asyncio-based execution engine

jbohren:
This PR is the first set of refactored partial changes from PR #196 relating only to the execution engine and I/O handling, without the "linked" develspace support nor improved `catkin clean` support. It also includes the changes from #247 (and supersedes #248). This is meant to be a "functional" PR and it is expected that it might need further small revisions before being merged. This also adds a Travis CI configuration that builds on Linux as well as OS X. 

[asciicast](https://asciinema.org/a/65urzf23q82rj45o1msamzsqg)
#### `catkin_tools/common.py`

The changes in `catkin_tools.common` include the following:
* Replacing `FakeLock` with an asyncio-based implementation
* Addition of `get_recursive_build_dependants_in_workspace` function which gives the packages that *depend on* a given package in the workspace. (This is used in the future)
* Replacement of `os.Popen` with `subprocess.Popen`
* Fixing bug where `slice_to_printed_length` fails if the length is larger than the length of the lookup array.
* Fixing uncaught `IOError` when the program receives a `sigint` during a `wide_log` message
* Adds a `mkdir_p` function which recursively makes directories

#### `catkin_tools/resultspace.py`

The main change to `catkin_tools.resultspace` is the addition of environment and env hook checksum caching. This enables faster builds by using the `resultspace` mechanism instead of the `build_env.sh` mechanism. The `resultspace` loading can also cache the environment, and updates the cache when a resultspace's Catkin env_hooks change in any way.

It also adds a `strict` flag which controls whether or not to check for a `.catkin` file before loading the environment.

#### `catkin_tools/argument_parsing.py`

The changes in `catkin_tools.argument_parsing` mainly support the new job server, which is initialized even when there isn't any support for the GNU Make jobserver, to allow for future extensions to the jobserver in other contexts (distcc, ninja, etc).

### Execution Engine

The execution engine consists of the main asyncio-based `Executor`, `Job`s, and job `Stage`s described in #196. Jobs for different catkin build types are defined via setuptools `entry_points`. This improves job parallelism, but also provides a mechanism for capturing `stdout` and `stderr` independently. As such, warnings and errors are now detected and printed in a much clearer manner.

This PR improves on the CMake and Catkin jobs defined in #196 by forgoing using the `build_env.sh` script, and instead using the `catkin_tools.resultspace`-based environment loading. This not only speeds up builds, but it also fixes a bug that prevented single-pass workspace building as described in [this thread](https://github.com/catkin/catkin_tools/pull/196#discussion-diff-46946416).

More details on the design of the execution engine are forthcoming via a documentation PR #250 

### Job Environments

Previously, each build job got its environment from a `build_env.sh` file, which was generated before the package was built. This shell script essentially sources the resultspace that each job is meant to be built against.

This PR adds the ability to cache the resulting environment from sourcing these resultspace setup file. In addition to speeding up all builds, it speeds up building already-built packages dramatically. For the Hydro `ros-base` workspace (172 packages), `catkin build` without env caching takes about 30 seconds, and `catkin build` with env caching takes about 10 seconds. This means that it's saving between 0.1-0.2 seconds per package, which adds up. Note that this effect becomes more dramatic as more workspaces are chained.

This PR removes the `build_env.sh` files entirely, and instead opts for a non-static way to get the environment for a job. This is the `--env` option which can be passed to any verb which supports it. So to get the environment in which `qt_gui_cpp` is built, you could run `catkin build --env qt_gui_cpp`, and it will print the environment to `stdout`. This is used instead of `build_env.sh` to reproduce build stages as well, as shown below:

```
_________________________________________________________________________________
Warnings   << qt_gui_cpp:make /home/jbohren/ws/ct_roscomm/build/_logs/qt_gui_cpp/build.make.000.log
cd /home/jbohren/ws/ct_roscomm/build/qt_gui_cpp; catkin build --env qt_gui_cpp | xargs -I %ENV% env %ENV% /usr/bin/make --jobserver-fds=6,7 -j; cd -
sip: Deprecation warning: qt_gui_cpp.sip:1: %Module version number should be specified using the 'version' argument
** WARNING expected ``)'' = 40
.................................................................................
```

Since this feature changes behavior, it is off by default, but can be enabled / disabled with `--env-cache` and `--no-env-cache`, respectively.

#### Enhancements

* #1 - Detection and reporting of build warnings
* #2 - New standalone execution pipeline
* #63 - Provides a summary at the end of the build
* #188 - By default, the `-p` option is set to whatever the `-j` option is
* #213 - OS X Travis CI configuration

#### Definite Bugfixes

* #139 - Running `catkin test` doesn't work under Hydro (fixed in #265)
* #208 - Fixes jobserver support on Python3.4
* #229 - Fixes unhelpful error message
* #270 - Fixes problems with `CATKIN_SHELL` environment variable

#### Likely Bugfixes (through different implementation)

* #59 - Ctrl-C should now always work
* #66 - UI should not get stuck

### Outstanding Issues

* [x] Disable buffered stderr when in interleaved mode (https://github.com/catkin/catkin_tools/pull/249#issuecomment-173027781)
* [x] Fix verbose interleaved output overlap with status message (https://github.com/catkin/catkin_tools/pull/249#issuecomment-173026313)
* [x] Fix CMake output parsing error (https://github.com/catkin/catkin_tools/pull/249#issuecomment-173028274)
* [x] Reincorporate interleaved output mode
* [x] Fix "packages have changed" warning (#273)
* [x] Make repro command more portable (done in #273)
* [x] Add `--env-cache` and `--no-env-cache` options to control caching of resultspace environments
* [x] Re-enable verbose summary when `--summarize` option is given 
* [x] Take over responsibility for populating `.catkin` file (implemented in #271)
* [x] Fix `CommandStage` mutable default argument (fixed in 8a542c6e)
* [x] Fix `CommandStage` `env` argument to make behavior clearer  (fixed in 8a542c6e)
* [x] Unicode handling causes problems in certain contexts ([comment](https://github.com/catkin/catkin_tools/pull/249#issuecomment-165198050) #256)
* [x] IO Logger encoding / decoding needs to be cleaned up
* [x] Clean up old test harness comments
* [x] `DESTDIR` support is broken (fixed in #251)
* [x] Resultspace loading fails to parse environment under Python3 (fixed in #251)
* [x] Console controller crashes if there are no packages in the job list (fixed in #251)
* [x] If the `catkin` package is in the workspace, it's skipped (fixed in #251)
* [x] This PR moves away from using `build_env.sh` files, and instead utilizes a  more platform-independent resultspace sourcing model, but `build_env.sh` is still used in the isolated devel setup. Enhancements are staged in #254.
 * [x] Migrate away from `build_env.sh` (completed in #254)
 * [x] Add build-env flag to `catkin build` i.e. `catkin build --env PKG` which dumps the resultspace environment to the console. This would enable debugging of the build environment used to build `PKG` as well as build stage repro. like: `cd /path/to/build && env "$(catkin build --env PKG)" make` (added in #254)


catkin_tools/verbs/catkin_build/build.py
```diff
@@ -569,291 +359,294 @@ def build_isolated_workspace(
...
-    packages_to_be_built = topological_order_packages(dict(packages_to_be_built))
+
+    # Also re-sort
+    try:
+        packages_to_be_built = topological_order_packages(dict(packages_to_be_built))
+    except AttributeError as err:
+        log(clr("[build] @!@{rf}Error:@| The workspace packages have a circular "
+                "dependency, and cannot be built. Please run `catkin list "
+                "--deps` to determine the problematic package(s)."))
+        return
```
jbohren: The above provides a more graceful failure for #229

xqms:
Sorry, answered to the wrong PR. I'm putting this here again so it's in the right place ;-)

Thanks for your work on separating this!

I just saw a unicode exception here:

```
Warnings << gazebo_ros_control:make /var/lib/jenkins/jobs/spacebot/workspace/build/_logs/gazebo_ros_control/build.make.000.log
Exception in thread Thread-1:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/home/max/install/catkin_tools/catkin_tools/execution/controllers.py", line 452, in run
    wide_log('\n'.join(lines))
  File "/home/max/install/catkin_tools/catkin_tools/common.py", line 450, in wide_log
    wide_log_fn(msg, **kwargs)
  File "/home/max/install/catkin_tools/catkin_tools/common.py", line 419, in disabled_wide_log
    log(msg, **kwargs)
  File "/home/max/install/catkin_tools/catkin_tools/common.py", line 272, in log
    print(*args, **kwargs)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 129: ordinal not in range(128)
```

The unicode character 2018 is "LEFT SINGLE QUOTATION MARK", emitted by gcc on my system:
```
[...] warning: enumeration value ‘EFFORT’ not handled in switch
```

xqms:
Sorry, this may have been caused by running `catkin_tools` under Jenkins, which apparently sets stdout encoding to ascii. Running `catkin build` straight in the terminal works just fine.

I'll investigate further.

jbohren:
> Sorry, this may have been caused by running `catkin_tools` under Jenkins, which apparently sets stdout encoding to ascii. Running `catkin build` straight in the terminal works just fine.

@xqms Yeah, overall, the unicode encoding/decoding needs to be cleaned up before or after this gets merged. Can you simulate the stdout encoding in a unit test?

xqms:
For reference, I was able to fix my issue by setting the environment variable `PYTHONIOENCODING=utf_8` when executing catkin_tools under jenkins. This forces the stdout encoding to UTF-8.

> @xqms Yeah, overall, the unicode encoding/decoding needs to be cleaned up before or after this gets merged. Can you simulate the stdout encoding in a unit test?

I guess we should define the wanted behavior first. I'd propose that bytes captured from a job should be passed through 1:1 without caring about the encoding. First decoding to unicode and then re-encoding again to some variable output encoding is just asking for trouble.

This might conflict with the need to actually parse the output (e.g. line splitting). Checking for the newline byte should still be possible without decoding to unicode, though...

NikolausDemmel:
One question: Why are we downloading and compiling catkin from github and then sourcing the resulting `setup.bash` in the travis config? It turns out that (on my OS X machine) 5 tests fail without this, but why should that be a prerequisite?

jbohren:
> One question: Why are we downloading and compiling catkin from github and then sourcing the resulting `setup.bash` in the travis config?

Well, we're getting catkin from github on Travis because it's a prerequisite for building the Catkin packages.

> It turns out that (on my OS X machine) 5 tests fail without this, but why should that be a prerequisite?

@NikolausDemmel Which tests are failing, and are you running the tests in a clean environment, or after sourcing *some* setup file? If it's a clean environment, I would expect any tests which build Catkin CMake packages to fail.

NikolausDemmel:
> Well, we're getting catkin from github on Travis because it's a prerequisite for building the Catkin packages.

Would it make sense to have a copy of catkin in the test suite for that?

> Which tests are failing, and are you running the tests in a clean environment, or after sourcing some setup file? If it's a clean environment, I would expect any tests which build Catkin CMake packages to fail.

Clean environment w-r-t ROS, i.e. I haven't sourced any setup file, catkin is not available. It makes sense as you say. Output of test command is: https://gist.github.com/NikolausDemmel/b7e5acea3ec3539672c3#file-catkin_tools-nosetests

jbohren:
>> Well, we're getting catkin from github on Travis because it's a prerequisite for building the Catkin packages.

>Would it make sense to have a copy of catkin in the test suite for that?

I don't think that's necessary. It might be best to fix the version (or even add more versions to the matrix), though. What do you think, @wjwwood?

>> Which tests are failing, and are you running the tests in a clean environment, or after sourcing some setup file? If it's a clean environment, I would expect any tests which build Catkin CMake packages to fail.

> Clean environment w-r-t ROS, i.e. I haven't sourced any setup file, catkin is not available. It makes sense as you say. Output of test command is: https://gist.github.com/NikolausDemmel/b7e5acea3ec3539672c3#file-catkin_tools-nosetests

Yeah, that's what I would expect, all's good.

NikolausDemmel:
> I don't think that's necessary. It might be best to fix the version (or even add more versions to the matrix), though.

I just find it a bit strange that after downloading `catkin_tools` and installing all python and system dependencies you can't successfully run the unit tests without making sure catkin is in your environment. That's why I think keeping a copy to make the tests self-contained might be reasonable. It would also fix the version of catkin that is tested against and can be updated explicitely when needed. 

But in any case there should be some documentation about this if catkin is not included. I believe on your `docs` branch you have already some stuff for developers, like how to add verbs etc, so that might be the best place.

jbohren:
> > I don't think that's necessary. It might be best to fix the version (or even add more versions to the matrix), though.

> I just find it a bit strange that after downloading `catkin_tools` and installing all python and system dependencies you can't successfully run the unit tests without making sure catkin is in your environment. That's why I think keeping a copy to make the tests self-contained might be reasonable. It would also fix the version of catkin that is tested against and can be updated explicitely when needed. 

Yeah, I'm just always weary of copying source code around. If anything, I'd have the tests check it out from source and build/source it.

> But in any case there should be some documentation about this if catkin is not included. I believe on your `docs` branch you have already some stuff for developers, like how to add verbs etc, so that might be the best place.

That sounds reasonable, I'll add that to #250.

wjwwood:
My preference would be to have the tests check out a copy of catkin during the test setup. Other necessities like `catkin_pkg` and `empy` are actually dependencies (directly or indirectly) so they should be installed when you run the tests.

mikepurvis:
By necessity, catkin_tools and catkin are joined at the hip— it makes sense to have tests on catkin_tools which will detect breaking changes in catkin proper.

wjwwood:
> By necessity, catkin_tools and catkin are joined at the hip— it makes sense to have tests on catkin_tools which will detect breaking changes in catkin proper.

I don't actually agree with the first half of that statement :smile:. It should be perfectly capable (in the future if not now) of installing a workspace full of pure cmake packages.

But I do agree that the main use case is to use it in conjunction with catkin. It's not clear to me what version of catkin it should checkout, e.g. the indigo-devel branch or the latest release tag and in the future when there are multiple "latest" releases to choose from, which one? It might even require it to be a configuration of the tests and to be added to the matrix of things to test.

jbohren:
> But I do agree that the main use case is to use it in conjunction with catkin. It's not clear to me what version of catkin it should checkout, e.g. the indigo-devel branch or the latest release tag and in the future when there are multiple "latest" releases to choose from, which one? It might even require it to be a configuration of the tests and to be added to the matrix of things to test.

Here's the simplest way to implement this, as shown over in #251
https://github.com/catkin/catkin_tools/blob/pre-0.4.0-destdir/.travis.yml#L14

.travis.before_install.bash
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  echo "AOK"
+  #if [ "$PYTHON" == "/usr/bin/python3.4" ]; then
+    #sudo add-apt-repository ppa:fkrull/deadsnakes -y
+    #sudo apt-get update
+    #sudo apt-get install python3.4 python3-dev
+  #fi
```
wjwwood: This looks like it should be removed?

.travis.before_script.bash
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  sudo apt-get install cmake build-essential
+  #sudo apt-get install cmake libgtest-dev build-essential
```
wjwwood: Should be removed or a TODO added.

.travis.before_install.bash
jbohren: Yeah this is left over from before 3.4 was available on Travis Precise.

.travis.yml
```diff
@@ -1,14 +1,51 @@
-language: python
-python:
-  - "2.7"
-  - "3.3"
+# Travis auto-virtualenv isn't supported on OS X
+#language: python
+#python:
+  #- "2.7"
```
jbohren: Yeah I'm not sure why gtest was in there. This can be removed.

wjwwood:
Does this need to be rebased?

jbohren:
> Does this need to be rebased?

@wjwwood I don't think it needs to be rebased, it built on #247 and doesn't interfere with the other ones that you merged yesterday.

jbohren:
@wjwwood I rebased it anyway, for clarity.

wjwwood:
Thanks the commits were a bit difficult to go through with the merge commits in there.

jbohren:
@wjwwood Will I step on your toes if I rebase this now that #264 has been merged?

catkin_tools/execution/stages.py
```diff
@@ -0,0 +1,127 @@
...
+    :param logger_factory: The factory to use to construct a logger (default: IOBufferProtocol.factory)
+
+    """
+
+    def __init__(
+            self,
+            label,
+            cmd,
+            cwd=os.getcwd(),
+            env={},
```
wjwwood: This is dangerous (mutable default arguments http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) please use `None` and do something like `env or {}` where you use it. I know it isn't a problem now, but someone could easily over look this if they update the function in the future.

wjwwood:
I can handle the rebase, go for it.

catkin_tools/execution/stages.py
jbohren: Roger.

jbohren:
Rebased on master after merging #264

jbohren:
@wjwwood I fixed the `CommandStage` env args and mutable default args in 8a542c6e.

jbohren:
>> By necessity, catkin_tools and catkin are joined at the hip— it makes sense to have tests on catkin_tools which will detect breaking changes in catkin proper.

> I don't actually agree with the first half of that statement :smile:. It should be perfectly capable (in the future if not now) of installing a workspace full of pure cmake packages.

Yeah, and while, `catkin` is a good name for the _original_ [catkin](https://en.wikipedia.org/wiki/Catkin) build tool (builds a bunch of CMake packages together as a single build), a better name for the `catkin` and `ament` commands would be `graft`, since they build workspaces of non-homogenous package types.

NikolausDemmel:
One question about the env-caching that has been on my mind for a few days, but I haven't gotten round to checking the code myself: I understand that each package can add files to that can modify the environment. Reading about checksums for the env hooks I assume that you re-evaluate the files only if the content changes. If this is the case, is it guaranteed that reavaluating the same env-hook file would result in the same environment? Being sheel scripts, couldn't they execute arbitrary (non-deterministic) code to set the environment?

jbohren:
> One question about the env-caching that has been on my mind for a few days, but I haven't gotten round to checking the code myself: I understand that each package can add files to that can modify the environment. Reading about checksums for the env hooks I assume that you re-evaluate the files only if the content changes. If this is the case, is it guaranteed that reavaluating the same env-hook file would result in the same environment? Being sheel scripts, couldn't they execute arbitrary (non-deterministic) code to set the environment?

Of course they could. They could even do something which is deterministic but time-varying. I don't think there are any real standards for env-hooks, but I feel like that's not something they're used for, at least not on these time scales. Feel free to prove me wrong if you can, though!

The way that the caching works, though, it only caches _in-process_, and doesn't save the environment to disk. That means that each call to `catkin build` will reevaluate all env-hooks, but won't reevalute the same env-hook multiple times. I don't think we need to worry about the result of an env-hook changing between building of packages in a _single_ invocation, and if it turns out someone is doing this, and it causes problems, we could always add a `--no-env-cache` option.

NikolausDemmel:
> That means that each call to catkin build will reevaluate all env-hooks, but won't reevalute the same env-hook multiple times. 

Ah ok, I didn't get that until now. But if the env hook file changes during the run of catkin build, it is re-evaluated, or what else is the checksum used for?

> I don't think we need to worry about the result of an env-hook changing between building of packages in a single invocation

Hmmm... I guess :-). I certainly agree that they should work that way, but I have no idea what kind of crazy stuff is out there.

One important one that comes to mind is the `roslib` env hook. That one depends on the contents of the `.catkin` file, doesn't it? I'm not when and how this file might change during the build of a workspace, in particular with all the devel-space variations.

jbohren:
> One important one that comes to mind is the `roslib` env hook. That one depends on the contents of the `.catkin` file, doesn't it? I'm not when and how this file might change during the build of a workspace, in particular with all the devel-space variations.

Right. So the `.catkin` file is already the source of a few problems, at the moment. The biggest one is synchronization between parallel invocations of CMake. The linked develspace mode (which is coming in the next big PR (2/3)) takes over responsibility for building this file, but maybe that needs to be bumped up into this PR.

Currently, each Catkin `cmake` invocation causes the `.catkin` file to be appended if the package name isn't already present in it. I'm going to modify the `catkin build` behavior to preempt this by writing (and over-writing) the `.catkin` file at the beginning of each build. This is actually _more_ similar to what `catkin_make` does, since usually this file is fully populated at configuration time with all packages in the workspace.

@wjwwood I'm pretty sure the issue @NikolausDemmel is bringing up is the issue you mentioned off-line about catkin-based python modules not being found.

jbohren:
@wjwwood @NikolausDemmel Check out #271, which takes over generation of the `.catkin` devel manifest file. I think this is a simple solution to better mirror the behavior of `catkin_make` wherein all of the source packages are registered in the `.catkin` file before any are built.

wjwwood:
My environment issue is no longer present when doing `catkin config --install;catkin build`. I'll just check to see if the problem comes up with environment caching or not (should tell us if https://github.com/catkin/catkin_tools/pull/265 solved the problem when using env caching).

catkin_tools/execution/stages.py
```diff
@@ -0,0 +1,151 @@
...
+            # Capture stderr and stdout separately
+            'stderr_to_stdout': stderr_to_stdout,
+        }
+
+    def update_env(self, base_env):
+        """Update the environment for this stage with the env args passed to the constructor."""
+        self.async_execute_process_kwargs['env'].update(base_env)
+        self.async_execute_process_kwargs['env'].update(self.env_overrides)
+
+    def get_repro(self, verb, jid):
```
wjwwood: Can you change the name of this to `get_reproduction_line`? `repro` looks too much like `repo` to me.

catkin_tools/execution/stages.py
```diff
@@ -0,0 +1,151 @@
...
+        # Add additional env args
+        env_override_repro = ' '.join([
+            '{}={}'.format(k, cmd_quote(v))
+            for k, v in self.env_overrides.items()
+        ])
+        if len(env_override_repro) > 0:
+            env_cmd = 'echo "$({}) {}"'.format(env_cmd, env_override_repro)
+
+        # Return the full command
+        return 'cd {}; {} | xargs -I %ENV% env %ENV% {}; cd -'.format(
```
wjwwood: Let's not use `xargs` here. For a few reasons:

- It's confusing and not obvious to new users.
- It doesn't work on Windows. (Important long term for ROS 2 support)
- It doesn't work the same on all OS's, e.g. BSD and OS X's `xargs` has different options than GNU version.

On my machine, when trying to reproduce a build using this:

```
$ cd /Users/william/indigo/build/rqt_nav_view; catkin build --env rqt_nav_view | xargs -I %ENV% env %ENV% /usr/local/bin/cmake /Users/william/indigo/src/rqt_robot_plugins/rqt_nav_view --no-warn-unused-cli -DCATKIN_DEVEL_PREFIX=/Users/william/indigo/devel -DCMAKE_INSTALL_PREFIX=/Users/william/indigo/install; cd -
env: %ENV%: No such file or directory
~/indigo
```

My guess is that this is due to a difference in the `xargs` on OS X versus the one on Linux.

I'd propose a new option for `build` or a new verb to handle this case:

```
$ catkin build --run-with-env-for-package <package name> <command> [arg1 [argN] [--]]
```

Or something like this:

```
$ catkin run_with_env_for_package --package <package name> <command> [arg1 [argN] [--]]
```

The description of this commands can explain that it works like `xargs -I {} env {} ...` but is portable.

The least effort change would be to figure out how to change the output so it would work on OS X and Linux.

wjwwood: It's not clear to me yet why this isn't working. I don't think it is a difference in `xargs` anymore, but I still like to avoid utilizing command line tools like this as part of the build or even recommended actions for the user.

wjwwood: It might be because some of my env variables have `'`'s in them? I'll have to dig in more this afternoon when I have some spare time.

jbohren: Alright, what if we mimic the `env` command but with a `--pkg` option, i.e.:

```
catkin env --pkg PKGNAME [NAME=VAL]... [COMMAND [ARG]...]
```

jbohren: > It might be because some of my env variables have `'`'s in them? I'll have to dig in more this afternoon when I have some spare time.

I don't think we should waste any more time on this, trying to make such a shell command work cross-platform is not going to be exceedingly difficult if not impossible. Command substitution does all kinds of weird stuff with string splitting and escaping, and it wouldn't work on windows anyway.

NikolausDemmel:
I have been using this for a bit now. It is pretty awesome. One thing I noticed was that the summary pretty much always says `Note: Workspace packages have changed, please re-source setup files to use them.` even if I rebuild a workspace without changing anything compared to the previous run. Is this intended?

```
[build] Summary: All 71 packages succeeded!
[build]   Ignored:   12 packages were skipped or are blacklisted.
[build]   Warnings:  None.
[build]   Abandoned: None.
[build]   Failed:    None.
[build] Runtime: 3.8 seconds total.
[build] Note: Workspace packages have changed, please re-source setup files to use them.
```

jbohren:
@NikolausDemmel I'm glad you like it! Also yeah, I've noticed that too, but keep forgetting to add it to the teacher. I think it's a check that was broken when I split this off from #196.  It'll get resolved before merge.

catkin_tools/execution/stages.py
jbohren: Fixed in #273 [here](https://github.com/catkin/catkin_tools/pull/273/files#diff-62ab2c322956d57c062b8f2b885c4c9cL45)

catkin_tools/execution/stages.py
jbohren: I've replaced the use of `xargs` and `env` with a `catkin env` verb in #273. More details in that PR.

wjwwood:
@jbohren what happened to the `--interleave` option?

jbohren:
> @jbohren what happened to the `--interleave` option?

It's just missing. The console controller can do it pretty easily, I just need to carry the option back through. Adding to checklist.

wjwwood:
@jbohren I tried to check a box on the OP and it looks like it might have cleared some of your edits... Do you think you can reproduce them? Sorry :grin:

jbohren:
> @jbohren I tried to check a box on the OP and it looks like it might have cleared some of your edits... Do you think you can reproduce them? Sorry :grin: 

I re-added one of them, everything is tracked in the commit log, so it's not too important if we lost them. They were all in #273 or fdaa2a7 anyway.

wjwwood:
I'll give fdaa2a7 a try.

wjwwood:
@jbohren There seems to be some issues with the output when I use `catkin build -iv`. Can you try it on your end again?

jbohren:
> @jbohren There seems to be some issues with the output when I use catkin build -iv. Can you try it on your end again?

Yeah, I think autopep8 screwed something up. Give me a minute.

jbohren:
Sorry, I had been testing with `--no-status`, fixed in 92c27ce
I needed to make sure the wide-logs ended with `\r`s

wjwwood:
You need to lead the output with `'\r' + (' ' * terminal_width()) + <output>`, otherwise you get lines like this: `[100%] Built target cpp_common] [1/4 jobs] [64 queued] [High Load] [cpp_common:make (100%) - 2.7]`

wjwwood:
Sorry that should be: `'\r' + (' ' * terminal_width()) + '\r' + <output>`

wjwwood:
Also, should the warnings be re-quoted when using `-i`? I get stuff like this:

```
[ ... ]
-- Found PY_em: /Library/Python/2.7/site-packages/em.pyc
-- Using empy: /Library/Python/2.7/site-packages/em.pyc
-- Using CATKIN_ENABLE_TESTING: ON
-- Call enable_testing()
-- Using CATKIN_TEST_RESULTS_DIR: /Users/william/indigo/build/rosconsole_bridge/test_results
-- Found gtest: gtests will be built
-- Using Python nosetests: /usr/local/bin/nosetests-2.7
-- catkin 0.6.16
-- Configuring done
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
-- Generating done
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   rosconsole_bridge

This warning is for project developers.  Use -Wno-dev to suppress it.

-- Build files have been written to: /Users/william/indigo/build/rosconsole_bridge
_________________________________________________________________________________________________________________________________________________________________________________
Warnings   << rosconsole_bridge:cmake /Users/william/indigo/build/_logs/rosconsole_bridge/build.cmake.000.log
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  MACOSX_RPATH is not specified for the following targets:

   rosconsole_bridge

This warning is for project developers.  Use -Wno-dev to suppress it.

cd /Users/william/indigo/build/rosconsole_bridge; catkin build --get-env rosconsole_bridge | catkin env -si  /usr/local/bin/cmake /Users/william/indigo/src/rosconsole_bridge --no-warn-unused-cli -DCATKIN_DEVEL_PREFIX=/Users/william/indigo/devel -DCMAKE_INSTALL_PREFIX=/Users/william/indigo/install; cd -
.................................................................................................................................................................................
```

I haven't checked the behavior when using `-v`.

wjwwood:
Another small thing I noticed are snippets like this one: `@{yf}{cf}--@{yf}| Found the following Boost libraries:` which appear to have been skipped over when doing substitutions.

wjwwood:
Other than the overwriting of the status line when using `-i`, feel free to ticket the additional items to be done separately from this pr.

jbohren:
@wjwwood OK. All the console controller issues are resolved in ff9a92a

wjwwood:
I think we need to print the status line after new input, but there might a performance trade-off.

For now I think it's good enough; we can address things after merging.

wjwwood:
I also think we may not be closing out child processes correctly. But I can investigate that more later.

jbohren:
> I think we need to print the status line after new input, but there might a performance trade-off.

Yeah, I was thinking that, too.

> For now I think it's good enough; we can address things after merging.

:clap: Awesome.

NikolausDemmel:
Nice!!

xqms:
Congrats on getting this merged - I think this has been a major step forward :-)

PR #251, opened 2015-12-16
destdir: Resurrecting destdir support, adding test case (PR#196)

jbohren:
@mikepurvis Your `DESTDIR` support (#240) was added after the main development in #196, so I wanted to make sure I resurrected the right functionality for use in #249. Can you give this branch a shot and tell me if it doesn't break what you need from `DESTDIR`?

Additionally, can you improve (or suggest to improve) the super-trivial test case in this PR to actually test what the `DESTDIR` pattern requires?

### DIFF
--- a/.travis.before_script.bash
+++ b/.travis.before_script.bash
@@ -1,10 +1,8 @@
 #!/usr/bin/env bash
 
 if [ "$TRAVIS_OS_NAME" == "linux" ]; then
-  sudo apt-get install cmake libgtest-dev build-essential python-setuptools python-pip
-  pip install trollius
+  sudo apt-get install cmake build-essential
+  #sudo apt-get install cmake libgtest-dev build-essential
 elif [ "$TRAVIS_OS_NAME" == "osx" ]; then
-  # noop?
-  pip install setuptools
-  pip install trollius
+  echo "All OSX-specific deps satisfied."
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,32 +6,46 @@
 language: generic
 matrix:
   include:
-    - python: 2.7
+    - dist: precise
+      python: 2.7
       language: python
       python: "2.7"
       os: linux
-      env: PYTHON=/usr/bin/python2.7
-    - python: 3.4
+      env: PYTHON=/usr/bin/python2.7 CATKIN_VERSION=groovy-devel
+    - dist: precise
+      python: 3.4
       language: python
       python: "3.4"
       os: linux
-      env: PYTHON=/usr/bin/python3.4
-    - python: 2
-      os: osx
-      env: PYTHON=/usr/bin/python
-    - python: 3
-      os: osx
-      env: PYTHON=/usr/local/bin/python3
+      env: PYTHON=/usr/bin/python3.4 CATKIN_VERSION=groovy-devel
+    - dist: trusty
+      python: 2.7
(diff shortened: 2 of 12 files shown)


mikepurvis:
Had to `pip install git+https://github.com/osrf/osrf_pycommon.git` first.

Looks like this breaks my use-case. Specifically, the new `generate_cached_setup.py` (replacement for build_env.sh, it seems?) is not DESTDIR-aware: it's looking in the final installdir, not ${DESTDIR}installdir. My grep-fu is failing me, though. Where is the code that generates that file?

jbohren:
`generate_cached_setup.py` is something that Catkin proper creates. I'll trace the information flow through the `build_env.sh` and setup files a little further. Any chance you could post a quick `DESTDIR` use case (even as a couple of lines in this thread)?

mikepurvis:
The example given in the [previous discussion](https://github.com/catkin/catkin_tools/issues/235#issue-119763096) is still a pretty good test:

```
rosinstall_generator bond --rosdistro indigo --deps --tar > bond.rosinstall
rosinstall src bond.rosinstall
catkin config --install --install-space /opt/foo
DESTDIR=$(pwd)/tmp catkin build
cat tmp/opt/foo/setup.sh | grep SETUP_DIR  # /opt/foo
```

Although when doing this from a clean terminal, I'm actually seeing a different problem (which happens even _without_ DESTDIR)— it's not building catkin first, so the find_package(catkin) on the other packages (eg, genmsg) is failing due to that before anything else.

jbohren:
> The example given in the previous discussion is still a pretty good test:

Ah, sorry, I missed that. I'll use that to test.

jbohren:
@mikepurvis Ok. I expect it to actually work now.

See the stronger test case here: https://github.com/catkin/catkin_tools/blob/074bb1c4dca41745a7281fa486b5884cc1178e36/tests/system/verbs/catkin_build/test_build.py#L236-L265

mikepurvis:
Test case looks better to me, but I'm actually still not getting a successful build on my side, potentially still due to some ordering issues? The actual project that I'm working on can be tried out like so:

```
sudo apt-get install python-rosdep devscripts debhelper -y
git clone https://github.com/mikepurvis/fat-rosdebs
cd fat-rosdebs
python generate-sourcedeb.py
cd clearpath-ros-base-0.0
debuild -us -uc --source-option=--create-empty-orig
```

With the current master of catkin_tools, this produces a ~2.5MB deb in the directory above that's a full ros-base, installed into `/opt/clearpath/ros-base`.

With this branch at present, it tries to build genmsg first, and fails because it can't find `catkin.environment_cache` (and `build/genmsg/catkin_generated/generate_cached_setup.py` still looks for `catkinConfig.cmake.in` in `/opt`): https://gist.github.com/mikepurvis/73d270bfdd44c640bb27

jbohren:
> Test case looks better to me, but I'm actually still not getting a successful build on my side, potentially still due to some ordering issues? The actual project that I'm working on can be tried out like so:

@mikepurvis Ok, I'll take another pass at it tonight. Thanks for the feedback!

jbohren:
@mikepurvis The latest revision f46a78a fixes a bug left over from PR #196 extracting where it was skipping the `catkin` package found in the workspace.

mikepurvis:
Looks good now, thanks!

PR #252, opened 2015-12-16
Fix notifications on OS X after API change.

NikolausDemmel:
Without this catkin build crashes on OSX when it tries to display the notification.

### DIFF
--- a/catkin_tools/notifications/impl.py
+++ b/catkin_tools/notifications/impl.py
@@ -23,7 +23,11 @@ from catkin_tools.utils import which
 this_dir = os.path.dirname(__file__)
 
 
-def _notify_osx(title, msg):
+def _notify_osx(title, msg, icon_image):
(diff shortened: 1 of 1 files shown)


jbohren:
@wjwwood lgtm.

NikolausDemmel:
So I was woundering how to have a similar visual distinction on OS X as we now have on Linux for the two different icons. On OS X, the icon is App specific and cannot be set arbitrarily. Since 10.9, it is however apparently possible to set a user content image, which is displayed in addition to the App icon. There seems to also be a private API that allows to set custom content for the main icon and have the app icon move to the side (like for itunes), though I'm not sure for which OS X versions this is valid.

References: http://stackoverflow.com/questions/19797841/nsusernotification-mavericks-custom-images

Any thoughts @wjwwood ?

NikolausDemmel:
thanks

PR #256, opened 2015-12-21
io: Handling ASCII stdout more gracefully.

jbohren:
@xqms Can you try running this branch in jenkins to better deal with the ASCII stdout?

This was the problem, as reported in https://github.com/catkin/catkin_tools/pull/249#issuecomment-164719546
> ```
Warnings << gazebo_ros_control:make /var/lib/jenkins/jobs/spacebot/workspace/build/_logs/gazebo_ros_control/build.make.000.log
Exception in thread Thread-1:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/home/max/install/catkin_tools/catkin_tools/execution/controllers.py", line 452, in run
    wide_log('\n'.join(lines))
  File "/home/max/install/catkin_tools/catkin_tools/common.py", line 450, in wide_log
    wide_log_fn(msg, **kwargs)
  File "/home/max/install/catkin_tools/catkin_tools/common.py", line 419, in disabled_wide_log
    log(msg, **kwargs)
  File "/home/max/install/catkin_tools/catkin_tools/common.py", line 272, in log
    print(*args, **kwargs)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 129: ordinal not in range(128)
```

> The unicode character 2018 is "LEFT SINGLE QUOTATION MARK", emitted by gcc on my system:
> ```
[...] warning: enumeration value ‘EFFORT’ not handled in switch
```

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -18,6 +18,7 @@ import datetime
 import os
 import re
 import subprocess
+import sys
 
 from catkin_pkg.packages import find_packages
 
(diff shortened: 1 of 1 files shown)


xqms:
Works for me under jenkins. The question is if this is any better than setting the output encoding to UTF-8

```
sys.setdefaultencoding('utf-8')
```

or always using `str.encode('utf-8')`...

jbohren:
> Works for me under jenkins. The question is if this is any better than setting the output encoding to UTF-8

I'm wondering if there are any scenarios where stdout needs to be ASCII, and if in those scenarios, you get a similar exception. As opposed to just bring set to ASCII while still supporting utf-8.

xqms:
> I'm wondering if there are any scenarios where stdout needs to be ASCII

If so, what would be our strategy in that case? Drop all non-ASCII characters?
I'd argue encoding to UTF-8 is a better alternative.

> As opposed to just bring set to ASCII while still supporting utf-8.

Not sure what that is supposed to mean ;-)

jbohren:
>> I'm wondering if there are any scenarios where stdout needs to be ASCII
> If so, what would be our strategy in that case? Drop all non-ASCII characters?
I'd argue encoding to UTF-8 is a better alternative.

Yeah, I think that's fine, as well. It might make sense to set a global flag the first time it happens, though, and print pre-encoded data, 

>>As opposed to just bring set to ASCII while still supporting utf-8.
>Not sure what that is supposed to mean ;-)

*typo: being set to ASCII

From what I understand, Python's understanding of the stdout encoding might not be correct.

jbohren:
Merged via a2cfdf7

Issue #139, opened 2015-01-23
running tests: `catkin test` alias doesn't always work


k-okada:
Hi, I have packages with rostest code which works ok with `catkin run_tests` but fails on `catkin test`, is it correct behavior?
Please see https://gist.github.com/anonymous/f5296c1ae4c5e778c2db

wjwwood:
The `test` target is setup by CTest, as I explained in this ROS answers question (which you asked :stuck_out_tongue:):

http://answers.ros.org/question/136951/how-to-specify-test-packge-in-catkin_make-test/

At any rate the `test` target should work, but I couldn't tell you why it's failing, since you didn't post any error output. Can you please add the output from the failure?

wjwwood:
Sorry I missed the part where you said `catkin test` versus `catkin_make test`. I think the `test` alias (which is what you are invoking) is just incorrect. It would still help to have your error output to confirm that is the issue.

wjwwood:
I'm not sure what the expected behavior is, @jbohren added it here:

https://github.com/catkin/catkin_tools/commit/8803c8964b0ae9805443f8eaaff3ccca176c5a86#diff-954512ee3976ccebcf65cb4af2cbcd3e

Maybe he can comment on what it's supposed to be doing.

k-okada:
`catkin run_tests` works ok but `catkin test` returns error
```
k-okada@kokada-virtual-machine:/tmp/aaa$ catkin test
==> Expanding alias 'test' from 'catkin test' to 'catkin build --verbose --make-args test --'
---------------------------------------------
Profile:                     default
Extending:          [cached] /opt/ros/indigo
Workspace:                   /tmp/aaa
Source Space:       [exists] /tmp/aaa/src
Build Space:        [exists] /tmp/aaa/build
Devel Space:        [exists] /tmp/aaa/devel
Install Space:     [missing] /tmp/aaa/install
DESTDIR:                     None
---------------------------------------------
Isolate Develspaces:         False
Install Packages:            False
Isolate Installs:            False
---------------------------------------------
Additional CMake Args:       None
Additional Make Args:        test
Additional catkin Make Args: None
---------------------------------------------
Workspace configuration appears valid.
--------------------------------------------- 
Found '1' packages in 0.0 seconds. 
Starting ==> testtest                                                                                           
[testtest] ==> '/tmp/aaa/build/testtest/build_env.sh /usr/bin/make cmake_check_build_system' in '/tmp/aaa/build/testtest' 
[...]

PR #265, opened 2015-12-24
[PR196+] jobs: Fixing environment required to run catkin test targets on pre-indigo catkin

jbohren:
This resolves issues described in #139 (i.e. running `catkin test` on Hydro). It does this by doing two things:

1. Overriding the `05.catkin-test-results.sh` file with the `06-ctr-nuke.sh` file after the Catkin CMake configure stage
2. Properly setting the `CATKIN_TEST_RESULTS_DIR` and `ROS_TEST_RESULTS_DIR` environment variables when running Catkin build `make` stage when `test` is in `make_args` 

What's nice is that this only requires changes in the catkin build job definition.

### DIFF
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -37,7 +37,24 @@ from .job import get_package_build_space_path
 from .job import makedirs
 
 
+CTR_NUKE_SH = """
+#!/usr/bin/env sh
+unset CATKIN_TEST_RESULTS_DIR
+unset ROS_TEST_RESULTS_DIR
+"""
+
+
+def ctr_nuke(logger, event_queue, prefix):
+    ctr_nuke_path = os.path.join(prefix, 'etc', 'catkin', 'profile.d')
+    ctr_nuke_filename = os.path.join(ctr_nuke_path, '06-ctr-nuke.sh')
+    mkdir_p(ctr_nuke_path)
+    if not os.path.exists(ctr_nuke_filename):
+        with open(ctr_nuke_filename, 'w') as ctr_nuke_file:
+            ctr_nuke_file.write(CTR_NUKE_SH)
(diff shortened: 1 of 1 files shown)


catkin_tools/jobs/catkin.py
```diff
@@ -102,6 +130,7 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
         'make',
         [MAKE_EXEC] + make_args,
         cwd=build_space,
+        env=ctr_env if 'test' in make_args else {},
```
wjwwood: Should the `ctr_env` extend the current env rather than replace it? Something like this:

```python
new_env = dict(os.environ)
new_env.update(ctr_env if 'test' in make_args else {})
# ...
        env=new_env,
```

jbohren: Yeah, that's the behavior of the `CommandStage` see: https://github.com/catkin/catkin_tools/pull/249/files#diff-62ab2c322956d57c062b8f2b885c4c9cR42

wjwwood: I see, that's a little confusion since other well known function signatures use `env` as an environment replacement. You might consider renaming this argument to `extra_env` or similar to make it clearer. But this looks ok as is.

wjwwood:
Can you put a comment some where (where ever you think most appropriate) in this pr which describes what is happening here and why we have to do it (perhaps referencing the related issues)?

wjwwood:
Never mind, I added the comments. I'll merge after CI catches up.

catkin_tools/jobs/catkin.py
jbohren: Yeah, I can change it so it's more obvious.

jbohren:
Rebased after rebasing #249.

PR #268, opened 2016-01-09
getting all args in bash is '$@'

k-okada:
without this fix, we'll get  following error on catkin 0.3.1
```
# source `catkin --locate-extra-shell-verbs`
$ catkin cd
bash: ${@[*]}: bad substitution
```

maybe original developer  using zsh,  I have checked patched version works both zsh and bash


wjwwood:
Makes sense to me. I'll go ahead and merge it, but @jbohren please take a look when you get a chance to make sure you agree.

PR #250, opened 2015-12-16
[PR196 *] Adding documentation for forthcoming changes in 0.4.0

jbohren:
**View online here: http://jbohren-ct.readthedocs.org/en/pre-0.4.0-docs/**

Includes the following:
 * new execution framework design
 * sets local sphinx style to readthedocs style
 * reorganizes main index for easier navigation

To build locally:
```bash
cd docs
make html
```

### Outstanding Issues

- [x] Add documentation on running unit tests (i.e. installing catkin proper) ([comment](https://github.com/catkin/catkin_tools/pull/249#issuecomment-165452329)) (added in b1e4315134abc62e223f3deaa9f32a62e769da8d)
- [x] Add migration guide / comparison guide (#22) (6dae6ce)
- [x] Add documentation for `--mem-limit` (6dae6ce)
- [x] Add documentation for all new features added by #249
 - [x] DESTDIR support
 - [x] update references to build_env.sh
 - [x] document --get-env option
 - [x] document repro lines
 - [x] document extended summary
 - [x] document .catkin file generation
 - [x] document interleaved mode
 - [x] Add "advanced" documentation on the `--env-cache` a.k.a. "turbo" option  (6dae6ce)
- [x] document new features added by #276
 - [x] using linked develspace
 - [x] changes to CLIs
 - [x] develspace layout / architecture
- [x] add troubleshooting note on `-isystem` and `-I` option precedence with overlays
- [x] adds section for external verbs #71

PR #271, opened 2016-01-12
[PR196 1/3+] Generate .catkin develspace manifest file

jbohren:
Taking over responsibility for populating and updating the `.catkin` devel manifest file. See: https://github.com/catkin/catkin_tools/pull/249#issuecomment-170764943

This also makes catkin "parallel-safe" mode unnecessary, i.e. https://github.com/ros/catkin/issues/675

@cmansley

### DIFF
--- a/catkin_tools/verbs/catkin_build/build.py
+++ b/catkin_tools/verbs/catkin_build/build.py
@@ -445,9 +445,43 @@ def build_isolated_workspace(
             else:
                 break
 
+    # Populate .catkin file if we're not installing
+    # NOTE: This is done to avoid the Catkin CMake code from doing it,
+    # which isn't parallel-safe. Catkin CMake only modifies this file if
+    # it's package source path isn't found.
+    if not context.install:
+        wide_log("[build] Checking devel manifest...")
+        dot_catkin_file_path = os.path.join(context.devel_space_abs, '.catkin')
+        # If the file exists, get the current paths
+        if os.path.exists(dot_catkin_file_path):
+            dot_catkin_paths = open(dot_catkin_file_path, 'r').read().split(';')
+        else:
+            dot_catkin_paths = []
+
+        # Update the list with the new packages (in topological order)
+        packages_to_be_built_paths = [
+            path
+            for path, pkg in packages_to_be_built
+        ]
+
+        new_dot_catkin_paths = [
+            os.path.join(context.source_space_abs, path)
+            for path, pkg in all_packages
+            if path in dot_catkin_paths or path in packages_to_be_built_paths
+        ]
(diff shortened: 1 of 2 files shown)


catkin_tools/verbs/catkin_build/build.py
```diff
@@ -445,9 +445,43 @@ def build_isolated_workspace(
...
+        packages_to_be_built_paths = [
+            path
+            for path, pkg in packages_to_be_built
+        ]
+
+        new_dot_catkin_paths = [
+            os.path.join(context.source_space_abs, path)
+            for path, pkg in all_packages
+            if path in dot_catkin_paths or path in packages_to_be_built_paths
+        ]
```
wjwwood: I'm not sure I care for this approach. The problem I have with it (unless I'm not interpreting the change correctly) is that packages which have not been built yet are already in the `.catkin` file when the first package is built. This is different than what happens normally (without environment caching) and I'm not sure what the consequences might be. At the very least I think it could lead to subtle bugs which only show up (or are only suppressed) when using catkin_tools, but do not when using something like catkin_make_isolated.

jbohren: To clarify, this will only add packages which are requested to be built. 

It's different than `catkin_make_isolated` (where each package gets added once it's configured and built), but more similar to `catkin_make` (where each package gets added during the configure stage).

I agree that this behavior is slightly different, but it's only slightly different on the first build. Successive builds with CM and CMI work with an already-populated `.catkin` file. If anything this makes the conditions of initial builds more similar to successive builds.

What we do know is that letting Catkin CMake populate the file in parallel has lead to malformed `.catkin` files.

wjwwood: I agree that the parallel access to the `.catkin` file is bad, but I'm still not convinced that env hooks shouldn't be able to calculate changes to an environment variable based on the state of the system which might change after each package is installed to the destination.

The reason I say that is that in the case of the parallel access there is a better approach to achieve the same goal. But for the case of dynamic env hooks, if someone is using them, (for example to calculate something based on what packages are installed to the destination), then there is no good alternative without support from the building tool.

This is what brings caching the environment hook changes into question for me.

Since you added an option to avoid env caching, however, I think this is about ready to merge.

jbohren: > Since you added an option to avoid env caching, however, I think this is about ready to merge.

Yeah, I agree it should be optional for now. I'll make sure to add some documentation on it and pros/cons over in #250. Anything else this needs before merge? I also like this implementation much better than the one that supports the linked devel space in #196.

Issue #225, opened 2015-08-05
Invoking `catkin config` (no arguments) clears out Make Args


tprk77:
I am using `0.3.0` installed from `pip`.

I can save `make` arguments using `catkin config ...` as expected, but if I try to read the config using `catkin config` (no arguments) the make arguments get cleared out. CMake arguments are not affected, if that helps. For example:

```text
tperkins@bluedragon:~/test_ws$ catkin config --make-args -j1 -l1 -- --cmake-args -DWHATEVER=1 --
-----------------------------------------------------------------
Profile:                     default
Extending:                   None
Workspace:                   /data/users/tperkins/test_ws
Source Space:       [exists] /data/users/tperkins/test_ws/src
Build Space:        [exists] /data/users/tperkins/test_ws/build
Devel Space:        [exists] /data/users/tperkins/test_ws/devel
Install Space:      [exists] /data/users/tperkins/test_ws/install
DESTDIR:                     None
-----------------------------------------------------------------
Isolate Develspaces:         False
Install Packages:            False
Isolate Installs:            False
-----------------------------------------------------------------
Additional CMake Args:       -DWHATEVER=1
Additional Make Args:        -j1 -l1
Additional catkin Make Args: None
Internal Make Job Server:    True
-----------------------------------------------------------------
Whitelisted Packages:        None
Blacklisted Packages:        None
-----------------------------------------------------------------
Workspace configuration appears valid.
-----------------------------------------------------------------
tperkins@bluedragon:~/test_ws$ cat .catkin_tools/default/config.yaml 
blacklist: []
build_space: build
catkin_make_args: []
cmake_args:
- -DWHATEVER=1
devel_space: devel
extend_path: null
install: false
install_space: install
isolate_devel: false
isolate_install: false
make_args:
- -j1
- -l1
source_space: src
use_internal_make_jobserver: true
whitelist: []
tperkins@bluedragon:~/test_ws$ catkin config
-----------------------------------------------------------------
Profile:                     default
Extending:                   None
Workspace:                   /data/users/tperkins/test_ws
Source Space:       [exists] /data/users/tperkins/test_ws/src
Build Space:        [exists] /data/users/tperkins/test_ws/build
Devel Space:        [exists] /data/users/tperkins/test_ws/devel
Install Space:      [exists] /data/users/tperkins/test_ws/install
DESTDIR:                     None
-----------------------------------------------------------------
Isolate Develspaces:         False
Install Packages:            False
Isolate Installs:            False
[...]

Issue #224, opened 2015-08-03
catkin build fails to find package if its only a <run_depend>


kodyensley:
My group is considering catkin_tools (we use 'standard' catkin now). After running the quickstart for catkin_tools, (http://catkin-tools.readthedocs.org/en/latest/quick_start.html#tl-dr), I now have four packages in my workspace, pkg_a, pkg_b, pkg_c, pkg_d. As per the guide, pkg_d has a build AND run dependency on pkg_a and pkg_b. (This all builds as expected.)

My issue comes when I modify the package.xml for pkg_d. 
```
<buildtool_depend>catkin</buildtool_depend>
<build_depend>pkg_a</build_depend>
<!--  <build_depend>pkg_b</build_depend> -->
<run_depend>pkg_a</run_depend>
<run_depend>pkg_b</run_depend>
```
This now says that pkg_a is a build and run dependency for pkg_d, but that pkg_b is only a run dependency. 

After running <b>catkin clean --all</b> for a fresh start, I run <b>catkin build pkg_d</b> and am greeted with the following error:
```
user@computer:/tmp/catkin_workspace_test$ catkin build pkg_d 
Creating build space directory, '/tmp/catkin_workspace_test/build' 
------------------------------------------------------------------------------------------------
Profile:                     default
Extending:             [env] /opt/ros/indigo
Workspace:                   /tmp/catkin_workspace_test
Source Space:       [exists] /tmp/catkin_workspace_test/src
Build Space:        [exists] /tmp/catkin_workspace_test/build
Devel Space:       [missing] /tmp/catkin_workspace_test/devel
Install Space:     [missing] /tmp/catkin_workspace_test/install
DESTDIR:                     None
------------------------------------------------------------------------------------------------
Isolate Develspaces:         False
Install Packages:            False
Isolate Installs:            False
------------------------------------------------------------------------------------------------
Additional CMake Args:       None
Additional Make Args:        None
Additional catkin Make Args: None
Internal Make Job Server:    True
------------------------------------------------------------------------------------------------
Whitelisted Packages:        None
Blacklisted Packages:        None
------------------------------------------------------------------------------------------------
Workspace configuration appears valid.
------------------------------------------------------------------------------------------------ 
Found '4' packages in 0.0 seconds. 
Starting ==> pkg_a                                                                                                                                                 
[...]

Issue #65, opened 2014-06-10
build: --isolate-devel crashes when not building all packages


NikolausDemmel:
The following failiure occurs in a workspace with more than one package and no existing devel folder (the use of the "--space-suffix" and "--source" options is irrelevant to this issue I assume):

```
$ catkin build --space-suffix test --source srctest --isolate-devel catkin
Creating buildspace directory, '/Users/demmeln/work/indigo_ws/buildtest'
----------------------------------------------------------------------
Workspace:                   /Users/demmeln/work/indigo_ws
Buildspace:                  /Users/demmeln/work/indigo_ws/buildtest
Develspace:                  /Users/demmeln/work/indigo_ws/develtest
Installspace:                /Users/demmeln/work/indigo_ws/installtest
DESTDIR:                     None
----------------------------------------------------------------------
Isolate Develspaces:         True
Install Packages:            False
Isolate Installs:            False
----------------------------------------------------------------------
Additional CMake Args:       None
Additional Make Args:        None
Additional catkin Make Args: None
----------------------------------------------------------------------
Found '2' packages in 0.0 seconds.
Starting ==> catkin
Finished <== catkin [ 3.3 seconds ]
[build] Runtime: 3.5 seconds                                                                                                                                                                                  [0/0 Active | 1/1 Completed]
Traceback (most recent call last):
  File "/usr/local/bin/catkin", line 9, in <module>
    load_entry_point('catkin-tools==0.0.0', 'console_scripts', 'catkin')()
  File "/Users/demmeln/work/catkin_tools/catkin_tools/commands/catkin.py", line 192, in main
    sys.exit(args.main(args) or 0)
  File "/Users/demmeln/work/catkin_tools/catkin_tools/verbs/catkin_build/cli.py", line 183, in main
    lock_install=not opts.no_install_lock
  File "/Users/demmeln/work/catkin_tools/catkin_tools/verbs/catkin_build/build.py", line 548, in build_isolated_workspace
    _create_unmerged_devel_setup(context)
  File "/Users/demmeln/work/catkin_tools/catkin_tools/verbs/catkin_build/build.py", line 233, in _create_unmerged_devel_setup
    first_source=leaf_sources[0],
IndexError: list index out of range
```

jbohren:
I took a moment to look into this. Now that `catkin build` computes a list of "unbuilt" packages, it makes it easy to avoid the leaf-sources error. See #276 for the patch.

jbohren:
Closed with #276

PR #276, opened 2016-01-21
[PR196 2/3] Adding linked develspace support

### LINKED ISSUE #45: leaf_sources out of bounds error
I'm on OSX Mavericks. I have a workspace with many packages and I'm debugging the build of one or two. When I delete the ```build``` and ```devel``` folders and try to build a single package with very few dependencies, I get an out of bounds error when writing the setup.bash script:

```
catkin build glog_catkin
-----------------------------------------------------------------------------------
Workspace:                   /Users/ptf/Work/ethz/code/ros_hydro_overlay_ws
Buildspace:                  /Users/ptf/Work/ethz/code/ros_hydro_overlay_ws/build
Develspace:                  /Users/ptf/Work/ethz/code/ros_hydro_overlay_ws/devel
Installspace:                /Users/ptf/Work/ethz/code/ros_hydro_overlay_ws/install
DESTDIR:                     None
-----------------------------------------------------------------------------------
Merge Develspaces:           False
Install Packages:            False
Isolate Installs:            False
-----------------------------------------------------------------------------------
Additional CMake Args:       None
----------------------------------------------------------------------------------- 
Found '73' packages in 0.1 seconds. 
Starting ==> catkin_simple                                                                              
Finished <== catkin_simple [ 0.1 seconds ]                                                              
Starting ==> gflags_catkin                                                                              
Finished <== gflags_catkin [ 0.4 seconds ]                                                              
Starting ==> glog_catkin                                                                                
Finished <== glog_catkin   [ 0.7 seconds ]                                                              
[build] Runtime: 1.9 seconds                                               [0/0 Active | 3/3 Completed] 
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
/usr/local/bin/catkin in <module>()
      7 if __name__ == '__main__':
      8     sys.exit(
----> 9         load_entry_point('catkin-tools==0.0.0', 'console_scripts', 'catkin')()
     10     )

/Library/Python/2.7/site-packages/catkin_tools-0.0.0-py2.7.egg/catkin_tools/commands/catkin.pyc in main(sysargs)
    190     # Finally call the subparser's main function with the processed args
    191     # and the extras which the preprocessor may have returned
--> 192     sys.exit(args.main(args) or 0)

/Library/Python/2.7/site-packages/catkin_tools-0.0.0-py2.7.egg/catkin_tools/verbs/catkin_build/cli.pyc in main(opts)
    181             interleave_output=opts.interleave_output,
    182             no_status=opts.no_status,
--> 183             lock_install=not opts.no_install_lock
    184         )
    185     finally:

/Library/Python/2.7/site-packages/catkin_tools-0.0.0-py2.7.egg/catkin_tools/verbs/catkin_build/build.pyc in build_isolated_workspace(context, packages, start_with, no_deps, jobs, force_cmake, force_color, quiet, interleave_output, no_status, lock_install)
    539         if not context.merge_devel:
    540             if not context.install:
--> 541                 _create_unmerged_devel_setup(context)
    542             else:
    543                 _create_unmerged_devel_setup_for_install(context)

/Library/Python/2.7/site-packages/catkin_tools-0.0.0-py2.7.egg/catkin_tools/verbs/catkin_build/build.pyc in _create_unmerged_devel_setup(context)
    231 {run_depends_sources}
    232 """.format(
--> 233         first_source=leaf_sources[0],
    234         leaf_sources='\n'.join(leaf_sources[1:]),
    235         run_depends_sources='\n'.join(run_depends_sources)

IndexError: list index out of range
```
If I run this in ipython and call debug, it looks like ```leaf_packages``` has a lot of packages from my src folder
```
leaf_packages
['rviz_animated_view_controller', 'aslam_vcharge_python', 'imu_camera_calibration', 'sm_doc', 'libsnappy', 'zeromq_catkin', 'aslam_python', 'aslam_camera_calibration', 'aslam_install', 'ceres_catkin', 'sm_matlab', 'protobuf_catkin', 'rviz_pyplot', 'aslam_vo_python', 'aslam_backend_bsplines_tutorial', 'aslam_cv_ros', 'aslam_simulation_python']
```
but as none of those are built, the ```setup.sh``` files are not there and ```leaf_sources``` ends up empty.

The same problem doesn't happen when I build with the ```--merge-devel``` option.

### LINKED ISSUE #87: No setup.bash

### LINKED ISSUE #148: configuration option to make "catkin build" work like --only-pkg-with-deps
I'm in the habit of using catkin_make's "--only-pkg-with-deps" argument to ensure I get a fully-populated workspace: both build_depends and exec_depends for a given package.  When I run "catkin build package_name", package_name's exec_depends aren't built, though.  This makes it problematic to use, as my package isn't functional until its exec_depends are present.  (Or am I missing something or doing something wrong?)

### LINKED ISSUE #227: CMAKE_EXPORT_COMPILE_COMMANDS fails if the build target directory already exists
I need to generate a compile_commands.json for my package so I ran `catkin build mypkg -DCMAKE_EXPORT_COMPILE_COMMANDS=ON` but this did not generate the compile_commands.json file like I expected. I think this was because the target directory already existed because when I deleted the target directory (build/mypkg), the file was generated

### LINKED ISSUE #272: Add version info and --version option to catkin command
Currently there's no easy way to get the version from the `catkin` command. It should be listed in the full help, as well as with `catkin --version`.

jbohren:
This adds the prototype linked develspace support from #196. 

## Why?

Catkin packages build their products into a usable FHS tree called a "develspace." We'd like to be able to clean the products from a specific package. Unfortunately, it's impossible to determine which files were generated by which package without some OS-specific hacks or wrapper programs such as `checkinstall`. See #196 and #194 for more details on the development and history of this implementation.

## Basic Design

This PR implements the same linked develspace support  and the same workarounds as in #196. It adds the additional functions required for the forthcoming enhancements to the `clean` verb, but does change it.

This adds a new `linked` develspace layout, which appears to the user as a merged develspace. This first builds Catkin packages into a *hidden* isolated develspace, and then symbolically links all of the created files into the *visible* merged develspace. All of these files are listed in a `devel_manifest.txt` file stored in the develspace. If two packages generate files with the same path, the linking code checks if the two files are identical. If so, it increments a counter in a `devel_collisions.txt` file. If they are different, it raises a warning. Pure CMake and other build types which support install logs can simply use their normal install targets with the visible develspace.

Making this work requires a major workaround due to the way Catkin generates the develspace setup files. Catkin's CMake macros generate setup files which encode the actual *hidden* isolated develspace. This means that they can't be used in the visible merged develspace. This means that these files need to be generated by some other means. The solution used by this PR is to run a "prebuild" step which builds either a minimal catkin package called `catkin_tools_prebuild` or if it exists in the source space, the `catkin` CMake package, itself.

This does have another benefit, which is that workspaces containing only non-Catkin packages will still have the normal Catkin setup files.  See #87 for an example use case.

## Interface

Since this PR includes three options for the develspace layout, the `catkin_tools` `Context` reflects the develspace layout with a string identifier which is either `merged`, `isolated`, or `linked`. This is reflected in the `catkin config` options and the context summary:

```
Devel Space:
  Options for configuring the structure of the devel space.

  --merge-devel         Build products from each catkin package into a single
                        merged devel spaces.
  --link-devel          Build products from each catkin package into isolated
                        spaces, then symbolically link them into a merged
                        devel space.
  --isolate-devel       Build products from each catkin package into isolated
                        devel spaces.
```

```
--------------------------------------------------
Extending:                   None
Workspace:                   /home/jbohren
Source Space:      [missing] /home/jbohren/src
Build Space:       [missing] /home/jbohren/build
Devel Space:       [missing] /home/jbohren/devel
Install Space:     [missing] /home/jbohren/install
DESTDIR:                     None
--------------------------------------------------
Devel Space Layout:          merged
Install Packages:            False
Isolate Installs:            False
--------------------------------------------------
Additional CMake Args:       None
Cache Job Environments:      False
--------------------------------------------------
Whitelisted Packages:        None
Blacklisted Packages:        None
--------------------------------------------------
```

## Additional Changes

* Moves common functionality out of poorly named `catkin_tools.jobs.job` module into `catkin_tools.common` and `catkin_tools.jobs.utils`.
* Clarifies and extends `context` path functions
* Gets rid of no-longer-used `build_env.sh` generation source
* Fixes some line-break handling and path formatting issues with the CMake io handler
* Fixes the dependency gathering functions so they preserve topological order
* Fixes #45 / #65 by using the unbuilt list for the isolated devel space list
* Fixes #148, #224 by including exec and run dependencies of top-level packages
* Fixes #227 by automatically calling `--force-cmake` if `cmake_args` changes
* Fixes regression #225
* Adds version, fixes #272 
* Fixes #87 if user uses a linked devel space

## Tasks

- [x] Incorporate `exec_depend` dependencies into build deps #278, #148
- [x] Review first draft by @wjwwood 
- [x] Use `dateutil` for copyright dates
- [x] Fix trailing commas / etc in https://github.com/catkin/catkin_tools/commit/d1ef3e1506c5e0ca29d463b14ef20c076a278289
- [x] Update autocomplete args
- [ ] Fix error / warning when a package depends on a metapackage (https://github.com/catkin/catkin_tools/pull/276#issuecomment-177517793)


jbohren:
@wjwwood This is good for review when you're ready. Until you get back on this, I'm going to update documentation of #249 over in #250.

jbohren:
@wjwwood I updated the isolated install stuff in this PR because we might want to use it instead of the "prebuild" ghost package to generate setup files. I think this is probably the most contentious part of the design. 

On one hand, if we use the prebuild package, it's messy, but we can always be sure that Catkin is generating the setup files for the linked develspace. On the other hand, we could generate them similarly to how the isolated devel pattern does, but then we need to make sure the setup files which are generated do everything that the Catkin ones do.

Currently, the first pattern is implemented, because that's what was implemented in #196, but if you think it's an ok burden to maintain the setup files, we could just have `catkin_tools` generate them, instead.

mikepurvis:
I tried this branch in my project and there's an ordering problem where catkin itself doesn't get built first, when it's part of the workspace, so everything following it obviously crashes and burns.

I can try to prepare a MWE later on if that would be helpful.

jbohren:
> I tried this branch in my project and there's an ordering problem where catkin itself doesn't get built first, when it's part of the workspace, so everything following it obviously crashes and burns.

@mikepurvis Thanks for giving this a shot, that bug should be fixed in #39edcdb. It was always removing `catkin` from the packages to build, even when it wasn't prebuilding it.

wjwwood:
The changes related to package ordering have enough issues that I think we should do it in a separate pr with more tests to assert it is behaving correctly (I'll help with that).

Am I correct in understanding that the linked devel space does not need those changes to work correctly?

jbohren:
> The changes related to package ordering have enough issues that I think we should do it in a separate pr with more tests to assert it is behaving correctly (I'll help with that).

> Am I correct in understanding that the linked devel space does not need those changes to work correctly?

No it does need them. As it stands now, even, `master` is broken since the resultspace loading assumes that the dependency gathering functions preserve the order of the packages that are passed into them.

wjwwood:
Ok, well lets keep them together for now since this is incrementally better, but more work needs to be done on those functions in the future.

jbohren:
> Ok, well lets keep them together for now since this is incrementally better, but more work needs to be done on those functions in the future.

Alright. I'll remove the `exclude_function` on the next pass, tonight, though, as per https://github.com/catkin/catkin_tools/commit/d1ef3e1506c5e0ca29d463b14ef20c076a278289#commitcomment-15715858

wjwwood:
Actually I'd prefer you didn't for now, and for someone (probably me) to go back and take a comprehensive look at those functions with tests for the corner cases. I think we can leave it as is since that is the behavior from before.

wjwwood:
The code review looks ok otherwise. I'm going to test it out locally (acceptance test it) after lunch and if it looks ok, then I'll merge after you address the two remaining check boxes on the OP.

jbohren:
> Actually I'd prefer you didn't for now, and for someone (probably me) to go back and take a comprehensive look at those functions with tests for the corner cases. I think we can leave it as is since that is the behavior from before.

Roger.

jbohren:
> The code review looks ok otherwise. I'm going to test it out locally (acceptance test it) after lunch and if it looks ok, then I'll merge after you address the two remaining check boxes on the OP.

Alright. You're fine with the hacky prebuild package pattern instead of generating the setup files directly?

wjwwood:
Yeah, I think that's ok.

We do things differently in `ament`, where all of the generated file's templates are in a pure python package like `catkin_pkg` so they're accessible from both Python and CMake and we have a Python function that can emulate the `configure_file()` function in CMake. Also, as a matter of design, the file generated by an individual package do not include file which collide with another package, i.e. it creates files in `<prefix>/share/<pkg name>` but not directly in `<prefix>`. The files that are dropped in the root of the the prefix are generated once by the tool. This pattern seems to be working out, so maybe we can back port some of these concepts to `catkin` in the future.

jbohren:
> Yeah, I think that's ok.

> We do things differently in ament, where all of the generated file's templates are in a pure python package like catkin_pkg so they're accessible from both Python and CMake and we have a Python function that can emulate the configure_file() function in CMake. Also, as a matter of design, the file generated by an individual package do not include file which collide with another package, i.e. it creates files in <prefix>/share/<pkg name> but not directly in <prefix>. The files that are dropped in the root of the the prefix are generated once by the tool. This pattern seems to be working out, so maybe we can back port some of these concepts to catkin in the future.

Alright, sounds good.

wjwwood:
@jbohren I'm immediately running into an issue where packages do not find catkin (even though it was built and is actually in the workspace). I was using `--install` at the same time as `--link-devel`. I'll keep trying different combinations.

wjwwood:
Without `--install` it gets past that issue.

Also, we might need to update the shell tab completion with the new options to `catkin config`.

jbohren:
@wjwwood Are you on the latest 9805079?

wjwwood:
```
% git log -1
commit 9805079774fc3958bb51746e417ef0628de8d31c
Author: Jonathan Bohren <jbo@jhu.edu>
Date:   Tue Jan 26 23:56:43 2016 -0500

    tests: Updating mock for package format 2
```

jbohren:
@wjwwood It was generating a "prebuild" job for `catkin` instead of a normal one. That should be fixed in  0124fcf.

wjwwood:
Ok, that cleared it up for me. My acceptance testing is going well, I'm just finishing up testing `--link-devel` with `--install` and I think we'll be good to go.

wjwwood:
When using the `--install` option there doesn't seem to be any `DYLD_LIBRARY_PATH` set. So `roscore`, for example, outputs stuff like this:

```
process[rosout-1]: started with pid [18787]
[rosout-1] process has died [pid 18787, exit code -5, cmd /Users/william/indigo/install/lib/rosout/rosout __name:=rosout __log:=/Users/william/.ros/log/7ed96514-c548-11e5-b796-a45e60b91273/rosout-1.log].
log file: /Users/william/.ros/log/7ed96514-c548-11e5-b796-a45e60b91273/rosout-1*.log
[rosout-1] restarting process
dyld: Library not loaded: libroscpp.dylib
  Referenced from: /Users/william/indigo/install/lib/rosout/rosout
  Reason: image not found
```

jbohren:
> When using the `--install` option there doesn't seem to be any `DYLD_LIBRARY_PATH` set. So `roscore`, for example, outputs stuff like this:

Is that a regression introduced by this PR?

wjwwood:
Not sure, I'd have to build with an older version. Can you check that the `LD_LIBRARY_PATH` is set on your machine when using `--install` and sourcing `install/setup.zsh`?

Another thing I noticed is that linked devel space does not have `.bash` or `.zsh` setup files. Is that intentional?

jbohren:
> Not sure, I'd have to build with an older version. Can you check that the `LD_LIBRARY_PATH` is set on your machine when using `--install` and sourcing `install/setup.zsh`?

If that OSRF OS X box is still up I can bisect that tonight.

> Another thing I noticed is that linked devel space does not have `.bash` or `.zsh` setup files. Is that intentional?

No, that's a regression caused by fixing the `--install` issue that you  had earlier. If the `catkin` package is in the workspace, it needs to be built without a linked develspace, which is why it was being built as a prebuild package. I can fix that correctly later tonight.

jbohren:
> When using the `--install` option there doesn't seem to be any `DYLD_LIBRARY_PATH` set. So `roscore`, for example, outputs stuff like this:

I couldn't reproduce this on `dosa`. Does this work on your machine with `catkin_make`?

mikepurvis:
Works great for me as of 6bc286af0e41a7f9bce79e981413bed72c712f1c.

mikepurvis:
Actually, one very small complaint— when a non-metapackage depends on a metapackage, there error message is rather unhelpful:

```
Abandoned <<< map_interface                          [ Depends on unknown jobs: rosbridge_suite ]
```

This is an error, but the message in this scenario could potentially be improved to recognize that `rosbridge_suite` is totally a valid package, and it's in the current workspace, it's just not valid for `map_interface` as a non-metapackage to depend on it.

jbohren:
@mikepurvis Thanks for the continued QA testing!

wjwwood:
That issue I reported is certainly to do with the new security thing in 10.11:

```
william@farl:/tmp
$ env | grep DYLD
william@farl:/tmp
$ . test.sh
william@farl:/tmp
$ env | grep DYLD
william@farl:/tmp
$ sw_vers -productVersion
10.11.2
william@farl:/tmp
$ cat test.sh
export DYLD_LIBRARY_PATH=/tmp
```

And on `dosa`:

```
dosa:tmp osrf$ env | grep DYLD
dosa:tmp osrf$ . test.sh
dosa:tmp osrf$ env | grep DYLD
DYLD_LIBRARY_PATH=/tmp
dosa:tmp osrf$ sw_vers -productVersion
10.10.5
dosa:tmp osrf$ cat test.sh
export DYLD_LIBRARY_PATH=/tmp
```

I'm gonna do a few more things and I'll merge this. We'll have to come back around to this issue.

wjwwood:
Let's ticket the last issue brought up by @mikepurvis, I'm going to merge this now.

wjwwood:
Thanks @jbohren :smile:

PR #281, opened 2016-01-22
build: Making ctr-nuke env hook go to the right place, fixes #280.

### LINKED ISSUE #280: DESTDIR broken on master
Looks like it mostly does the right thing, but tries to create the initial path in the un-prefixed location. Previously I needed the following change:

```
diff --cc catkin_tools/verbs/catkin_build/job.py
index 1fcb3b5,1fcb3b5..e41a437
--- a/catkin_tools/verbs/catkin_build/job.py
+++ b/catkin_tools/verbs/catkin_build/job.py
@@@ -229,6 -229,6 +229,8 @@@ class CMakeJob(Job)
          # Determine the location of where the setup.sh file should be created
          if self.context.install:
              setup_file_path = os.path.join(install_space, 'setup.sh')
++            if self.context.destdir:
++                setup_file_path = self.context.destdir + setup_file_path
              if not self.context.isolate_install and os.path.exists(setup_file_path):
                  return commands
          else:  # Create it in the devel space
```

But I'm having trouble getting this work with the new backend. I tried making [similar changes here](https://github.com/catkin/catkin_tools/blob/c5c88071750c4d66253faf83bbd3c312c9f1576e/catkin_tools/jobs/cmake.py#L178), but it seemed to continue to fail in the same way.

jbohren:


mikepurvis:
:+1:

PR #282, opened 2016-01-24
Add trollius dependency in requirements.txt

kartikmohta:
I think this got added in the source recently and is missing from the requirements.txt file

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ catkin_pkg
 setuptools
 sphinxcontrib-ansi
 sphinxcontrib-programoutput
+trollius


kartikmohta:
It seems like there are more dependencies required right now (osrf_pycommon).

jbohren:
@kartikmohta Yeah, `osrf_pycommon` still needs to be released. See https://github.com/osrf/osrf_pycommon/issues/9 if you want to help out with that.

wjwwood:
Even still `trollius` is used directly by catkin tools so I think we should merge this.

It should also have `osrf_pycommon`, but as @jbohren pointed out, it hasn't been released yet. I'll have to do that before the next release of catkin tools since master uses it.

PR #284, opened 2016-01-29
Fixes the bash completion path for empty prefix.

iwanders:
I noticed this issue yesterday when I installed catkin tools; there was no bash completion out of the box. I did some digging and came to the conclusion that the completion files are installed in the wrong folder. Tested on a fresh trusty64 virtual machine with `sudo python setup.py install -v`:

In the latest version the relevant line from the installation log is:
> copying completion/catkin_tools-completion.bash -> /usr/local/etc/bash_completion.d

Before 9bec9e7bd7dc9d26a4372e14e3911bb175ee04d3, this line was: 
> copying completion/catkin_tools-completion.bash -> /etc/bash_completion.d

For the default empty prefix, I believe it should be the same output as `/usr`'s, such that the installation path of the bash completion file is `/etc/bash_completion.d`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ def _resolve_prefix(prefix, type):
         if sys.prefix.startswith(osx_system_prefix):
             return '/usr/share'
     elif type == 'bash_comp':
-        if prefix == '/usr':
+        if (prefix == '/usr') or (prefix == ''):
             return '/'
         if sys.prefix.startswith(osx_system_prefix):
             return '/'


wjwwood:
If you install with `python setup.py install` then it installs all of the catkin_tools into `/usr/local`. I don't think it should be different for the bash completion scripts. For example, the executable is on `/usr/local/bin` (which means `/usr/local/bin` must be on the PATH) and the Python modules are installed to `/usr/local/lib/python2.7/dist-packages/` (which means that path must be considered by Python by default).

I think the correct approach is to add something like this to your `~/.bashrc`:

```bash
for f in /usr/local/etc/bash_completion.d/*; do
    source $f
done
```

If `catkin-tools` was installed using a `.deb` and was installed to `/usr`, then the correct place would be to install it to `/etc/bash_completion.d` and that's the current behavior.

wjwwood:
To summarize, I believe this is expected and correct behavior, and so I'm going to close this as wont fix.

However, I understand the convenience of the previous behavior and I'd support changes which are some how conceptually consistent but also more convenient. Perhaps pointing to how another project does installation of completion files would be helpful.

I'd also support adding a message at the end of the `setup.py` install when using Ubuntu and installing to `/usr/local` which tells the user what they need to do in order to enable tab completion.

iwanders:
Before submitting the new pull request on this topic, I searched in vain on how other projects handle this: Both Bazaar and Mercurial do not appear to install any autocompletion files when installed by invoking `setup.py` and youtube-dl also installs into `/usr/local/etc/bash_completion.d`.

With the default installation, inside the `PermissiveInstall` instance the value of `self.prefix` actually holds `/usr` and `self.install_data` equals `/usr/local`. Another alternative would be to install the bash completion files to `/usr/share/bash-completion/completions/`, which is still in the `/usr` prefix. This has the advantage that this directory is automatically included for bash completion. However, I am not knowledgeable about whether writing there would be acceptable, so I opted for the solution you proposed.

PR #295, opened 2016-02-15
Add completion for bt alias

wkentaro:


jbohren:
Is it possible to use the output of `catkin -a` to tab-complete to expand aliases?

wkentaro:
Thanks for the comment. I updated the commit. please review.

wkentaro:
friendly ping.

NikolausDemmel:
> Is it possible to use the output of `catkin -a` to tab-complete to expand aliases?

Did you mean actually replacing the alias on the command line with the expanded version upon tab completion? I guess this not address by this PR.

I'm not sure if you'd want expansion or not. Maybe you'd want the simpler alias in your bash history rather than a (possibly convoluted) expanded command. Or maybe you have an alias `bu`, then when typing  `catkin bu[TAB]`, would you expect expansion of the alias `bu`, or completion to the verb `build`? Nevertheless I can still see how auto-expanding aliases could be a nice self-documentation. Is there a solution that is a compromise?

jbohren:
> Nevertheless I can still see how auto-expanding aliases could be a nice self-documentation. Is there a solution that is a compromise?

I think what we'd want is a command similar to `catkin -a` which gives the recursive expansion of an alias. Then the completion script would expand the alias on the line into the aliased command.

For making the aliases more visible, maybe we should show fully expanded aliases with `catkin -h`?

NikolausDemmel:
> I think what we'd want is a command similar to catkin -a which gives the recursive expansion of an alias. Then the completion script would expand the alias on the line into the aliased command.

I'm not sure if you are proposing to complete the alias on the command line as in `catkin bt [tab]` completing to `catkin build --this`, or if you are saying that `catkin bt [tab]` should show the same completions as `catkin build --this [tab]` would, but not expand the alias visible to the user.

Considering this again I believe we should only do the latter. This would also be in line with how completion works for aliases of `git`. I.e. if you have a git alias `re` for `remote`, `git re [tab]` shows you the completion for as if you had typed `git remote [tab]`, but does not expand the alias on the command line.

> Or maybe you have an alias bu, then when typing catkin bu[TAB], would you expect expansion of the alias bu, or completion to the verb build? 

To answer my own question, when you type `catkin b[tab]` it should offer to complete to either `catkin build` or `catkin bt` or any other command or alias that starts with `b`.

completion/catkin_tools-completion.bash
```diff
@@ -54,7 +56,7 @@ _catkin()
   if [[ ${COMP_CWORD} -eq 1 || ${catkin_opts} == *${prev}* ]] ; then
     COMPREPLY=($(compgen -W "${catkin_verbs}" -- ${cur}))
   else
-    if [[ "${COMP_WORDS[@]}" == *" build"* ]] ; then
+    if [[ "${COMP_WORDS[@]}" =~ ".* build|bt.*" ]] ; then
```
wjwwood: This is very fragile. I really don't like the change to include `bt`. When it was `build` at least you knew that was the verb. With this change, this list would need to be extended anytime a new "build based" verb alias was added, otherwise option completion would not work.

The "right" way to do it would be to expand any verb aliases into a local variable (note not replacing it on the command line for the user) and checking what the "real verb" is based on that expansion. Then you could gate build option completion based on the real verb being build. This would work for any verb alias based on build. However, that sounds complicated, so I'd rather merge this without build options working with `bt`, i.e. remove the changes on this line. Then at least the verb aliases would auto complete, but we wouldn't have the inconsistent behavior that some aliases can complete but others cannot.

mikepurvis:
Closing due to inactivity.

PR #299, opened 2016-02-26
Correct typos

fudger:
The catkin_tools docs are excellent and I am sure they will be read a lot. So I felt it is worth pulling a request to correct a few typos I spotted while reading through the docs.

### DIFF
--- a/docs/quick_start.rst
+++ b/docs/quick_start.rst
@@ -4,7 +4,7 @@ Quickstart
 This chapter gives a high-level overview of how to use ``catkin_tools`` and the
 ``catkin`` command. This shows how to use the different command verbs to create
 and manipulate a workspace. For a more in-depth explanation of the mechanics of
-catkin workspaces, see :doc:`Workspace Mechanics <mechanics>`, and for thorogh
+catkin workspaces, see :doc:`Workspace Mechanics <mechanics>`, and for thorough
 usage details see the individual verb documentation.
 
 TL;DR
@@ -79,7 +79,7 @@ contains the current working directory.
 An important property which deserves attention is the summary value labeled
 ``Extending``. This describes other collections of libraries and packages which
 will be visible to your workspace. This is process called "workspace chaining."
-The value can be come from a few different sources, and can be classified in
+The value can come from a few different sources, and can be classified in
 one of the three following ways:
 
 - No chaining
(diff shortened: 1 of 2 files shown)


docs/quick_start.rst
```diff
@@ -145,7 +145,7 @@ Now that there are some packages in the workspace, Catkin has something to build
 .. note::
 
     Catkin utilizes an "out-of-source" and "aggregated" build pattern. This
-    means that not only will temporary or final build products never be placed
+    means that not only will temporary or final build products ever be placed
```
wjwwood: Is this right? Read it aloud I would have said never, but then again it could be more grammatically correct to say ever. Perhaps it should be reworded entirely to avoid the risk of a double negative.

wjwwood:
Other than a comment, lgtm. I will ask that squash your commits before I merge. Thanks!

docs/quick_start.rst
fudger: I think rewording would be best. I did not want to change the meaning, only correct typos, that's why I did not commit the following:
> This means that temporary or final build products will never be placed in a package’s source directory or anywhere in the source space. Instead, all build directories are aggregated in the build space and all final build products like executables, libraries, etc., will be put in the devel space.

wjwwood: The text you posted there sounds right to me. If you'll update the pr and squish the commits I'll merge in the morning. Thanks again!

fudger: Done! Happy to contribute a tiny bit.

jbohren:
@wjwwood can you hold of on merging this, and let me merge the changes into the big doc PR? Otherwise there are going to be a crazy number of conflicts.

wjwwood:
@jbohren sure, just try to preserve the attribution if you can.

jbohren:
@wjwwood yeah I'll do my best.

wjwwood:
@jbohren Is the big issue that you have rewrapped lines in the new documentation?

Recently I've become convinced that one sentence per line (no matter how long it is) is the superior way to write documentation. Simply for this reason (assuming the wrapping is causing the conflicts).

jbohren:
> @jbohren Is the big issue that you have rewrapped lines in the new documentation?

> Recently I've become convinced that one sentence per line (no matter how long it is) is the superior way to write documentation. Simply for this reason (assuming the wrapping is causing the conflicts).

Yeah, I agree that one sentence per line is best, but no, the real reason is that both the mechanics and quickstart docs have changes dramatically.

jbohren:
@fudger I've merged your changes into #250, thanks for the contribution!

PR #305, opened 2016-03-22
Don't fail with devel space layout check on first build.

NikolausDemmel:
Addresses the first part of https://github.com/catkin/catkin_tools/pull/293#issuecomment-199951230


jbohren:
I think I ended up making these changes, independently over in #293. Is this ok to close?

NikolausDemmel:
Sure. Thanks!

PR #307, opened 2016-03-31
normalize extend_path

rhaschke:
... to get rid of trailing or double slashes which cause the consistency check to fail without real reason:

An extend_path `/opt/ros/indigo/` will not be found in catkin's `CMAKE_PREFIX_PATH` which will list the path without the trailing slash.


wjwwood:
lgtm, thanks

PR #313, opened 2016-04-08
catkin_build: fix cmake args check

xqms:
The check as it was would force a cmake run on every package if
cmake_args were saved in the context, since then
ctx.cmake_args != opts.cmake_args.

Since opts.cmake_args are already appended to ctx.cmake_args by
Context.load, the second check is unnecessary.

### DIFF
--- a/catkin_tools/verbs/catkin_build/cli.py
+++ b/catkin_tools/verbs/catkin_build/cli.py
@@ -364,7 +364,7 @@ def main(opts):
     # Get the last build context
     build_metadata = get_metadata(ctx.workspace, ctx.profile, 'build')
 
-    if build_metadata.get('cmake_args') != ctx.cmake_args or build_metadata.get('cmake_args') != opts.cmake_args:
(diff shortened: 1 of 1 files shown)


xqms:
Actually, one could argue that force-cmake is always in the responsibility of the user. If he specifies cmake args only for one run (i.e. does not save them), then catkin_tools has no way of tracking that.

jbohren:
@xqms This has already been fixed in #293 in the exact same way (https://github.com/catkin/catkin_tools/pull/293/files#diff-a07688918e375a3e61cecbe117f4d208R369)

xqms:
@jbohren Ah okay, didn't notice that. However, if #293 will take longer to merge, it might be nice to have this already - this is hurting performance for a lot of users. Otherwise, feel free to close this.

wjwwood:
Just as an update on #293, it's on my agenda for the weekend (probably Sunday).

xqms:
I'm closing this as #293 is merged.

PR #314, opened 2016-04-11
remove apparently unused variables

wjwwood:
In two places there seems to be unused variables. In my experience this either means someone intended to use them but ended up not using them for some reason or they intended to use them and accidentally used something else. @jbohren can you have a look at these and see which is the case. I think we can just merge this as-is.


jbohren:
@wjwwood I rebased this on master and fixed the conflict, it should be good to merge once the tests pass.

wjwwood:
Thanks.

PR #315, opened 2016-04-11
put warned "jid's" in the warneds dict

wjwwood:
@jbohren I think this is the right fix, but I detected it by noticing that the `warneds` variable was unused. This is a great example of why having a tool in your editor to warn you about unused variables is a good thing.

I believe this might be contributing to a bug I noticed, but hadn't yet reported where the notification would say "built successfully without warnings" when in fact there were warnings. Let me know what you think.


jbohren:
@wjwwood I think this is a good fix, but it shouldn't have any effect since the detailed summary doesn't mention the number of packages that produced warnings.

PR #318, opened 2016-04-14
pre-0.4.0 Documentation

jbohren:
Rebased PR thanks to @wjwwood to supersede #250.


jbohren:
@wjwwood It looks like the `enchant` library is available on OS X for Python2 but not Python3. This might be a bug with PyEnchant 1.6.6: https://github.com/rfk/pyenchant/pull/63

wjwwood:
Can you just conditionally import it in the `conf.py`?

jbohren:
> Can you just conditionally import it in the conf.py?

Possibly, now that it's passing on all CI variants. I just wanted to do something fast to get it working.

jbohren:
@wjwwood d2809e8 drops the extra sphinx conf, and conditionally uses spelling based on the tag supplied with `-t` at the cli.

jbohren:
@wjwwood All checks passing.

jbohren:
And building on RTD: http://catkin-tools.readthedocs.org/en/pre-0.4.0-docs-rebase/

wjwwood:
While reviewing the documentation I noticed: https://github.com/catkin/catkin_tools/issues/319

wjwwood:
Pushed some commits.

wjwwood:
I'm gonna merge this and we can fix up things in follow up PR's and issues.

Thanks @jbohren this is a really great improvement to the docs. :tada:

wjwwood:
@jbohren looks like RTD is failing on master after merging this:

https://readthedocs.org/projects/catkin-tools/builds/3910003/

Something about duplicate entries of `osrf-pycommon`. Not sure what it is on about.

wjwwood:
Sure enough it was in the `requirement.txt` file, I fixed it with abb17d7

dirk-thomas:
Imo the "Migrating from catkin_make" pages suffers from the fact that it mixes two aspects. The step to move from `catkin_make` to an *isolated* build (like `catkin_make_isolated`) and then the step to `catkin build`. By better separating these two, their differences and similarities as well as their common problems I think a lot of previous (nd likely upcoming) questions could be avoided. Since basically users are not aware of the isolation aspect in the first place. I think a two-step migration would help users to first get their code working with `catkin_make_isolated` before reporting "issues" with `catkin build`.

Issue #319, opened 2016-04-14
out of place package.xml in the devel space


wjwwood:
After following the quick start guide (new documentation): http://catkin-tools.readthedocs.org/en/pre-0.4.0-docs-rebase/quick_start.html

You can see in the generated documentation, and if you reproduce it locally, that the devel space has a `package.xml` in the root, which seems wrong. Is that intended or a bug?

wjwwood:
It's not there if you build a workspace that contains `catkin`. It looks to be related to the "prebuild".

jbohren:
At some point we ended up moving the prebuild package source into the same target as it's private develspace, so those files were getting linked too (including CMakeLists.txt). This is now fixed in #321.

jbohren:
This is fixed with #321.

PR #321, opened 2016-04-14
metadata: always initialize metadata, avoid overwriting migration

jbohren:
This fixes an issue where a user can lose workspace configuration when migrating from 0.3.1 to 0.4.0.

Also fixes:
- #319

### DIFF
--- a/catkin_tools/metadata.py
+++ b/catkin_tools/metadata.py
@@ -78,7 +78,7 @@ def get_profiles_path(workspace_path):
 
 
 def get_paths(workspace_path, profile_name, verb=None):
-    """Get the path to a metadata directory and verb-specific metadata file.
+    """Construct the path to a metadata directory and verb-specific metadata file.
 
     Note: these paths are not guaranteed to exist. This function simply serves
     to standardize where these files should be located.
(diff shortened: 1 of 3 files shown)


catkin_tools/metadata.py
```diff
@@ -152,23 +174,28 @@ def migrate_metadata(workspace_path, old_version):
...
+                isolate_devel = metadata.get('isolate_devel')
+                if isolate_devel is not None:
                     del metadata['isolate_devel']
-                metadata['devel_layout'] = ('isolated' if isolated_devel else 'merged')
+                devel_layout = ('isolated' if isolate_devel else 'merged')
+                metadata['devel_layout'] = devel_layout
 
             # Save the new metadata
-            update_metadata(workspace_path, profile_name, verb, metadata, no_init=True)
+            new_metadata = update_metadata(workspace_path, profile_name, verb, metadata, no_init=True, merge=False)
```
wjwwood: `new_metadata` is unused, is that correct?

wjwwood: Also, I think this line is too long (which is failing the CI).

jbohren:
@NikolausDemmel This should fix the migration problem you saw.

jbohren:
@wjwwood This should be good to merge, I cleaned up the leftover stuff from debugging that was in the earlier commit.

PR #324, opened 2016-04-15
setup: Installation with `--user` fails with data files

jbohren:
This PR enables users to use the standard `--user` and `--home` options to install to `~/.local`. Otherwise, the following happens:

```
$ python setup.py install --user
running install
running build
running build_py
running install_lib
running install_data
copying completion/catkin_tools-completion.bash -> /etc/bash_completion.d
error: /etc/bash_completion.d/catkin_tools-completion.bash: Permission denied
```

This also suppresses misinformative help as described in #291.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -97,7 +97,7 @@ To enable tab completion, add the following to your '~/.bashrc':
 
 parser = argparse.ArgumentParser(add_help=False)
 prefix_group = parser.add_mutually_exclusive_group()
-prefix_group.add_argument('--user','--home', action='store_true')
+prefix_group.add_argument('--user', '--home', action='store_true')
 prefix_group.add_argument('--prefix', default=None)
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -94,11 +95,14 @@ def run(self):
                         'etc/bash_completion.d',
                         'catkin_tools-completion.bash')))
 
-parser = argparse.ArgumentParser()
-parser.add_argument('--prefix', default=None,
-                    help='prefix to install data files')
+parser = argparse.ArgumentParser(add_help=False)
+prefix_group = parser.add_mutually_exclusive_group()
+prefix_group.add_argument('--user','--home', action='store_true')
```
wjwwood: pep8: space after commas

wjwwood: ccbda1c

PR #328, opened 2016-04-16
docs: fix link to milestones

scpeters:
It currently doesn't render the link as clickable (see top of docs):

http://catkin-tools.readthedocs.org/en/latest/index.html

PR #329, opened 2016-04-16
Fix link on index

wjwwood:
I think this is a fix that's truer to the original intent, should replace: https://github.com/catkin/catkin_tools/pull/328


scpeters:
is there a preview of the docs from CI?

wjwwood:
@scpeters no, but we can generate them on demand.

scpeters:
as long as it makes the link clickable I approve

wjwwood:
I checked it locally.

PR #331, opened 2016-04-16
Replace dependant with dependent

### LINKED ISSUE #330: Spelling: dependant vs dependent
The spelling `dependant` (with an `a`) is used in several command-line parameters and places within the code (first noticed in the [catkin clean docs](http://catkin-tools.readthedocs.org/en/latest/verbs/catkin_clean.html)). This is the British spelling, which looks weird to me, but it is also not consistent, since the US spelling is used as well on that same page.

scpeters:
Closes #330.
~~~
sed -i -e 's@dependant@dependent@' `grep -rlI dependant .`
~~~

I didn't check this at all, I just ran the sed script.

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -298,8 +298,8 @@ def get_recursive_run_depends_in_workspace(packages, ordered_packages):
     )
 
 
-def get_recursive_build_dependants_in_workspace(package_name, ordered_packages):
-    """Calculates the recursive build dependants of a package which are also in
+def get_recursive_build_dependents_in_workspace(package_name, ordered_packages):
+    """Calculates the recursive build dependents of a package which are also in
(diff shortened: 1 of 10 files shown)


wjwwood:
I think those changes look reasonable. Thanks @scpeters!

@jbohren I'll leave this to you to merge, that way you have a chance to chime in before.

scpeters:
I missed one (b55fe8a) because I didn't append `g` to the search / replace `sed` command.

jbohren:
> I missed one (b55fe8a) because I didn't append `g` to the search / replace `sed` command.

Yeah and `catkin clean` could definitely use more unit tests (which would have caught this).

Thanks for the sedding, @scpeters .

scpeters:
no problem!

PR #340, opened 2016-04-19
clean: Fixing non-space cleaning warning

jbohren:
The additional warnings added recently actually prevent cleaning of individual packages. The `catkin clean` cli exits before actually cleaning.

@wjwwood This should go out for immediate release after merging.

Thanks to @wkentaro for finding this after 0.4.0 was released to the wild!


wkentaro:
Is this supporting prompt of asking for yes or no?

jbohren:
> Is this supporting prompt of asking for yes or no?

No, the user shouldn't be prompted for cleaning specific packages.

wkentaro:
I see. thanks.

PR #342, opened 2016-04-19
common: Fixing mishandling of environments with unicode values

### LINKED ISSUE #338: UnicodeEncodeError with non-ASCII character in environment
On my system, I have the environment variable `DEBFULLNAME=Timo Röhling` (note the non-ASCII umlaut in my last name) set, which causes `catkin build` to raise the following exception:

    [build] Found '275' packages in 0.0 seconds.                                   
    [build] Package table is up to date.                                           
    [build] An internal error occurred!                                            
      File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/build.py", line 549, in build_isolated_workspace
        continue_without_deps=False))
      File "/usr/lib/python2.7/dist-packages/catkin_tools/execution/executor.py", line 375, in run_until_complete
        return loop.run_until_complete(coroutine)
      File "/usr/lib/python2.7/dist-packages/trollius/base_events.py", line 273, in run_until_complete
        return future.result()
      File "/usr/lib/python2.7/dist-packages/trollius/futures.py", line 277, in result
        raise self._exception
    UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 484: ordinal not in range(128)

Once I `unset DEBFULLNAME`, everything works.

You can reproduce the bug by running `eval $(printf 'export NON_ASCII=\xc3\xb6')` at the bash prompt before invoking `catkin build`.

jbohren:
It looks like `shlex.split` behaves very differently on Python 2.7 and Python 3.4. This patch is an awkward solution that defines a compatibility wrapper at import time.There are so many places that unicode/string handling is different between Python2 and Python3, and it'd be great to nail them all down in a cleaner way.

Fixes #338.


wjwwood:
I've run into a similar issue with `shlex`. In fact I've "backported" it in the past by copy-pasting the version from Python 3.5 into my code base and using it conditionally if the version < 3.5.

This looks good to me.

PR #344, opened 2016-04-19
controllers: fixing long delay when using a slow status rate

### LINKED ISSUE #337: did not exist immediately  when smaller limit-status-rate
when we set very small number on the limit-status-rate, the catkin build did not return until the system shows next status info, for example;
```
catkin build --limit-status-rate 0.01
```
will wait for 100 sec, even if you compile a single empty package.

jbohren:
Fixes #337


wjwwood:
lgtm

k-okada:
thanks you for quick fix

PR #345, opened 2016-04-19
Add test for unicode in env

scpeters:
Regression test for issue #338. I committed it [before the fix](https://github.com/scpeters/catkin_tools/commits/confirm_unicode_failure) to [confirm the failure](https://travis-ci.org/scpeters/catkin_tools/builds/124279451).


wjwwood:
Thanks @scpeters!

PR #349, opened 2016-04-20
re-enable --locate-extra-shell-verbs, see #244

k-okada:


catkin_tools/commands/catkin.py
```diff
@@ -155,6 +155,9 @@ def catkin_main(sysargs):
         help='Forces catkin to output in color, even when the terminal does not appear to support it.')
     add('--no-color', action='store_true', default=False,
         help='Forces catkin to not use color in the output, regardless of the detect terminal type.')
+    add('--locate-extra-shell-verbs', action='store_true',
+        help='Returns the full path of extra shell verbs settings files')
+
```
wjwwood: This extra line is causing CI to fail.

jbohren:
This was changed to `catkin locate --shell-verbs` in order to avoid polluting the main catkin args. If we re add it, it should not show up in help.

jbohren:
See #352 for a PR which adds a deprecation warning.

Issue #347, opened 2016-04-20
builds broken due to KeyError in controllers.py


defunctzombie:
[//]: # (Thank you for reporting an issue for catkin_tools!)

Builds do not work due to a KeyError in controllers.py

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

### System Info

* Operating System: `Linux 3.19.0-56-generic #62~14.04.1-Ubuntu SMP Fri Mar 11 11:03:15 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`
* Python Version: 2.7.6
* Version of catkin_tools: 0.4.1
* ROS Distro: indigo (on ubuntu trusty)

### Build / Run Issue

Builds via `catkin build` do not work because of a logic issue exposed in https://github.com/catkin/catkin_tools/commit/69ecfce4a7e6f881bcf8ed64701374924fcc56a9.

Builds fail with the following error at random times:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/dist-packages/catkin_tools/execution/controllers.py", line 615, in run
    del active_stages[event.data['job_id']]
KeyError: 'launch_task'
```

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [ ] Works with `catkin_make`
* [ ] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [ ] I did  not `read this`

I did not try catkin_make

### Expected Behavior

The build to not fail with a threading issue.

[//]: # (Please describe what you expected or think should happen.)

### Actual Behavior

Fails at stated above.

[//]: # (Please describe what actually happens.)

[//]: # (Please put long CLI output into Gists: http://gist.github.com)
[...]

PR #351, opened 2016-04-20
Fixes to executor controller logic and timing of active status

jbohren:
This fixes a bug in the executor controller logic which was exposed by #344, which aimed to maintain the same logic, but change the delay times.

Thanks to @k-okada (#337) and @defunctzombie (#347) for helping expose and fix this bug!


wjwwood:
It seems to be working for me. Code changes makes sense too. Thanks @jbohren.

PR #352, opened 2016-04-20
shell-verbs: adding instruction to use new command for getting shell verbs

jbohren:
This supersedes #349.

@k-okada, sorry for the disruption.


wjwwood:
lgtm

PR #353, opened 2016-04-21
resultspace: Fixing environment cache checking

jbohren:
Currently, it could blow away environment if there are no env hooks in the resultspace (fixes #350)

@mikepurvis Give this branch a shot.

PR #355, opened 2016-04-22
testing: Adding tests (and ci configs) which fail on El Capitan (0.4.x)

jbohren:


wjwwood:
Needs a rebase now? (there seem to be commits from the one I just merged #353)

jbohren:
Yeah I'll rebase this. Also we can hold off on merging it until we figure out what the fix is.

jbohren:
@wjwwood So I added two commits which make `ctest` work with plain CMake packages on El Capitan with SIP.

First, a491706 simply enables `CMAKE_MACOSX_RPATH` by default. This fixes running `ctest` in a linked develspace.

Second, 9c8b3be adds a Catkin package to the merged develspace test, which is apparently necessary to set up the environment. This means that there's something that the Catkin setup files are doing that the vanilla CMake ones aren't.

wjwwood:
Just catching up on this issue, I do not think we can set `CMAKE_MACOSX_RPATH` by default in catkin tools because I think it is a choice that the developer and/or the packages being built should make. So we'll either have to decline this whole pr, or find a way to run these tests without SIP interfering. On my machines I just ended up disabling SIP, as it was just too painful to deal with.

mikepurvis:
Closing due to inactivity.

PR #359, opened 2016-04-22
testing: Adding tests (and ci configs) which fail on El Capitan (for 0.3.1)

jbohren:


jbohren:
@wjwwood Is it possible to trigger a build of this on the CI?

wjwwood:
Travis will not work if the branch cannot be merged (which is the case here).

You'll need to setup a `0.3` branch and make pull request against that.

Issue #354, opened 2016-04-22
DYLD_LIBRARY_PATH missing in 'make test'


mkoval:
### System Info

* Operating System: 15.2.0 Darwin Kernel Version 15.2.0: Fri Nov 13 19:56:56 PST 2015; root:xnu-3248.20.55~2/RELEASE_X86_64 x86_64 i386 MacBookPro10,1 Darwin
* Python Version: Python 2.7.11
* Version of catkin_tools: 0.4 and 6bc4b53590e4f1f2f4ed757c4f813e13f2adc370
* ROS Distro: indigo

### Build / Run Issue

I am building a non-Catkin `CMake` package with `catkin build` that uses `make test` to run tests through CMake's CTest integration. I typically run the tests like this:
```console
$ catkin build -i -p 1 --no-deps --make-args test -- my_package
```
This previously worked fine on OS X. Running the tests now fails because the dynamic linker cannot find the libraries build by `my_package`:
```
      Start 43: test_VanDerCorput
43/43 Test #43: test_VanDerCorput ...........................***Exception: Other  0.01 sec
dyld: Library not loaded: @rpath/libdart.6.0.dylib
  Referenced from: /Users/mkoval/development/test-ws/build/aikido/tests/util/test_VanDerCorput
  Reason: image not found
```

I am able to manually run the test (i.e. by invoking the test executable) after sourcing `setup.bash`. This works because `setup.bash` adds the necessary directory to `DYLD_LIBRARY_PATH`:
```
$ echo $DYLD_LIBRARY_PATH
/Users/mkoval/development/test-ws/devel/lib:/opt/ros/indigo/lib
```
I suspect something in `catkin build` is stripping `DYLD_LIBRARY_PATH` from the environment before running `make test`.

This is similar to #350, since that is an issue with `LD_LIBRARY_PATH` being cleared, but I am not sure if it is exactly the same: I am not having any problems with `catkin_tools_prebuild`. I also tried the proposed fix #353 to no avail.

jbohren:
> This is similar to #350, since that is an issue with `LD_LIBRARY_PATH` being cleared, but I am not sure if it is exactly the same: I am not having any problems with `catkin_tools_prebuild`. I also tried the proposed fix #353 to no avail.

[...]

PR #360, opened 2016-04-22
testing: Adding tests (and ci configs) which fail on El Capitan (0.3.x)

jbohren:


jbohren:
According to https://travis-ci.org/catkin/catkin_tools/builds/125156823 it looks like `build_env.sh` fails to handle paths with whitespace on el cap, as well.

jbohren:
@wjwwood So I don't think #354 is a regression, based on the same error on el cap: https://travis-ci.org/catkin/catkin_tools/builds/125169012

What do you think?

wjwwood:
I agree, I think this is SIP at work. I don't think it is a regression of the `0.4` version of `catkin_tools`.

wjwwood:
I'm going to close this.

Issue #361, opened 2016-04-28
Build errors not displayed


dornhege:
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

### System Info

* Operating System: Trusty
* Python Version: 2.7.6
* Version of catkin_tools: 0.4.2
* ROS Distro: indigo

### Build / Run Issue

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

Whenever there is a compiler error (g++) with the current version of catkin tools it is not displayed, i.e., I get:
```
Starting  >>> test_pkg                                                                                 
_______________________________________________________________________________________________________
Errors     << test_pkg:make ...rosws/test/logs/test_pkg/build.make.002.log                 
make[2]: *** [CMakeFiles/test_pkg_node.dir/src/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/test_pkg_node.dir/all] Error 2
make: *** [all] Error 2
cd ...rosws/test/build/test_pkg; catkin build --get-env test_pkg | catkin env -si  /usr/bin/make --jobserver-fds=6,7 -j; cd -
.......................................................................................................
```

The listed log file correctly contains the build errors.
```
rosws/test/src/test_pkg/src/main.cpp:5:2: error: #error Test
 #error Test
  ^
make[2]: *** [CMakeFiles/test_pkg_node.dir/src/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/test_pkg_node.dir/all] Error 2
make: *** [all] Error 2
```
`catkin build -v` also shows the errors directly together with all the other verbose-ness.

I believe that not showing errors or warnings is not the intended behavior and these should show in the build output.

jbohren:
@dornhege Can you give a little more info about your shell and terminal emulator?

This works for me, building the following package on 14.04 / 0.4.2:
https://github.com/jbohren/test_pkg

[...]

PR #362, opened 2016-04-28
build: Adding option to show full stdout if a stage fails

jbohren:


jbohren:
This fixes #361

wjwwood:
As I mentioned in the other issue, my personal opinion is that this should be the default behavior and there could be a option or setting that switches it to the current behavior. (This underscores the need for a workspace agnostic catkin tools settings, something like ~/.catkin_tools.conf)

mikepurvis:
Strongly in support of this. Any chance it can target 0.4.3?

wjwwood:
I was going to open a new pr that inverts the logic, so that printing stdout was the default behavior and you could configure catkin_tools through a user level config to do the current behavior. However, I haven't had time to do this. Hopefully I'll have a week to catch up on this repository and the others like `rviz`in the coming month.

Levi-Armstrong:
Would this address issue #419?

wjwwood:
@Levi-Armstrong I believe so.

rhaschke:
> Would this address issue #419?
@wjwwood I don't think so. #419 will be fixed with #400. This PR collects all output (stdout + stderr) from the process and prints it on stdout, doesn't it?

wjwwood:
@rhaschke I think you may be right. I am getting the "how much to print out" versus "where to print what out" confused.

PR #365, opened 2016-04-29
completion: Adding new zsh-specific completion script, improvements to bash completion

jbohren:
It also includes @rhaschke's contributions from #209.


wjwwood:
lgtm

Issue #311, opened 2016-04-05
UnicodeDecodeError decode error during linking


NikolausDemmel:
I just got this error during linking the devel products (this is on `pre-0.4.0-clean-pkgs`):

```
Errors     << zeno_tracking_lib:symlink /home/den2pal/work/ros_ws/logs/zeno_tracking_lib/build.symlink.003.log                                                                                                                             
Stage `symlink` failed with arguments:
[...]

PR #368, opened 2016-05-04
io: Better tracebacks, fixing unicode error

jbohren:
This reproduces the problem seen in #311 and improves traceback reporting from `FunctionStage` job stages. It also adds a fix for #311, though it's a quick fix and feels sub-optimal. There should be a better way of handling this for Python2 and Python3 without incurring so much additional overhead.


jbohren:
@wjwwood Is this a reasonable thing to do until we make an optimization pass over the whole system?

PR #373, opened 2016-05-14
fix bash completion

rhaschke:
Merge of #365 broke bash completion. Compared to my original PR #209, #365 leaves some old, but obsolete code. This PR removes it at thus fixes bash completion.


jbohren:
Oh wow yeah that's really broken.

PR #374, opened 2016-05-18
Update installing.rst

ahertle:
Removed the '$' at the beginning of each command for more convenient copy&paste.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -16,15 +16,15 @@ First you must have the ROS repositories which contain the ``.deb`` for ``catkin
 
 .. code-block:: bash
 
-    $ sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu `lsb_release -sc` main" > /etc/apt/sources.list.d/ros-latest.list'
-    $ wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
+    sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu `lsb_release -sc` main" > /etc/apt/sources.list.d/ros-latest.list'
+    wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
 
 Once you have added that repository, run these commands to install ``catkin_tools``:
 
 .. code-block:: bash
 
-    $ sudo apt-get update
-    $ sudo apt-get install python-catkin-tools
+    sudo apt-get update
+    sudo apt-get install python-catkin-tools
 
 Installing on other platforms with pip
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -33,7 +33,7 @@ Simply install it with ``pip``:
 
 .. code-block:: bash
 
-    $ sudo pip install -U catkin_tools
+    sudo pip install -U catkin_tools
 
 Installing from source
 ^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


jbohren:
@wjwwood I'm fine with this change, do you have any issues with it?

wjwwood:
I'd prefer not to. In all of the documentation I interact with I try to use the `$` before commands. You might think it "nanny-state" of me, but I think it's better to make you look at what you're copying and pasting it into the terminal. It also differentiates the quote sections that are commands and the ones that are example output from commands. Also, you can use this to distinguish bash (`$`) from zsh (`%`) or even root (`#`) which is a common practice in Linux documentation.

Consider:

```
$ ls
CHANGELOG.rst         MANIFEST.in           build                 catkin_tools.egg-info docs                  setup.py              tests
LICENSE               README.md             catkin_tools          completion            requirements.txt      stdeb.cfg
```

Versus

```
ls
CHANGELOG.rst         MANIFEST.in           build                 catkin_tools.egg-info docs                  setup.py              tests
LICENSE               README.md             catkin_tools          completion            requirements.txt      stdeb.cfg
```

I realize that that is not exactly the same as a block with just a list of commands to run and no example output, but it can still be confusing to novice users or people who just copy any quoted text and paste it without reading what it is:

> Do this:
```
cd somewhere
touch something
cd -
```

versus some example output inline in documentation:

> You output should look something like this:
```
Running commands:
rm -rf ~/*
echo "maybe a bit contrived..."
```

Things are already inconsistent (there are many places where commands do not have `$` before them), but if we're heading towards a consistent state this would not be the right direction I would push.

I'd be willing to compromise with a widget that prefixed the `$` but had a button to hide them (sort of like similar widgets that show line numbers but they can be hidden for copy-paste).

jbohren:
> I'd be willing to compromise with a widget that prefixed the $ but had a button to hide them (sort of like similar widgets that show line numbers but they can be hidden for copy-paste).

@ahertle Could you look into a readthedocs.org - compatible way of doing that?

ahertle:
I could not find a widget to dynamically show/hide line numbers in the documentation. Apparently line numbers can only be enabled in the source. So I am giving up on the idea of toggling $ widgets...

My actual problem with the current version is the first code-block:
```
$ sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu `lsb_release -sc` main" > /etc/apt/sources.list.d/ros-latest.list'
$ wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
```

The first line is longer than the page width which makes reading/understanding that line harder.
Would a line break like this be acceptable?
```
$ sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu `lsb_release -sc` main" \
   > /etc/apt/sources.list.d/ros-latest.list'
$ wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
```

wjwwood:
A line break is fine by me, however, I'd break it like this probably:

```bash
$ sudo sh \
    -c 'echo "deb http://packages.ros.org/ros/ubuntu `lsb_release -sc` main" \
        > /etc/apt/sources.list.d/ros-latest.list'
$ wget http://packages.ros.org/ros.key -O - | sudo apt-key add -
```

That way it is clear that the line break between the echo and the pipe are actually inside the single quoted things. Is that ok with you? If so I'll take your commit, put one of my own on top to add the extra break/indention, and merge it.

ahertle:
Fine with me :)

wjwwood:
Great, done. Thanks!

PR #377, opened 2016-05-26
add support for symlinked directories in private devel space

dominiquehunziker:
### Build / Run Issue

Symlinks in the package devel space linking to directories are not properly forwarded in the shared linked devel space layout, i.e., an empty directory is created instead of a symlink.

_Use case:_ I symlink a data directory to the devel space [like this](https://github.com/dominiquehunziker/catkin_tools_test_ws/blob/master/src/py_dummy_pkg/CMakeLists.txt#L9-L15) to make use of package data for Python packages in the devel space [here](https://github.com/dominiquehunziker/catkin_tools_test_ws/blob/master/src/py_dummy_pkg/src/py_dummy_pkg/loader.py#L4).

Sample workspace: https://github.com/dominiquehunziker/catkin_tools_test_ws

```
. /opt/ros/indigo/setup.sh
catkin clean -y
catkin config --link-devel
catkin build
. devel/setup.bash
rosrun py_dummy_pkg runner.py
```
### Suggested fix

Basically, the code/logic for files is reused (maybe this should be refactored?). However, I'm uncertain if the check for collisions is sufficient and if they should be handled differently from the files.

### DIFF
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -243,14 +243,33 @@ def link_devel_products(
 
         # create directories in the destination develspace
         for dirname in dirs:
+            source_dir = os.path.join(source_path, dirname)
             dest_dir = os.path.join(dest_path, dirname)
 
-            if not os.path.exists(dest_dir):
-                # Create the dest directory if it doesn't exist
-                os.mkdir(dest_dir)
-            elif not os.path.isdir(dest_dir):
-                logger.err('Error: Cannot create directory: ' + dest_dir)
-                return -1
+            if os.path.islink(source_dir):
+                # Store the source/dest pair
+                products.append((source_dir, dest_dir))
+
+                if os.path.exists(dest_dir):
+                    if os.path.realpath(dest_dir) != os.path.realpath(source_dir):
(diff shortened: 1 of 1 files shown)


remod:
I have the same issue. An easy fix which works for me is to replace

`os.walk(source_devel_path)`

with

`os.walk(source_devel_path, followlinks=True)`

Instead of cloning the symlink this will symlink its content.

dominiquehunziker:
Yes, that works also.
However, the behavior is not exactly the same. In your case you would have to rerun catkin if you add a file to the symlinked directory. In my case this is transparent and, therefore, reflects the Python behavior in the devel space better.

NikolausDemmel:
This change looks good to me. Haven't tested though.

wjwwood:
I just tested this and I was able to reproduce the issue and see that this pr fixes the issue. The code changes are pretty involved, but they make sense to me. @jbohren if you ever have time, it would be good for you to review this change too. However, I'm going to merge it for now as it seems to work and not break anything I can see.

Thanks @dominiquehunziker and everyone else who chipped in on the discussion. Sorry it took so long to merge.

PR #388, opened 2016-07-15
Add option to clean pyc files

paulbovbel:
@mikepurvis @jliviero


paulbovbel:
Pre-emptive, disregard please

PR #389, opened 2016-07-19
Add option to clean pyc files

paulbovbel:
Stale .pyc files in the workspace have led to several strange issues seen during development.

Adding (default on) option to clean out .pyc files from the .src directory whenever the workspace is cleaned.

https://github.com/paulbovbel/catkin_tools/pull/1

### DIFF
--- a/catkin_tools/verbs/catkin_clean/cli.py
+++ b/catkin_tools/verbs/catkin_clean/cli.py
@@ -19,6 +19,7 @@ try:
 except NameError:
     raw_input = input
 
+import fnmatch
 import os
 import shutil
 import sys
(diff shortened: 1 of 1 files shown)


paulbovbel:
Ping, any interest in this?

wjwwood:
I'm hesitant about this because it is Python specific and because it deletes files in the source space (which is a cardinal sin in my book). It might be ok in this case, but I'm not completely convinced. The `pyc` files should only be an issue if you delete the equivalent `py` file but not the `pyc` file. I think you need to provide more evidence as to why you think it necessary to remove these files automatically. I think you might find that these are not the root of the problem.

mikepurvis:
A potential alternative is to [prevent them ever being created](https://docs.python.org/2/using/cmdline.html#envvar-PYTHONDONTWRITEBYTECODE).

paulbovbel:
Fair enough, I'm on board with keeping the src space inviolate

evenator:
Would it make sense to set `PYTHONDONTWRITEBYTECODE` in devel space setup scripts? Or is that too much spooky action at a distance?

PR #393, opened 2016-08-11
Update build.py

DLu:


wjwwood:
@DLu this needs to be rebased since https://github.com/catkin/catkin_tools/pull/395 was merged.

wjwwood:
Actually, never mind this change is trivial. Thanks for the fix!

PR #395, opened 2016-08-15
update flake8 test since api breakage

wjwwood:
This is causing travis to always fail.


mikepurvis:
LGTM, seems you'll have to move `brew install enchant` outside of the Py3 test block here:

https://github.com/catkin/catkin_tools/blob/master/.travis.before_install.bash

wjwwood:
Sounds right.

mikepurvis:
Ship it. 👍

Issue #396, opened 2016-08-17
warnings and errors from stderr should be output on stderr (instead of stdout)


rhaschke:
The compact output summary of catkin tools is really great.
However, using it with qtcreator (> 4), poses the issue, that warnings and errors are no longer on stderr, but they are redirected to stdout by catkin_tools. For this reason, qtcreator doesn't parse those warnings/errors anymore to allow for source-code hyperlinking.

Could you consider to output warnings/errors on stderr again?

dseifert:
This is not only an issue with QtCreator 4.x but also earlier versions. And I would expect also other IDEs. It was also shortly discussed in #157. A quick work-around is to simply pipe the entire `catkin build` output to stderr - the result is not pretty, but it usually works well enough.

rhaschke:
Ping.

wjwwood:
@rhaschke I was looking at these this week. I'm still trying to figure out how https://github.com/catkin/catkin_tools/pull/400 and https://github.com/catkin/catkin_tools/pull/362 and https://github.com/catkin/catkin_tools/pull/362#issuecomment-215560993 all fit together.

wjwwood:
Closed by #424.

PR #400, opened 2016-09-05
direct stderr output to stderr again

rhaschke:
Fix for #396.

### DIFF
--- a/catkin_tools/verbs/catkin_build/cli.py
+++ b/catkin_tools/verbs/catkin_build/cli.py
@@ -158,6 +158,8 @@ the --save-config argument. To see the current config, use the
         help='Print output from commands in ordered blocks once the command finishes.')
     add('--interleave-output', '-i', action='store_true', default=False,
         help='Prevents ordering of command output when multiple commands are running at the same time.')
+    add('--split-stderr', '-e', action='store_false', dest='merge_stderr_into_stdout', default=True,
+        help='Direct stderr output to stderr again, instead of merging it with stdout (which is the default)')
     add('--no-status', action='store_true', default=False,
         help='Suppresses status line, useful in situations where carriage return is not properly supported.')
     add('--summarize', '--summary', '-s', action='store_true', default=None,
@@ -410,6 +412,7 @@ def main(opts):
         force_color=opts.force_color,
         quiet=not opts.verbose,
         interleave_output=opts.interleave_output,
+        merge_stderr_into_stdout=opts.merge_stderr_into_stdout,
         no_status=opts.no_status,
         limit_status_rate=opts.limit_status_rate,
         lock_install=not opts.no_install_lock,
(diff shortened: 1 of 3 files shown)


catkin_tools/verbs/catkin_build/cli.py
```diff
@@ -158,6 +158,8 @@ def prepare_arguments(parser):
         help='Print output from commands in ordered blocks once the command finishes.')
     add('--interleave-output', '-i', action='store_true', default=False,
         help='Prevents ordering of command output when multiple commands are running at the same time.')
+    add('--split-stderr', '-e', action='store_false', dest='merge_stderr_into_stdout', default=True,
+        help='Direct stderr output to stderr again, instead of merging it with stdout (which is the default)')
```
NikolausDemmel: The help should maybe not contain the "again"?

NikolausDemmel:
I notice that when running `catkin build -v -j1` I get the output from stdout not continuously but all in one go once it is finished (for each build stage). I'm not sure if that is supposed to be that way or not, but in any case, how does that play together with this addition.

Would I get the full output of stderr and stdout interleaved correctly, or would I get them as two blocks one after another and not interleaved and I run with the added option to split stdout and stderr?

rhaschke:
> I notice that when running `catkin build -v -j1` I get the output from stdout not continuously.

I guess, this is intended. Otherwise output from different jobs might be interleaved. You can enforce interleaved output with option `--interleave-output`. I'm not sure, catkin automatically evaluates `-j1` to modify it's output mode.

> Would I get the full output of stderr and stdout interleaved correctly, or would I get them as two blocks one after another and not interleaved when I run with the added option to split stdout and stderr?

Didn't tried this, yet. In any case, the actual output you will see strongly depends on the buffering mode that is performed on stdout/stderr. By default stdout is buffered, while stderr is not, which will mix up the actual output!

wjwwood:
@NikolausDemmel You would need to use `-i` if you wanted the output from each build command to "stream" to the screen, otherwise it will just print in blocks as each step finishes.

NikolausDemmel:
Thanks for the hints. I wasn't actually aware of that flag.

wjwwood:
Oh and you don't need the `-i` if you use `-p1`. Basically, `-p1` ensures only one process would be printing at a time and therefore it is safe to print immediately without a prefix. The `-i` is designed for when you have multiple processes in flight and you need to prefix them with the package name (and stage I think) so that you can disambiguate what came from where when they interleave.

wjwwood:
Looking at this now, is there any benefit to outputting the error to stdout? It is already being captured separately, this only affects the printing, and so it shouldn't affect the output ordering. A calling program could always pipe the stderr of catkin build to stdout if they wished. I'm trying to come up with a reason that catkin build should need to do it itself.

@jbohren I imagine you're super busy at your new gig (congrats btw 😄), but perhaps you could give your perspective on why errors currently go to stdout in the first place?

I'm inclined to drop the option and just output errors to stderr always.

wjwwood:
I created https://github.com/catkin/catkin_tools/pull/424 to replace this one (could not push here), but I kept the commits from this pr. Thanks for pushing the initial pr @rhaschke.

PR #403, opened 2016-09-30
Setting build verb to generate .catkin file in develspace even if the…

jgrogers:
I have noticed that "linked" devel workspaces are broken if the user is also building an install workspace.  The .catkin file is not being generated in the linked devel workspace if the user is also building an install workspace.

### DIFF
--- a/catkin_tools/verbs/catkin_build/build.py
+++ b/catkin_tools/verbs/catkin_build/build.py
@@ -332,7 +332,7 @@ def build_isolated_workspace(
     # NOTE: This is done to avoid the Catkin CMake code from doing it,
     # which isn't parallel-safe. Catkin CMake only modifies this file if
     # it's package source path isn't found.
-    if not context.install:
+    if True: #not context.install:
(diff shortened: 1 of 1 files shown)


NikolausDemmel:
As mentioned in the issue, I don't think this is supposed to work and I suspect that there are many things that don't work even if you add this file in the devel space.

wjwwood:
I would not recommend using the devel space when you're installing the packages. In that case you should use the resulting install space.

If this is a use case people want to work, then I'll have to ask them to organize amongst themselves and test this solution out to make sure it works and then summarize what you tried out and why. With help testing it, or a summary of why it should be ok logically, I'll consider this for merge. It would also be nice to have a test for it.

Normally this is what I would do before merging the pr, but given my time constraints and other issues on this repository and the fact that it not a use case I recommend, I will defer this for now and see if anyone cares enough about it to help out with the review.

mikepurvis:
In my opinion, the organization of the develspace is pretty much completely irrelevant once the `--install` flag is set. We could almost just impose `--isolate-devel` implicitly when installing.

PR #414, opened 2016-12-09
Improve handling for SHELL environment variable:

apetrone:
* Use boolean conditional for checking shell_path to catch empty strings.

* Throw an error if shell_path does not exist.


wjwwood:
I tried to rebase this without closing it but something happened... I reopened as https://github.com/catkin/catkin_tools/pull/421.

PR #415, opened 2016-12-22
Use ioctl() to determine width of controlling terminal

roehling:
This PR gets rid of the `tput cols` subprocess, which is slow and pollutes log files with `tput: No value for $TERM and no -T specified` messages if no controlling terminal is present.

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -400,6 +400,19 @@ def log(*args, **kwargs):
             unicode_error_printed = True
 
 
+__warn_terminal_width_once_has_printed = False
+
+
+def __warn_terminal_width_once():
+    global __warn_terminal_width_once_has_printed
+    if __warn_terminal_width_once_has_printed:
+        return
+    __warn_terminal_width_once_has_printed = True
+    print('WARNING: Could not determine the width of the terminal. '
+          'This warning will only be printed once.',
+          file=sys.stderr)
+
+
 def terminal_width_windows():
     """Returns the estimated width of the terminal on Windows"""
     from ctypes import windll, create_string_buffer
@@ -409,6 +422,7 @@ def terminal_width_windows():
 
     # return default size if actual size can't be determined
     if not res:
+        __warn_terminal_width_once()
         return 80
 
     import struct
@@ -429,6 +443,7 @@ def terminal_width_linux():
             height, width = struct.unpack("hh", ioctl(f.fileno(), TIOCGWINSZ, "1234"))
     except (IOError, OSError, struct.error):
         # return default size if actual size can't be determined
+        __warn_terminal_width_once()
         return 80
     return width
 
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -420,9 +420,16 @@ def terminal_width_windows():
 
 def terminal_width_linux():
     """Returns the estimated width of the terminal on linux"""
-    width = subprocess.Popen('tput cols', shell=True, stdout=subprocess.PIPE, close_fds=False).stdout.readline()
-
-    return int(width)
+    from fcntl import ioctl
+    from termios import TIOCGWINSZ
+    import struct
+    try:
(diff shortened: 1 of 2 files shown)


catkin_tools/common.py
```diff
@@ -420,9 +420,16 @@ def terminal_width_windows():
...
-    return int(width)
+    from fcntl import ioctl
+    from termios import TIOCGWINSZ
+    import struct
+    try:
+        with open(os.ctermid(), "rb") as f:
+            height, width = struct.unpack("hh", ioctl(f.fileno(), TIOCGWINSZ, "1234"))
+    except (IOError, OSError, struct.error):
+        # return default size if actual size can't be determined
+        return 80
```
gavanderhoorn: Should an error be printed (once?) if the size cannot be determined? Might avoid users not understanding why `catkin_tools` output is limited to 80 chars in case the ioctl doesn't work for them.

roehling: I'd say the 80 column default should work well enough that no error message is necessary. It's not difficult to add though.

gavanderhoorn: Maybe an error was a bit strong, a warning would be better.

I'm not worried about it not working, I'm just thinking that it might avoid some confusion when users are somehow ending up with 80 char output on a 200+ col screen and they have no way of finding out why other than looking through the source code.

wjwwood:
I add 2e5695b, let me know what you guys think.

catkin_tools/common.py
```diff
@@ -399,6 +400,19 @@ def log(*args, **kwargs):
...
+
+
+def __warn_terminal_width_once():
+    global __warn_terminal_width_once_has_printed
+    if __warn_terminal_width_once_has_printed:
+        return
+    __warn_terminal_width_once_has_printed = True
+    print('WARNING: Could not determine the width of the terminal. '
+          'This warning will only be printed once.',
+          file=sys.stderr)
```
gavanderhoorn: Should this be: "WARNING: Could not determine the width of the terminal, will use default of 80 columns. This warning will only be printed once"?

wjwwood: That's a good idea, I added a note about the default of 80 in the message in 	7d586d4.

I also changed the `WARNING` to `NOTICE`, but that's just to check a hunch about the failing CI.

roehling: I prefer notice to warning anyway :-)
Speaking of noticing, I saw that only the OSX build is failing, maybe the terminal ioctl() does not work with that OS?

wjwwood:
Ok, that looks good to me. Thanks @roehling and @gavanderhoorn, if you guys have more comments, we can pick them up in a follow up pr.

catkin_tools/common.py
wjwwood: In my latest commit it is passing, I think it was picking up the `WARNING` in a check for whether or not there were warnings 😄. So changing it to `NOTICE` was an easy workaround.

gavanderhoorn:
Looks ok.

A comment: we should probably introduce a constant instead of using the magic nr `80` [here](https://github.com/catkin/catkin_tools/blob/c5e3a189608b6efd20e6229b2c5376323d051d86/catkin_tools/common.py#L427), [here](https://github.com/catkin/catkin_tools/blob/c5e3a189608b6efd20e6229b2c5376323d051d86/catkin_tools/common.py#L411-L414), [here](https://github.com/catkin/catkin_tools/blob/c5e3a189608b6efd20e6229b2c5376323d051d86/catkin_tools/common.py#L448) and [here](https://github.com/catkin/catkin_tools/blob/c5e3a189608b6efd20e6229b2c5376323d051d86/catkin_tools/common.py#L458) (all in `common.py`).

gavanderhoorn:
A question: is there a specific reason there are three places a default is returned? Is that just to keep `terminal_width_windows()` and `terminal_width_linux()` usable without having to through `terminal_width()`?

PR #421, opened 2017-01-06
Improve handling for SHELL environment variable:

wjwwood:
* Use boolean conditional for checking shell_path to catch empty strings.

* Throw an error if shell_path does not exist.

This is a rebased version of https://github.com/catkin/catkin_tools/pull/414

### DIFF
--- a/catkin_tools/resultspace.py
+++ b/catkin_tools/resultspace.py
@@ -97,15 +97,15 @@ def get_resultspace_environment(result_space_path, base_env=None, quiet=False, c
         )
 
     # Determine the shell to use to source the setup file
-    shell_path = os.environ.get('SHELL', None)
-    if shell_path is None:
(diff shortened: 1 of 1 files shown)


mingless:
It might be irrelevant, but in the dotfiles I am using SHELL is being set to 'bash' rather than '/bin/bash' which resulted in this commit breaking my catkin_tools. Since the fix is simple and I am not the original creator of the script I am not reporting any particular issue, but rather asking if it is good to assume that SHELL must be '/bin/bash' and cannot be 'bash'.

wjwwood:
I'm not sure if we can safely assume that the value in `SHELL` is absolute (i.e. we can run a subprocess with `shell=Fase`), but we currently do assume that. You can make an issue about this and someone can look into it as they have time.

PR #443, opened 2017-03-13
Correctly merge envvars from isolated workspaces.

### LINKED ISSUE #382: Isolated result spaces broken when not extending
### System Info
- Operating System: Ubuntu, OS X
- Python Version: `python --version`
- Version of catkin_tools: 0.4.3
- ROS Distro: Indigo
### Build / Run Issue

The following sequence results in an error:

```
$ rosinstall_generator --rosdistro indigo roscpp --deps --tar > roscpp.rosinstall
$ rosinstall src roscpp.rosinstall -j8
$ catkin config --isolate-install --isolate-devel --install  # also fails with --no-install
$ catkin build  # also fails with -p1
[ snip ]
Starting  >>> roscpp_serialization
___________________________________________________________________________________
Errors     << roscpp_serialization:cmake 
CMake Error at /Users/mikepurvis/isolate_install_experiments/devel/catkin/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
  Could not find a package configuration file provided by "roscpp_traits"
  with any of the following names:

    roscpp_traitsConfig.cmake
    roscpp_traits-config.cmake

  Add the installation prefix of "roscpp_traits" to CMAKE_PREFIX_PATH or set
  "roscpp_traits_DIR" to a directory containing one of the above files.  If
  "roscpp_traits" provides a separate development package or SDK, be sure it
  has been installed.
Call Stack (most recent call first):
  CMakeLists.txt:3 (find_package)
```

The following variants in the same workspace run to completion with a successful result:

```
src/catkin/bin/catkin_make_isolated
src/catkin/bin/catkin_make_isolated --install
src/catkin/bin/catkin_make_isolated --install --merge
some/other/workspace/env.sh catkin build
```

Not terribly clear what is going on here, but it seems that somehow the environment in the isolated build case isn't quite getting populated correctly.

mikepurvis:
Fixes #382. With this change, isolated devel and isolated install spaces both do the right thing:

```
rosinstall_generator --rosdistro indigo roscpp --deps --tar > roscpp.rosinstall
rosinstall src roscpp.rosinstall -j8
catkin config --isolate-install --isolate-devel --install  # or --no-install
catkin build
```

For packages with many dependencies, the `loadenv` stage can end up taking a nontrivial amount of time, since it's invoking as many `env.sh` instances as there are dependencies. However, this can be mitigated by using the `--env-cache` config option.

@wjwwood @jbohren

### DIFF
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -380,7 +380,7 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
     stages.append(FunctionStage(
         'loadenv',
         loadenv,
-        locked_resource='installspace',
+        locked_resource=None if context.isolate_install else 'installspace',
         job_env=job_env,
         package=package,
         context=context
@@ -504,7 +504,7 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
             [MAKE_EXEC, 'install'],
             cwd=build_space,
             logger_factory=CMakeMakeIOBufferProtocol.factory,
-            locked_resource='installspace'
+            locked_resource=None if context.isolate_install else 'installspace'
         ))
 
     return Job(
--- a/catkin_tools/jobs/utils.py
+++ b/catkin_tools/jobs/utils.py
@@ -57,6 +57,11 @@ def loadenv(logger, event_queue, job_env, package, context):
     if context.destdir is not None:
         job_env['_CATKIN_SETUP_DIR'] = context.package_dest_path(package)
 
+    updated_env = {}
+
+    # Get the envvars associated with each workspace. Order is unimportant here since
+    # there will either be only one env loader path (linked/merged result space) or
+    # there will be many, but dependencies don't have a fixed order.
     for env_loader_path in env_loader_paths:
         if logger:
             logger.out('Loading environment from: {}'.format(env_loader_path))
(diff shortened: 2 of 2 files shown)


catkin_tools/jobs/catkin.py
```diff
@@ -504,7 +504,7 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
             [MAKE_EXEC, 'install'],
             cwd=build_space,
             logger_factory=CMakeMakeIOBufferProtocol.factory,
-            locked_resource='installspace'
+            locked_resource=None if context.isolate_install else 'installspace'
```
mikepurvis: No more need to lock the install space if each package has its own one to write to.

mikepurvis:
Yes, will fix. I'm also wondering if it'd be worth trying to unit test that merging logic. Is there a way to do it which doesn't require moving that stuff off to a different function?

Are there any of the larger "workspace" tests which we can make cover the isolate-devel and isolate-install cases to guard against regression here?

catkin_tools/jobs/utils.py
```diff
@@ -66,8 +71,21 @@ def loadenv(logger, event_queue, job_env, package, context):
...
+            else:
+                updated_env[key] = set(values_str.split(':'))
+
+    # Now merge the updated environment into this job's environment, discarding duplicate
+    # values and ensuring that new ones are prepended.
+    for key, new_values_set in updated_env.items():
+        if key in job_env:
+            prepend_values = new_values_set - set(job_env[key].split(':'))
+            if prepend_values:
+                job_env[key] = ':'.join(prepend_values) + ':' + job_env[key]
```
mikepurvis: Another way could to drop the `if` and just do:

    job_env[key] = ':'.join(list(prepend_values) + job_env[key].split(':'))

Probably by keeping around the list since it's already split up a line above. The way it is seemed cleared to me though.

catkin_tools/jobs/catkin.py
mikepurvis: Possible this isn't necessary given that the underlying logic seems to swap out a `FakeLock`:

https://github.com/catkin/catkin_tools/blob/2442e1f971ae6bbd0c14fa6da183ee095ae099e3/catkin_tools/verbs/catkin_build/build.py#L535-L537

However, that's not turned on automatically when in isolated install mode, so maybe that's the better change to be making here?

mikepurvis:
Thanks for the quick merge! 👍

catkin_tools/jobs/utils.py
mikepurvis: Bug here, see #448.

PR #444, opened 2017-03-14
Add tests for isolated builds

mikepurvis:
These should fail until #443 is merged.


tests/system/verbs/catkin_build/test_build.py
```diff
@@ -50,7 +48,8 @@ def create_tree_workspace(wf, build_type, n_pkg_layers, n_children=2):
     for i in range(n_pkgs):
         wf.create_package(
             'pkg_{}'.format(i),
-            depends=(['pkg_{}'.format(floor(i - 1) / n_children)] if i > 0 else []))
+            build_type=build_type,
+            depends=(['pkg_{}'.format(int((i - 1) / n_children))] if i > 0 else []))
```
mikepurvis: This was previously ignoring the passed-in `build_type`, and the use of `floor` was resulting in dependencies like `pkg_2.5`.

tests/system/verbs/catkin_build/test_build.py
```diff
@@ -93,9 +92,35 @@ def test_build_dry_run():
...
+            n_pkgs = create_tree_workspace(wf, 'catkin', 2)
+            wf.create_package('pkg_dep', build_type='catkin',
+                              build_depends=['pkg_{}'.format(n) for n in range(n_pkgs)])
+            wf.build()
+
+            assert catkin_success(['config', '--isolate-devel'])
+            assert catkin_success(BUILD)
+            assert os.path.exists('devel/pkg_dep')
+            assert not os.path.exists('install')
+    assert_no_warnings(out)
```
mikepurvis: Could condense these into one, but I'm not sure it's worth it.

mikepurvis:
Good to go in now. Can confirm that this passes locally when rebased on top of the changes from #443.

tests/system/verbs/catkin_build/test_build.py
```diff
@@ -93,9 +92,35 @@ def test_build_dry_run():
...
 
 
-def test_build_all_isolated():
-    """Test building all packages in an isolated workspace"""
-    pass  # TODO: Implement test
+def test_build_all_isolate_install():
+    """Test building dependent catkin packages with isolated installspace."""
+    with redirected_stdio() as (out, err):
+        with workspace_factory() as wf:
+            n_pkgs = create_tree_workspace(wf, 'catkin', 2)
```
mikepurvis: It's not possible to test CMake packages in this way, since the simplified factory-generated ones don't include CMake modules which make them findable by dependencies.

tests/system/workspace_factory.py
```diff
@@ -51,6 +51,21 @@ class Package(object):
     <build_type>{build_type}</build_type>
   </export>"""
 
+        CATKIN_CMAKELISTS_TEMPLATE = """
+cmake_minimum_required(VERSION 2.8.3)
+project({name})
+find_package(catkin REQUIRED COMPONENTS {catkin_components})
+catkin_package()"""
```
mikepurvis: There's a simplifying assumption here that a factory catkin package only depends on other catkin packages. All current tests pass with this assumption in place, so I see no reason to over complicate things.

tests/system/verbs/catkin_build/test_build.py
wjwwood: That's true, unless you wanted to test something about a plain cmake package being a leaf. Also, despite that drawback, this function could be useful to get started and the test could add a CMake Module for the plain cmake package itself.

tests/system/verbs/catkin_build/test_build.py
wjwwood: Seems fine to me.

tests/system/workspace_factory.py
wjwwood: That's true, but at some point tests may just have to hand craft a `CMakeLists.txt` and not rely on a more and more complex function to generate everything for them.

wjwwood:
I restarted travis, but not sure if this has to be rebased instead.

wjwwood:
The Linux jobs look ok, just waiting on Travis' macOS jobs to catch up.

PR #449, opened 2017-03-25
Fix logic which merges environment PATH variables.

### LINKED ISSUE #448: Non-path envvars with colons get mangled
This is a regression introduced by the changes in #443. Specifically, @tspicer01 and I are seeing issues with `$ROS_MASTER_URI`, because the logic splits on `:` and can end up reordering the pieces within a given workspace, so we get crazy stuff in our environments like:

```
ROS_MASTER_URI=http:11311://localhost
```

I'll have a fix and some proper tests up shortly.

mikepurvis:
Compared to the previous implementation, this:

- Preserves order of both old and new path values.
- Is only applied to envvars which end in `PATH`.
- Has tests which clarify exactly what its behaviour is.

Fixes #448.

### DIFF
--- a/catkin_tools/jobs/utils.py
+++ b/catkin_tools/jobs/utils.py
@@ -67,12 +67,12 @@ def merge_envs(job_env, overlay_envs):
                     # here so that we can cheaply append to it, representing a prepend in the final
                     # PATH var, and because we need to maintain the order of underlay paths.
                     if key in job_env:
-                        values = job_env[key].split(':')
+                        values = job_env[key].split(os.pathsep)
                         values.reverse()
                         merge_path_values[key] = values
                     else:
                         merge_path_values[key] = []
-                merge_path_values[key].extend(values_str.split(':'))
+                merge_path_values[key].extend(values_str.split(os.pathsep))
             else:
                 # For non-PATH keys, simply assign the value. This may not always
                 # be correct behaviour, but we don't have the information here to
@@ -87,9 +87,7 @@ def merge_envs(job_env, overlay_envs):
             if value not in seen_values:
                 seen_values.add(value)
                 new_values_list.append(value)
-
-        new_values_list.reverse()
-        job_env[key] = ':'.join(new_values_list)
+        job_env[key] = os.pathsep.join(reversed(new_values_list))
 
 
 def loadenv(logger, event_queue, job_env, package, context):
@@ -110,7 +108,13 @@ def loadenv(logger, event_queue, job_env, package, context):
             cached=context.use_env_cache,
             strict=False))
 
-    merge_envs(job_env, envs)
+    # Avoid using merge logic if not required (in the non-isolated resultspace
+    # case. It has corner cases which may trip up the unwary, so having the
+    # option to switch to a merged resultspace is a good fallback.
+    if len(envs) > 1:
+        merge_envs(job_env, envs)
+    elif len(envs) == 1:
+        job_env.update(envs[0])
     return 0
 
 
--- a/catkin_tools/jobs/utils.py
+++ b/catkin_tools/jobs/utils.py
@@ -50,6 +50,48 @@ def get_env_loaders(package, context):
     return sources
 
 
+def merge_envs(job_env, overlay_envs):
+    '''
+    In the merged/linked case of single env, this function amounts to a straight
+    assignment, but a more complex merge is required with isolated result spaces,
+    since a package's build environment may require extending that of multiple
+    other result spaces.
+    '''
+    merge_path_values = {}
+
+    for overlay_env in overlay_envs:
+        for key, values_str in overlay_env.items():
+            if key.endswith('PATH'):
+                if key not in merge_path_values:
+                    # Seed the list with any values already in the environment. We reverse the list
+                    # here so that we can cheaply append to it, representing a prepend in the final
+                    # PATH var, and because we need to maintain the order of underlay paths.
+                    if key in job_env:
+                        values = job_env[key].split(':')
+                        values.reverse()
+                        merge_path_values[key] = values
+                    else:
+                        merge_path_values[key] = []
+                merge_path_values[key].extend(values_str.split(':'))
+            else:
+                # For non-PATH keys, simply assign the value. This may not always
+                # be correct behaviour, but we don't have the information here to
+                # know how to do anything else.
+                job_env[key] = values_str
+
+    # For the path values, do a deduplicating merge.
+    for key, values_list in merge_path_values.items():
+        seen_values = set()
+        new_values_list = []
+        for value in values_list:
+            if value not in seen_values:
+                seen_values.add(value)
+                new_values_list.append(value)
+
+        new_values_list.reverse()
+        job_env[key] = ':'.join(new_values_list)
+
+
 def loadenv(logger, event_queue, job_env, package, context):
     # Get the paths to the env loaders
     env_loader_paths = get_env_loaders(package, context)
@@ -57,35 +99,18 @@ def loadenv(logger, event_queue, job_env, package, context):
     if context.destdir is not None:
         job_env['_CATKIN_SETUP_DIR'] = context.package_dest_path(package)
 
-    updated_env = {}
-
-    # Get the envvars associated with each workspace. Order is unimportant here since
-    # there will either be only one env loader path (linked/merged result space) or
-    # there will be many, but dependencies don't have a fixed order.
+    envs = []
     for env_loader_path in env_loader_paths:
         if logger:
             logger.out('Loading environment from: {}'.format(env_loader_path))
-        resultspace_env = get_resultspace_environment(
+        envs.append(get_resultspace_environment(
             os.path.split(env_loader_path)[0],
             base_env=job_env,
             quiet=True,
             cached=context.use_env_cache,
-            strict=False)
-        for key, values_str in resultspace_env.items():
-            if key in updated_env:
-                updated_env[key] |= set(values_str.split(':'))
-            else:
-                updated_env[key] = set(values_str.split(':'))
+            strict=False))
 
-    # Now merge the updated environment into this job's environment, discarding duplicate
-    # values and ensuring that new ones are prepended.
-    for key, new_values_set in updated_env.items():
-        if key in job_env:
-            prepend_values = new_values_set - set(job_env[key].split(':'))
-            if prepend_values:
-                job_env[key] = ':'.join(prepend_values) + ':' + job_env[key]
-        else:
-            job_env[key] = ':'.join(new_values_set)
+    merge_envs(job_env, envs)
     return 0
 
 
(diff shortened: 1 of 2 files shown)


mikepurvis:
@wjwwood Thoughts on this?

tests/unit/test_jobs.py
```diff
@@ -0,0 +1,33 @@
...
+def test_merge_envs_basic():
+    job_env = { 'PATH': '/usr/local/bin:/usr/bin', 'FOO': 'foo' }
+
+    merge_envs(job_env, [
+        { 'PATH': '/usr/local/bin:/bar/baz/bin' },
+        { 'BAR': 'bar' } ])
+
+    # Validate that the known path was not moved from the existing order, and the unfamiliar
+    # path was correctly prepended.
+    assert job_env['PATH'] == '/bar/baz/bin:/usr/local/bin:/usr/bin'
```
wjwwood: This doesn't seem like what I'd expect from the deduplication... Isn't `/usr/local/bin:/bar/baz/bin:/usr/bin` more correct?

catkin_tools/jobs/utils.py
```diff
@@ -50,42 +50,67 @@ def get_env_loaders(package, context):
...
+    In the merged/linked case of single env, this function amounts to a straight
+    assignment, but a more complex merge is required with isolated result spaces,
+    since a package's build environment may require extending that of multiple
+    other result spaces.
+    '''
+    merge_path_values = {}
+
+    for overlay_env in overlay_envs:
+        for key, values_str in overlay_env.items():
+            if key.endswith('PATH'):
```
wjwwood: This is sooooo fragile. Imagine something like `MY_CONFIG_PATH=file:///path/to/config.conf` and that this is set by someone else as `MY_CONFIG_PATH=file:///path/to/override_config.conf`. This function will not handle that correctly. Obviously this is a made up example which is unlikely to happen, but if this function is being applied in all configurations of the tools, then if it does happen it will be broken and the user cannot even work around it.

So, is this function only being used in `--env-cache` or is it always being called, regardless of the tool configuration?

tests/unit/test_jobs.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # Confirm that a key only in the original env persists.
+    assert job_env['FOO'] == 'foo'
```
wjwwood: It would be good to see a non-PATH env var that contains a `:` as a sanity check, especially since `ROS_MASTER_URI` tripped it up in the beginning if I remember correctly.

catkin_tools/jobs/utils.py
```diff
@@ -50,42 +50,67 @@ def get_env_loaders(package, context):
...
+                if key not in merge_path_values:
+                    # Seed the list with any values already in the environment. We reverse the list
+                    # here so that we can cheaply append to it, representing a prepend in the final
+                    # PATH var, and because we need to maintain the order of underlay paths.
+                    if key in job_env:
+                        values = job_env[key].split(':')
```
wjwwood: I have no delusions that this will work on Windows, but there is a placeholder for this in Python that is portable:

https://docs.python.org/2.7/library/os.html#os.pathsep

catkin_tools/jobs/utils.py
mikepurvis: This function is used with and without `--env-cache`— it is required regardless because of the situation described in https://github.com/catkin/catkin_tools/issues/382#issuecomment-285739921. In short, each new `setup.bash` blows away the environment that was present before it, so this function is required any time there are multiple isolated result spaces which must be merged together to form the environment for a package's build (the only other way I can think of to deal with this would be to somehow mutate the various setup files to "chain up" the resultspaces, as if they had been built one after the other like with `catkin_make_isolated`).

So, we could turn this off for non-isolated-resultspace builds. We could also set an explicit whitelist of envvars for it to act on (`ROS_PACKAGE_PATH`, `CMAKE_PREFIX_PATH`, `CPATH`, `PYTHONPATH`, `PKG_CONFIG_PATH`, `PATH`). Or both. Let me know your preference.

catkin_tools/jobs/utils.py
mikepurvis: Cool, will do.

tests/unit/test_jobs.py
mikepurvis: The logic is that it's introducing `/usr/local/bin` and `/bar/baz/bin` as new paths. It deduplicates first, and maintains order of new paths second, so `/usr/local/bin` is tossed as part of deduplication, and `/bar/baz/bin` is prefixed onto the existing paths.

What would be the logic that would lead to '/bar/baz/bin:/usr/local/bin:/usr/bin'?

tests/unit/test_jobs.py
mikepurvis: Sure, yeah; I'll modify the test.

catkin_tools/jobs/utils.py
wjwwood: Ah, that's right thanks for the reminder and https://github.com/catkin/catkin_tools/issues/382#issuecomment-285739921.

I guess this isn't as bad since it really only applies to isolated devel or install. That at least gives the user a way out, i.e. use a merged option.

I'm still worried about the visibility of this issue, but I don't see an effective way to raise this concern with the user. We could compare each matching env variable with a whitelist and let them know with a notice that this might not be how you expect this variable to be handled. But again, I don't see how to aggregate these and give them to the user in a place that they'll likely see, like at the end of the program.

So I guess what I'm saying is, leave it unless you can think of a way to try and notify the users when something unexpected might be happening.

tests/unit/test_jobs.py
wjwwood: I understand your logic, but what I'm talking about is expectation.

If I put myself in the shoes of the person setting `/usr/local/bin:/bar/baz/bin`, I might expect that the merged path might contain other stuff in between my entries, e.g. `X:/usr/local/bin:Y:/bar/baz/bin:Z`, but I would not expect my entries to end up reordered, where the first occurrence of `/bar/baz/bin` comes up before the first occurrence of `/usr/local/bin`.

I think that "append only if not in list" makes sense, and I think "remove duplicates in list and then prepend" makes sense. But what you've got here is "prepend only if not in list" which doesn't make sense to me because it can cause the order of contributed entries to switch.

catkin_tools/jobs/utils.py
mikepurvis: In the implementation as currently proposed, the function _is_ also getting invoked in the non-isolated case. It's not really doing anything, but it's there ready to trip up a user in the scenario you describe— even without having a colon in the variable, you could get behaviour like:

1. Package installed in the underlay sets `FOO_PATH=bar`
2. Package in the current merged/linked resultspace overrides that to `FOO_PATH=baz`
3. Environment setup for next package build merges those into `FOO_PATH=baz:bar`, when the intended behaviour was a straight up override, not a prepend.

Hence the suggestion of either a whitelist of keys where prepending is _known_ to be correct behaviour, or else disabling this function altogether for the non-isolated case.

wjwwood: Oh, why is it necessary in the non-isolated case (just when I thought I was getting a handle on this thing)?

If it is "always on", then yeah I think a whitelist is the best alternative. I'd start with:

- `PATH`
- `CPATH`
- `LD_LIBRARY_PATH`
- `DYLD_LIBRARY_PATH`
- `DYLD_FALLBACK_LIBRARY_PATH`
- `CMAKE_PREFIX_PATH`
- `PKG_CONFIG_PATH`
- `PYTHONPATH`

And any others you can think of.

That being said, I don't think this can be a long term solution. I'm more convinced than ever that this mechanism is just fundamentally broken. I think we'll have to go back to `setup.bash --extend` like mechanics, despite the performance hit. At least I think it should be an option to operate completely with shell scripts, no env parsing/caching/manipulation by `catkin_tools`.

tests/unit/test_jobs.py
wjwwood: The most naive, but correct in my opinion, is to just prepend with no deduplication.

catkin_tools/jobs/utils.py
mikepurvis: It isn't necessary in the non-isolated case; it was only made always on because I was thinking it would be non-harmful, and nice to keep a lesser-used piece of code in service to prevent it from rotting. I'm adding a commit which avoids the merge when there's only a single new env.

tests/unit/test_jobs.py
mikepurvis: That doesn't work though, since each dependency will independently include its own copy of the underlay chain. So if you imagine an underlay U containing packages A,B,C, and then a workspace being built which contains updated versions of the three packages, with A' and B' being the isolated resultspaces that have been built. When it comes time to build C (which depends on A and B), what you want in your env path vars is A':B':U or B':A':U, and what you definitely do _not_ want is A':U:B':U or B':U:A':U, since then you'll have one or the other of the updated packages shadowed behind the version in the underlay.

In the real world, I don't think many/any packages are contributing more than one entry to any of the X_PATH vars, and even for those that do, relative order is maintained _within entries that don't overlap with something already present_. Given that, I think what is here is a reasonable compromise in simplicity without getting into an actual constraint satisfaction solver or whatever— particularly if it's constrained to the isolated case, which is a bit off the beaten track no matter how you slice things.

catkin_tools/jobs/utils.py
```diff
@@ -50,42 +50,71 @@ def get_env_loaders(package, context):
...
-            job_env[key] = ':'.join(new_values_set)
+            strict=False))
+
+    # Avoid using merge logic if not required (in the non-isolated resultspace
+    # case. It has corner cases which may trip up the unwary, so having the
+    # option to switch to a merged resultspace is a good fallback.
+    if len(envs) > 1:
+        merge_envs(job_env, envs)
+    else:
+        job_env.update(envs[0])
```
mikepurvis: Here's the logic which now avoids the merge function if not required by the presence of multiple resultspaces to extend.

tests/unit/test_jobs.py
mikepurvis: Have added this to the "nonpaths" test down below.

tests/unit/test_jobs.py
wjwwood: I see what you mean now. That's unresolvable I think as-is because each package brings a set of underlays which cannot be deduplicated at the "underlay" level. This might be something that can be addressed by a feature in ament (hopefully a back portable thing). Each package has two setup files, `setup.*sh` and `localsetup.*sh`. The latter only modifies the environment for that package. That means you can start with an underlay and know what each package brings to the table separately from their dependencies.

Anyways if you can remove this code path from the non-isolated case then this will have to be good enough I think.

catkin_tools/jobs/utils.py
wjwwood: +1

mikepurvis:
Likewise!

tests/unit/test_jobs.py
mikepurvis: The general merging case is much harder that this one— here we benefit from a bunch of assumptions, chiefly that all the envs being merged share a common underlay, and that since there are expected to be no duplicate files across the set of workspace resultspaces, the order they end up in in the overlay doesn't actually matter, as long as they're all present in the final list of paths, and all ahead of the underlay paths.

PR #452, opened 2017-04-13
Specify return code when build interrupted.

mikepurvis:
The main alternative here might be to return `errno.EINTR`, which evaluates to 4 on at least Linux and OS X. However, interrupting other common command line utilities on both platforms seems to always yield a 130:

```
$ find /
.
.
.
^C
$ echo $?
130
```

Same with `cat`:

```
$ cat
^C
$ echo $?
130
```

### DIFF
--- a/catkin_tools/verbs/catkin_build/build.py
+++ b/catkin_tools/verbs/catkin_build/build.py
@@ -576,6 +576,8 @@ def build_isolated_workspace(
     except KeyboardInterrupt:
         wide_log("[build] Interrupted by user!")
         event_queue.put(None)
+        
+        return 130  # EOWNERDEAD return code is not part of the errno module.
 
 
 def _create_unmerged_devel_setup(context, unbuilt):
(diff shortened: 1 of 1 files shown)


catkin_tools/verbs/catkin_build/build.py
```diff
@@ -576,6 +576,8 @@ def build_isolated_workspace(
     except KeyboardInterrupt:
         wide_log("[build] Interrupted by user!")
         event_queue.put(None)
+
```
wjwwood: Trailing whitespace here I think.

wjwwood:
Sorry, didn't notice you added a description after the fact.

wjwwood:
Why not just `1`? Shouldn't any non-zero code work?

mikepurvis:
Could be `1`, totally. I was just curious if there was a convention among other tools about what to do, and it seems there is, so I followed it.

wjwwood:
`130` is fine by me.

mikepurvis:
:+1:

PR #455, opened 2017-04-28
Better message when missing a required command line tool.

### LINKED ISSUE #437: Unexplicit error when cmake is not installed
I just reinstalled a computer and tried to build one of my project based on catkin (without using ROS) and I faced an unexplicit error (at least for neophytes like me)

```
  File "/usr/local/bin/catkin", line 11, in <module>
    load_entry_point('catkin-tools==0.4.4', 'console_scripts', 'catkin')()
  File "/usr/local/lib/python2.7/dist-packages/catkin_tools/commands/catkin.py", line 267, in main
    catkin_main(sysargs)
  File "/usr/local/lib/python2.7/dist-packages/catkin_tools/commands/catkin.py", line 262, in catkin_main
  File "/usr/local/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/cli.py", line 420, in main
    summarize_build=opts.summarize  # Can be True, False, or None
  File "/usr/local/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/build.py", line 441, in build_isolated_workspace
    prebuild=True)
  File "/usr/local/lib/python2.7/dist-packages/catkin_tools/jobs/catkin.py", line 447, in create_catkin_build_job
    occupy_job=True
  File "/usr/local/lib/python2.7/dist-packages/catkin_tools/execution/stages.py", line 91, in __init__
    raise ValueError('Command stage must be a list of strings: {}'.format(cmd))
ValueError: Command stage must be a list of strings: [None, '/home/medrimonia/rhoban/workspace/src/catkin', '--no-warn-unused-cli', '-DCATKIN_DEVEL_PREFIX=/home/medrimonia/rhoban/workspace/devel_release/.private/catkin', '-DCMAKE_INSTALL_PREFIX=/home/medrimonia/rhoban/workspace/install_release', '-DCMAKE_BUILD_TYPE=Release']
```
I have been able to track down the source of the error which was the value of `CMAKE_EXEC` which was `None`. Although I agree that it is obvious that cmake has to be installed, I think that error could be more explicit if a specific error was thrown by checking the value of `CMAKE_EXEC` at line 439 of `catkin.py`.

mikepurvis:
Looks like this:

```
$ catkin build
----------------------------------------------------------------
Extending:          [cached] /Users/mikepurvis/config_ws/install
Workspace:                   /Users/mikepurvis/config_ws
----------------------------------------------------------------
Source Space:       [exists] /Users/mikepurvis/config_ws/src
Log Space:          [exists] /Users/mikepurvis/config_ws/logs
Build Space:        [exists] /Users/mikepurvis/config_ws/build
Devel Space:        [exists] /Users/mikepurvis/config_ws/devel
Install Space:      [exists] /Users/mikepurvis/config_ws/install
DESTDIR:            [unused] None
----------------------------------------------------------------
Devel Space Layout:          linked
Install Space Layout:        merged
----------------------------------------------------------------
Additional CMake Args:       None
Cache Job Environments:      False
----------------------------------------------------------------
Whitelisted Packages:        None
Blacklisted Packages:        None
----------------------------------------------------------------
Workspace configuration appears valid.
----------------------------------------------------------------
[build] Found '6' packages in 0.0 seconds.                                                                                
[build] Error: Cannot find required tool `make` on the PATH, is it installed?
```

Fixes #437.

### DIFF
--- a/catkin_tools/jobs/utils.py
+++ b/catkin_tools/jobs/utils.py
@@ -25,6 +25,19 @@ from catkin_tools.resultspace import get_resultspace_environment
 from catkin_tools.execution.events import ExecutionEvent
 
 
+class CommandMissing(Exception):
+    '''A required command is missing.'''
+
+    def __init__(self, name):
+        super(CommandMissing, self).__init__(
+                'Cannot find required tool `%s` on the PATH, is it installed?' % name)
+
+
+def require_command(name, which):
+    if not which:
+        raise CommandMissing(name)
+
+
 def get_env_loaders(package, context):
     """Get a list of env loaders required to build this package."""
 
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -37,6 +37,7 @@ from .commands.make import MAKE_EXEC
 from .utils import copyfiles
 from .utils import loadenv
 from .utils import makedirs
+from .utils import require_command
 from .utils import rmfiles
 
 
@@ -432,6 +433,8 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
             prefix=context.package_dest_path(package)
         ))
 
+        require_command('cmake', CMAKE_EXEC)
+
         # CMake command
         stages.append(CommandStage(
             'cmake',
@@ -472,6 +475,8 @@ def create_catkin_build_job(context, package, package_path, dependencies, force_
             [MAKE_EXEC, 'clean'] + make_args,
             cwd=build_space,
         ))
+
+    require_command('make', MAKE_EXEC)
 
     # Make command
     stages.append(CommandStage(
--- a/catkin_tools/jobs/cmake.py
+++ b/catkin_tools/jobs/cmake.py
@@ -33,6 +33,7 @@ from .commands.make import MAKE_EXEC
 from .utils import copyfiles
 from .utils import loadenv
 from .utils import makedirs
+from .utils import require_command
 from .utils import rmfiles
 
 from catkin_tools.execution.jobs import Job
@@ -260,6 +261,8 @@ def create_cmake_build_job(context, package, package_path, dependencies, force_c
         dest_path=os.path.join(metadata_path, 'package.xml')
     ))
 
+    require_command('cmake', CMAKE_EXEC)
+
     # CMake command
     makefile_path = os.path.join(build_space, 'Makefile')
     if not os.path.isfile(makefile_path) or force_cmake:
@@ -290,6 +293,8 @@ def create_cmake_build_job(context, package, package_path, dependencies, force_c
             [MAKE_EXEC, 'clean'] + make_args,
             cwd=build_space,
         ))
+
+    require_command('make', MAKE_EXEC)
 
     # Make command
     stages.append(CommandStage(
--- a/catkin_tools/verbs/catkin_build/cli.py
+++ b/catkin_tools/verbs/catkin_build/cli.py
@@ -48,6 +48,7 @@ from catkin_tools.context import Context
 
 import catkin_tools.execution.job_server as job_server
 
+from catkin_tools.jobs.utils import CommandMissing
 from catkin_tools.jobs.utils import loadenv
 
 from catkin_tools.metadata import find_enclosing_workspace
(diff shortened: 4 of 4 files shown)


catkin_tools/verbs/catkin_build/build.py
```diff
@@ -461,49 +462,53 @@ def build_isolated_workspace(
...
+                    "[build] @!@{yf}Warning:@| Skipping package `{}` because it "
+                    "has an unsupported package build type: `{}`"
+                ).format(pkg.name, build_type))
+
+                wide_log(clr("[build] Note: Available build types:"))
+                for bt_name in build_job_creators.keys():
+                    wide_log(clr("[build]  - `{}`".format(bt_name)))
+
+    except CommandMissing as e:
+        sys.exit(clr("[build] @!@{rf}Error:@| {0}").format(e))
```
mikepurvis: Intent here would be to use a tuple ultimately to catch other kinds of exceptions that may be necessary to throw for problems discovered during job creation.

catkin_tools/jobs/utils.py
```diff
@@ -25,6 +25,21 @@
 from catkin_tools.execution.events import ExecutionEvent
 
 
+class CommandMissing(Exception):
+    '''A required command is missing.'''
+
+    def __init__(self, name):
+        super(CommandMissing, self).__init__(
+                'Cannot find required tool `%s` on the PATH, is it installed?' % name)
```
mikepurvis: Happy to move this to its own `exceptions.py` file, didn't want to bother with that unless there was a clear need.

catkin_tools/jobs/utils.py
```diff
@@ -25,6 +25,20 @@
...
+
+
+def require_command(name, which):
+    if not which:
+        import errno
```
wjwwood: Why do you need `errno` here?

catkin_tools/verbs/catkin_build/build.py
wjwwood: This looks ok, but what do you think about putting this try catch around all of the the `build_isolated_workspace` function? Indenting this much code is just making it harder to read imo, and I think it should be ok to include the rest of the function in this try-catch right?

mikepurvis: You mean moving it to `catkin_build/cli.py`? That can be done. It'll mean importing the excepting there instead of here, which probably doesn't really matter.

The main thing is not catching everything— we still want stack traces for unknown problems.

catkin_tools/jobs/utils.py
mikepurvis: Oh fail, that's a leftover— an earlier version of this threw OSError instead. Removed.

mikepurvis: The function is simple enough now that it could probably just be inlined; I don't really care one way or another about that.

catkin_tools/verbs/catkin_build/build.py
wjwwood: Yeah, that also sounds fine, though I was talking about just making a new function here that calls the current function in a try-catch, but your suggestion makes more sense.

In the end it's a small thing, but among other things it makes this diff hard to read.

catkin_tools/jobs/utils.py
wjwwood: It's fine as-is for me.

catkin_tools/verbs/catkin_build/build.py
mikepurvis: It would be great if git/github's diff were smarter about recognizing blocks that have changed only in indentation.

gavanderhoorn: Not sure whether there is any UI for this, but you can get the diff renderer to ignore whitespace changes by appending `w=1` to the url: https://github.com/catkin/catkin_tools/pull/455/files?w=1.

mikepurvis: Oh, nice! TIL.

mikepurvis:
Okay, I think this is good now.

catkin_tools/jobs/utils.py
mikepurvis: (A good motivation to do so might be eliminating the use of `sys.exit` within `build_isolated_workspace`, in favour of throwing other exceptions that resolve to the string errors messages. Doing this would make the function more cooperative as far as being wrapped or called from other tools.)

PR #458, opened 2017-05-26
Implementation of pluggable spaces.

mikepurvis:
As discussed in https://github.com/mikepurvis/catkin_tools_document/issues/3#issuecomment-301650982.

This permits plugins to register new first-class space types. Some potential candidate uses:

- A `docs` space for stashing doxygen/sphinx output.
- A `tests` space for placing junit testing reports.
- A `merge` space for placing post-build merged installspaces assembled from workspaces built with the `--install-isolated` option.
- A `symbols` space for placing debug symbols stripped from binaries during/after the the build.

I haven't yet updated the `catkin_clean` verb to take advantage of this, but will do so shortly.

FYI @wjwwood @tspicer01

----

Test failure is due to a [breakage upstream with the Sphinx spellcheck](https://bitbucket.org/dhellmann/sphinxcontrib-spelling/issues/13/with-sphinx-161-contractions-result-in). I'd suggest either turning off the spellcheck for now, or locking into an older version of Sphinx where the problem didn't happen.

### DIFF
--- a/catkin_tools/argument_parsing.py
+++ b/catkin_tools/argument_parsing.py
@@ -89,7 +89,7 @@ def add_cmake_and_make_and_catkin_make_args(parser):
 
     add = parser.add_mutually_exclusive_group().add_argument
     add('--make-args', metavar='ARG', dest='make_args', nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to make.'
+        help='Arbitrary arguments which are passes to make. '
              'It collects all of following arguments until a "--" is read.')
     add('--no-make-args', dest='make_args', action='store_const', const=[], default=None,
         help='Pass no additional arguments to make (does not affect --catkin-make-args).')
@@ -97,7 +97,7 @@ def add_cmake_and_make_and_catkin_make_args(parser):
     add = parser.add_mutually_exclusive_group().add_argument
     add('--catkin-make-args', metavar='ARG', dest='catkin_make_args',
         nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to make but only for catkin packages.'
+        help='Arbitrary arguments which are passes to make but only for catkin packages. '
              'It collects all of following arguments until a "--" is read.')
     add('--no-catkin-make-args', dest='catkin_make_args', action='store_const', const=[], default=None,
         help='Pass no additional arguments to make for catkin packages (does not affect --make-args).')
--- a/catkin_tools/argument_parsing.py
+++ b/catkin_tools/argument_parsing.py
@@ -82,14 +82,14 @@ def add_cmake_and_make_and_catkin_make_args(parser):
 
     add = parser.add_mutually_exclusive_group().add_argument
     add('--cmake-args', metavar='ARG', dest='cmake_args', nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to CMake. '
+        help='Arbitrary arguments which are passed to CMake. '
              'It collects all of following arguments until a "--" is read.')
     add('--no-cmake-args', dest='cmake_args', action='store_const', const=[], default=None,
         help='Pass no additional arguments to CMake.')
 
     add = parser.add_mutually_exclusive_group().add_argument
     add('--make-args', metavar='ARG', dest='make_args', nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to make. '
+        help='Arbitrary arguments which are passed to make. '
              'It collects all of following arguments until a "--" is read.')
     add('--no-make-args', dest='make_args', action='store_const', const=[], default=None,
         help='Pass no additional arguments to make (does not affect --catkin-make-args).')
@@ -97,7 +97,7 @@ def add_cmake_and_make_and_catkin_make_args(parser):
     add = parser.add_mutually_exclusive_group().add_argument
     add('--catkin-make-args', metavar='ARG', dest='catkin_make_args',
         nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to make but only for catkin packages. '
+        help='Arbitrary arguments which are passed to make but only for catkin packages. '
              'It collects all of following arguments until a "--" is read.')
     add('--no-catkin-make-args', dest='catkin_make_args', action='store_const', const=[], default=None,
         help='Pass no additional arguments to make for catkin packages (does not affect --make-args).')
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -49,19 +49,12 @@ class Context(object):
     This context can be locked, so that changing the members is prevented.
     """
 
-    DEFAULT_LOG_SPACE = 'logs'
-    DEFAULT_SOURCE_SPACE = 'src'
-    DEFAULT_BUILD_SPACE = 'build'
-    DEFAULT_DEVEL_SPACE = 'devel'
-    DEFAULT_INSTALL_SPACE = 'install'
+    CATKIN_SPACES_GROUP = 'catkin_tools.spaces'
+
+    SPACES = {}
 
     STORED_KEYS = [
         'extend_path',
-        'source_space',
-        'log_space',
-        'build_space',
-        'devel_space',
-        'install_space',
         'devel_layout',
         'install',
         'isolate_install',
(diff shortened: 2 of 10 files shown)


setup.py
```diff
@@ -155,6 +155,13 @@ def run(self):
             'catkin = catkin_tools.jobs.catkin:description',
             'cmake = catkin_tools.jobs.cmake:description',
         ],
+        'catkin_tools.spaces': [
+            'build = catkin_tools.spaces.build:description',
+            'devel = catkin_tools.spaces.devel:description',
+            'install = catkin_tools.spaces.install:description',
+            'log = catkin_tools.spaces.log:description',
+            'source = catkin_tools.spaces.source:description',
```
mikepurvis: If desired, these could be collapsed into a single file, so:

```
            'build = catkin_tools.spaces:build',
             'devel = catkin_tools.spaces:devel',
             'install = catkin_tools.spaces:install',
             'log = catkin_tools.spaces:log',
             'source = catkin_tools.spaces:source',
```

mikepurvis:
Okay, the story for properly handling `catkin clean` is a bit of an interesting one. I'd be delighted for some input on this from the maintainers. Cleaning is really two distinct functions which have very little to do with each other:

 - [clean_profile](https://github.com/catkin/catkin_tools/blob/2442e1f971ae6bbd0c14fa6da183ee095ae099e3/catkin_tools/verbs/catkin_clean/cli.py#L151), which cleans at the space level. This one could I think be handled by a generic implementation which just removes the specified directory, though potentially there could be the option for the space to supply its own more sophisticated function that has other behaviours.
- [clean_packages](https://github.com/catkin/catkin_tools/blob/2442e1f971ae6bbd0c14fa6da183ee095ae099e3/catkin_tools/verbs/catkin_clean/clean.py#L96), which cleans package-by-package _using the functions specified in the `build_type` plugins_. To support this properly I think would require an additional function provided by the space for "clean package X from your space". In some cases that simply may not be applicable, so the default implementation should probably be a no-op. But supplying at least a hook here would permit a `catkin document; catkin clean foo` sequence which properly removes documentation generated for one of the packages present.

The interesting thing about this the question of whether the existing `create_catkin_clean_job` and `create_cmake_clean_job` (which are almost identical) should be broken up and have the relevant common space-related cleanup functions moved into space-supplied clean-package functions.

https://github.com/catkin/catkin_tools/blob/2442e1f971ae6bbd0c14fa6da183ee095ae099e3/catkin_tools/jobs/catkin.py#L485
https://github.com/catkin/catkin_tools/blob/2442e1f971ae6bbd0c14fa6da183ee095ae099e3/catkin_tools/jobs/cmake.py#L328

I think in the short term I will focus on the clean_profile side of things as my immediate interest is getting support for `catkin clean -y` to do the right thing. But perhaps if there's a resolution on how to handle cleaning individual packages for plugin-supplied spaces, that can be tackled in a follow-on PR.

mikepurvis:
@wjwwood Thoughts on the above discussion? IMO this PR is mergeable as-is, and the work to improve `catkin_clean` could be done as a follow-up. That might actually be desirable in terms of reducing the overall size of each change.

catkin_tools/argument_parsing.py
```diff
@@ -89,15 +89,15 @@ def add_cmake_and_make_and_catkin_make_args(parser):
 
     add = parser.add_mutually_exclusive_group().add_argument
     add('--make-args', metavar='ARG', dest='make_args', nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to make.'
+        help='Arbitrary arguments which are passes to make. '
```
wjwwood: Why the trailing space, also there is a typo in this help (not new), `passes` -> `passed`

catkin_tools/argument_parsing.py
```diff
@@ -89,15 +89,15 @@ def add_cmake_and_make_and_catkin_make_args(parser):
...
              'It collects all of following arguments until a "--" is read.')
     add('--no-make-args', dest='make_args', action='store_const', const=[], default=None,
         help='Pass no additional arguments to make (does not affect --catkin-make-args).')
 
     add = parser.add_mutually_exclusive_group().add_argument
     add('--catkin-make-args', metavar='ARG', dest='catkin_make_args',
         nargs='+', required=False, type=str, default=None,
-        help='Arbitrary arguments which are passes to make but only for catkin packages.'
+        help='Arbitrary arguments which are passes to make but only for catkin packages. '
```
wjwwood: Same question and typo, lol.

catkin_tools/context.py
```diff
@@ -75,12 +68,57 @@ class Context(object):
...
+            return getattr(self, '__%s_space' % space)
+
+        def space_setter(self, value):
+            if self.__locked:
+                raise RuntimeError("Setting of context members is not allowed while locked.")
+            setattr(self, '__%s_space' % space, value)
+            setattr(self, '__%s_space_abs' % space, os.path.join(self.__workspace, value))
+
+        def space_exists(self):
+            "Returns true if the space exists"
```
wjwwood: Use `"""` here, unless you have a specific reason not to?

catkin_tools/context.py
```diff
@@ -75,12 +68,57 @@ class Context(object):
...
+            space_abs = getattr(self, '__%s_space_abs' % space)
+            return os.path.exists(space_abs) and os.path.isdir(space_abs)
+
+        setattr(cls, '%s_space' % space, property(space_getter, space_setter))
+        setattr(cls, '%s_space_abs' % space, property(space_abs_getter))
+        setattr(cls, '%s_space_exists' % space, space_exists)
+
+    @classmethod
+    def setup_space_keys(cls):
+        '''
```
wjwwood: I believe `"""` is the convention most used in this file, please use that rather than `'''`.

catkin_tools/context.py
```diff
@@ -411,20 +446,20 @@ def summary(self, notes=[]):
                 "--init`.")]
         if not self.source_space_exists():
             summary_warnings += [clr(
-                "Source space `@{yf}{_Context__source_space_abs}@|` does not yet exist.")]
+                "Source space `@{yf}{__source_space_abs}@|` does not yet exist.")]
```
wjwwood: Shouldn't this use the `self.source_space_abs` property? Why is the `_Context` prefix no longer needed, isn't it required due to the name mangling:

https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references

docs/mechanics.rst
```diff
@@ -235,7 +235,6 @@ Sourcing these setup scripts adds this workspace and any "underlaid" workspaces
 
 .. [1] GNU/Linux Only
 .. [2] Mac OS X Only
-.. [3] Windows Only
```
wjwwood: Was this unused, or...?

setup.py
wjwwood: This is fine I think.

wjwwood:
I'm in favor of merging this and dealing with the generalization of catkin clean afterwards. Honestly I didn't have much to do with `catkin clean`, so I probably have less of an idea of how to proceed there than you do. But everything you said makes sense with respect to updating `clean_profile` being straightforward, but `clean_package` needing to be generalized to support different approaches to cleaning a package out of a space.

catkin_tools/argument_parsing.py
mikepurvis: The string gets concatenated with the string immediately below. Without the change, you get:

```
$ catkin build -h
[ ... ]
  --make-args ARG [ARG ...]
                        Arbitrary arguments which are passes to make.It
                        collects all of following arguments until a "--" is
                        read.
```

Will fix passes -> passed as well.

catkin_tools/context.py
mikepurvis: This is a C&P from the old `source_space_exists` function. Will update.

catkin_tools/context.py
mikepurvis: I would have though so, but this was what worked. It may be an indication that I'm not using `property` correctly in the setup of these generated properties. I shall investigate and report back.

docs/mechanics.rst
mikepurvis: Yes, it was producing a Sphinx build warning.

catkin_tools/context.py
mikepurvis: Yeah, I'm not really sure what is going on here, but the behaviour is definitely correct in that each instance of `Context` has its own set of these properties, eg:

```
$ ~/env/bin/python
Python 2.7.6 (default, Oct 26 2016, 20:30:19)
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from catkin_tools.context import Context
>>> Context.setup_space_keys()
>>> c = Context('.')
>>> c.source_space
'src'
>>> d = Context('.')
>>> d.source_space
'src'
>>> c.source_space = 'foo'
>>> c.source_space
'foo'
>>> c.source_space_abs
'/home/mikepurvis/test_ws/foo'
>>> d.source_space
'src'
```

catkin_tools/context.py
mikepurvis: Fixed.

Potentially this could be called on import, but it felt saner to defer and do it lazily on first use.

catkin_tools/context.py
```diff
@@ -49,19 +49,12 @@ class Context(object):
...
     """
 
-    DEFAULT_LOG_SPACE = 'logs'
-    DEFAULT_SOURCE_SPACE = 'src'
-    DEFAULT_BUILD_SPACE = 'build'
-    DEFAULT_DEVEL_SPACE = 'devel'
-    DEFAULT_INSTALL_SPACE = 'install'
+    CATKIN_SPACES_GROUP = 'catkin_tools.spaces'
+
+    SPACES = {}
```
mikepurvis: Given the usage patterns seen below, there might be a case to be made here for using a list rather than a dict, and creating a new `Space` class rather than just assigning the entry_point dict directly.

I don't think it much matters for now, so my inclination would be to leave it and add the extra layer as need demands.

setup.py
mikepurvis: Cool. Yes I think it makes sense to give them their own files, particularly if we're likely to see functions attached to these related to cleaning or other behaviours.

Issue #459, opened 2017-05-30
Package list caching in zsh autocompletion does not work


pklip:
### System Info

* Operating System: Linux abtz0gta 4.8.0-46-generic #49~16.04.1-Ubuntu SMP Fri Mar 31 14:51:03 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 2.7.12
* Version of catkin_tools: catkin_tools 0.4.4
* ROS Distro: kinetic
* Shell: zsh

### Issue

Autocompletion in zsh with _catkin does not actually use the caching mechanism to cache the package list, although the code intends to do so. Only the environment variable _workspace_packages is set once in the current shell and then never saved to the cache or updated anymore.

### Expected Behavior

On autocompletion (<kbd>TAB</kbd>) the current package list should be saved to [catkin_workspace]/.catkin_tools/cache/workspace_packages and be retrieved on further completions if the workspace did not change and the variable _workspace_packages has not already been set

### Actual Behavior

The cache is never used.

### Steps to Reproduce the Issue

Copy completion/_catkin to /usr/local/share/zsh/site-functions (or anywhere in your $fpath) and make sure compinit is called in your .zshrc (preferrably at the very end after any plugins).

timonegk:
The issue was fixed in #475.

PR #460, opened 2017-05-30
Fixed zsh completion to actually use zsh caching for package list (#459)

pklip:
See Issue #459.

Fixes:
* Due to typo in line 105 _workspace_source_space was never set and "-ot" comparison always failed later on. 
* In line 131 return value of _catkin_get_enclosing_workspace was saved as _workspace_root which only returns 0 or 1. The global variable is set internally in the function.
* If statement in line 151 to check when to rebuild cache was wrong and incomprehensible.


wjwwood:
@pklip sorry, while trying to rebase I screwed this pr up, see https://github.com/catkin/catkin_tools/pull/475.

PR #461, opened 2017-05-30
Disable Sphinx spell check for now.

mikepurvis:
Test builds are failing at present, pending resolution to this issue: https://bitbucket.org/dhellmann/sphinxcontrib-spelling/issues/13

PR #462, opened 2017-05-31
Revert "Disable Sphinx spell check for now."

wjwwood:
Only do this when the permanent solution is found, see: https://github.com/catkin/catkin_tools/pull/461#issue-232341090

Reverts catkin/catkin_tools#461

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ script:
   # Build documentation
   - pushd docs
   - make html
-  - sphinx-build . build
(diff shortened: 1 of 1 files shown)


pklip:
I would appreciate this being merged so that #460 can also be merged :+1:

mikepurvis:
@pklip The change which allows #460 to pass CI has already merged. If you force-push to your branch, it will now pass.

This undoes that change for when upstream fixes the underlying problem.

PR #466, opened 2017-06-09
Fix typo in documentation of option env-cache

mbinna:
The text mixed up the two sub-words of the option. In the latest catkin_tools version 0.4.4, the option is named ``env-cache``, not ``cache-env``.


mikepurvis:
:+1:

PR #467, opened 2017-06-12
Fixed doc and completion based on current CLI for interleave output

cmuck:
The current CLI for the interleave output proposes to use --interleave-output instead of --interleave.
In addition some trailing whitespaces were removed.


wjwwood:
Thanks!

Issue #445, opened 2017-03-15
pip installing catkin_tools doesn't install trollius


jlack1987:
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

### System Info

* Operating System: `uname -a` - 4.8.0-41-generic #44~16.04.1-Ubuntu SMP Fri Mar 3 17:11:16 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: `python --version` - 3.5
* Version of catkin_tools: `catkin --version` OR *Git revision* - 0.4.4
* ROS Distro: `echo $ROS_DISTRO` - kinetic

### Build / Run Issue
Run issue

### Expected Behavior
[//]: # (Please describe what you expected or think should happen.)
I would just expect it to configure the catkin workspace to extend `/opt/ros/kinetic` and use the install space.

### Actual Behavior

[//]: # (Please describe what actually happens.)
For context I have been doing all this in a docker. To reproduce, the docker ros:kinetic-ros-base will do. The following snippet shows the problem i'm having.
```
$ catkin config --extend /opt/ros/$ROS_DISTRO/ --install
Traceback (most recent call last):
  File "/usr/local/bin/catkin", line 7, in <module>
    from catkin_tools.commands.catkin import main
  File "/usr/local/lib/python3.5/dist-packages/catkin_tools/commands/catkin.py", line 23, in <module>
    from catkin_tools.common import is_tty
  File "/usr/local/lib/python3.5/dist-packages/catkin_tools/common.py", line 23, in <module>
    import trollius as asyncio
ImportError: No module named 'trollius'
```

### Steps to Reproduce the Issue

```bash
pip install -U catkin_tools
catkin config
```

slivingston:
I also observe this behaviour on a similar platform: Ubuntu 16.04, Python 3.5.2, catkin_tools v0.4.4 installed via `pip`.

Part of a solution was merged in pull request #282, but it remains to express the dependency in setup.py. In particular, `trollius` should be listed [in `install_requires` in setup.py](https://github.com/catkin/catkin_tools/blob/8b2d68f65f6a43d732dd1a2467f287b8fecfd3ae/setup.py#L13-L18).

atomoclast:
I too am experiencing this issue.

slivingston:
[...]

PR #474, opened 2017-07-16
Add trollius to setup.py `install_requires` list

slivingston:
closes issue #445
This patch can be viewed as continuing the work of pull request #282

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,6 +15,7 @@ install_requires = [
     'setuptools',
     'PyYAML',
     'osrf-pycommon > 0.1.1',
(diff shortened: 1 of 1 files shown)


slivingston:
Note that the install context that leads to the error reported in #445 is not detected by the Travis CI configuration because `trollius` is explicitly installed [on line 26 in the `before_install` clause of .travis.yml](https://github.com/catkin/catkin_tools/blob/2280a2a51f86949af7896af3876cb1f4c6e7bf46/.travis.yml#L26).

Issue #464, opened 2017-06-08
Verb alias doesn't support arguments with spaces


Xfel:
I'm trying to create a verb alias for generating eclipse project files. The regular command for this is:
```bash
catkin build  --force-cmake -G"Eclipse CDT4 - Unix Makefiles"
```
So I tried to define the alias like this:
```yaml
eclipse: 'build  --force-cmake -G"Eclipse CDT4 - Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug'
```

This does not work since the argument contains spaces and the quoting is not recognized.

I'd suggest to either use something like shlex to process the alias definition, or allow using a yaml list to define the arguments.

wjwwood:
That sounds reasonable, please consider a pull request to either use shlex or allow commands to be a list in YAML. Both sound reasonable to me.

wjwwood:
I think this is closed with https://github.com/catkin/catkin_tools/pull/476.

PR #476, opened 2017-07-18
Enable arguments with spaces in alias definition

Xfel:
This is the solution for #464. It supports both using quotes as understood by shlex in the string and using a yaml list to specify the single arguments.

The only downside to this is that it uses the function `pipes.quote` which is apparently deprecated.

### DIFF
--- a/catkin_tools/commands/catkin.py
+++ b/catkin_tools/commands/catkin.py
@@ -19,6 +19,7 @@ from datetime import date
 import os
 import pkg_resources
 import sys
+import pipes
 
 from catkin_tools.common import is_tty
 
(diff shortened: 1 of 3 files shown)


Xfel:
Please excuse the two commits, I couldn't get the tests to run on my machine, so I had to push and let travis do the work. Those issues should be fixed now. I did however find another issue that is not covered by the tests, comes from a moved function between python2.7 and python3.4. What is the recommended way to deal with this issue?

wjwwood:
> Please excuse the two commits, I couldn't get the tests to run on my machine, so I had to push and let travis do the work.

No worries, I'll squash with GitHub when merging.

PR #477, opened 2017-07-25
Treat spaces as handled config context options

mbinna:
Adding keyed parameters for those options inhibits the following warning when using profiles.

> Unhandled config context options: {}

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -247,6 +247,11 @@ class Context(object):
         space_suffix=None,
         whitelist=None,
         blacklist=None,
+        devel_space=None,
+        source_space=None,
+        log_space=None,
+        build_space=None,
+        install_space=None,
         **kwargs
     ):
         """Creates a new Context object, optionally initializing with parameters


mbinna:
With this change, the spaces are not recognized anymore. For example, when switching profiles with different space suffixes, the directories to the spaces are not updated correctly. So this change does not work as expected. However, I am still wondering how to fix the warning 

> Unhandled config context options: { ... }

Therefore, I am keeping this pull request open.

wjwwood:
Can you please give a more detailed set of steps to reproduce the warning?

mbinna:
@wjwwood Sure. Here we go:

1. Start with clean workspace with only src/ directory (no .catkin_tools/ directory)
2. `catkin config` -> does not show a warning
3. `catkin profile add release`
4. `catkin profile set release`
5. `catkin config` -> shows the warning 

Tested with HEAD of master (da79d37bd5912c4d9e0ab1b097554388a1a9592f).

wjwwood:
@mbinna I was able to reproduce the issue, but I had to do `catkin config --init` in the first step.

I'll look into the issue asap.

racko:
This change doesn't quite work. Try

```
$ catkin config -d path/to/catkin_ws/devel_linked
```

and check the `Devel Space:` output. It will still say `path/to/catkin_ws/devel`.

Explanation: If you list the spaces as parameters then they won't appear in kwargs. As a result they won't be used in the following code:
https://github.com/catkin/catkin_tools/blob/47a5b930415bed71077aef3f3ad8e5965e4278cb/catkin_tools/context.py#L309-L315
Instead the defaults will be read from `space_dict`. That's why the output above stills says `path/to/catkin_ws/devel`: It's the default.

That's why I created https://github.com/catkin/catkin_tools/pull/489. Here I pop the kwargs during the loop and check the remaining number of kwargs afterwards: https://github.com/catkin/catkin_tools/pull/489/commits/f67767c4b50f010f7309f5929d6a05c7f0b17e63

mikepurvis:
Replaced by #489.

PR #480, opened 2017-08-12
Fixed jobserver not working with GNU make >= 4.2

racko:
--jobserver-fds has been renamed to --jobserver-auth.

Also calling make with both --jobserver-auth and -j leads to 'warning:
-jN forced in submake: disabling jobserver mode.'

See https://lists.gnu.org/archive/html/info-gnu/2016-05/msg00013.html

### DIFF
--- a/catkin_tools/execution/job_server.py
+++ b/catkin_tools/execution/job_server.py
@@ -63,23 +63,25 @@ def memory_usage():
     return None, None
 
 
-JOBSERVER_SUPPORT_MAKEFILE = b'''
+JOBSERVER_SUPPORT_MAKEFILE_OLD = b'''
 all:
 \techo $(MAKEFLAGS) | grep -- '--jobserver-fds'
 '''
 
+JOBSERVER_SUPPORT_MAKEFILE = b'''
+all:
+\techo $(MAKEFLAGS) | grep -- '--jobserver-auth'
+'''
 
-def test_gnu_make_support():
+
+def test_gnu_make_support_common(makefile_content):
     """
-    Test if the system 'make' supports the job server implementation.
-
-    This simply checks if the `--jobserver-fds` option is supported by the
-    `make` command. It does not tests if the jobserver is actually working
-    properly.
+    Test if "make -f MAKEFILE -j2" runs successfullyn when MAKEFILE
+    contains makefile_content.
     """
 
     fd, makefile = mkstemp()
-    os.write(fd, JOBSERVER_SUPPORT_MAKEFILE)
+    os.write(fd, makefile_content)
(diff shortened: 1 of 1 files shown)


racko:
@wjwwood I noticed that you commented on my other pull request #481, but not on this one. Did you just not find the time or are there other issues?

I assumed that the build failure has nothing to do with my commit. I don't understand the error messages like "VAR2 NOT DEFINED!" and if anything, my commit could only cause catkin to build with a single processor. It could not cause this kind of build failure, or am I missing something? Could I retrigger the build bot?

racko:
I noticed that the build log mentioned that the build timed out. I changed the commit so the deprecated --jobserver-fds behavior still works. Strange that "make-with-auth + catkin-with-fds" simply runs single-threaded while "make-with-fds + catkin-with-auth"  seems to block and do nothing oO

Well ... now it works :)

wjwwood:
I think I was poking the CI job to see if the issue was your pr or something wrong with the CI setup.

Based on what you told me, it sounds like might be a race condition? I'll try to spend some time testing it out more thoroughly but I have some other stuff to do today.

racko:
Oh, I think (hope) that my initial pull request was buggy, but that the new version is fine :P

v4hn:
Thank you @racko .
@wjwwood I stayed away from catkin_tools up to now because it did nothing but idle on my system with a current gnu make.
This patch resolves the problem and works well. Please review again.

jbohren-hbr:
@v4hn have you verified that this works with gcc 4.1 and gcc 4.2?

v4hn:
> have you verified that this works with gcc 4.1 and gcc 4.2?

I guess you mean `make`, not `gcc`.

No I have not, but CI did.

jbohren-hbr:
> > have you verified that this works with gcc 4.1 and gcc 4.2?

> I guess you mean `make`, not `gcc`.

You got me. 

> No I have not, but CI did. I tested (and for the last 3 months used...) it with gnu make 4.2.1

That's good to hear, we'll pull it here at honeybee for further testing on make 4.1.

mikepurvis:
I haven't yet hit this as I'm on Xenial/Bionic in my day-to-day and gnu make 4.2+ doesn't drop until Cosmic:

https://packages.ubuntu.com/search?keywords=make

However, it seems like a harmless change. Should it go in?

v4hn:
> However, it seems like a harmless change. Should it go in?

Yes. This is still a mandatory patch to get catkin_tools working with current make packages.

mikepurvis:
Sounds good! Merging now, and will validate before release on Ubuntu X and Homebrew.

PR #481, opened 2017-08-14
Fixed '_catkin_get_enclosing_package:14: command not found: xmllint'

racko:
Occurred when doing auto-complete at 'catkin build --work[TAB]'. Fixing
the xmllint path led to 'warning: failed to load external entity
"$path/package.xml"'

Don't know enough about these zsh completion scripts to understand why we need absolute paths but "/bin/readlink" was already there.

Also I don't like removing the quotes around the $path variable because it might contain spaces, but again, `"$(/bin/readlink -f $path/..)"` did it first.

### DIFF
--- a/completion/_catkin
+++ b/completion/_catkin
@@ -85,7 +85,7 @@ _catkin_get_enclosing_package() {
     if [[ -e "$path/package.xml" ]]; then
       _debug_log "Found package root $path"
       # Get the package name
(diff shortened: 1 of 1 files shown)


wjwwood:
@racko thanks for the patch. Did you try it with the `'`'s left in? Or does it fail with them in?

racko:
Closed because it turned out that the issues where more complex: See #482

Issue #297, opened 2016-02-17
pkg_resources.entry_points is really slow


jbohren:
On my 2.9Ghz i7 with an EVO850 SSD running Ubuntu 12.04:

```bash
$ time python -c 'from pkg_resources import load_entry_point'
python -c 'from pkg_resources import load_entry_point'  0.14s user 0.01s system 97% cpu 0.152 total
```

This means that using calls to the catkin script in tab-completion is a very laggy experience, and explains why there's a ubiquitous >200ms latency when running `catkin` commands. I think as it stands, catkin_tools is over-using the `entry_points` interface. For example it's not necessary for the main `catkin` command, nor is it really necessary for the built-in verbs. 

I think the tool could be much more responsive if we used an alternative to `entry_points` for the verbs and build types. For example, mercurial uses a simple interface in `.hgrc` to specify plugins explicitly, and I think we would be better off using a system like that.

Currently `catkin` creates `~/.config/catkin` and `catkinrc` could contain something like:

```yaml
verbs:
  lint: catkin_lint.main
build-types:
  catkin: catkin_tools.build_types.catkin
  cmake: catkin_tools.build_types.cmake
  ament: ament_pkg.catkin_plugin 
```

Loading and reading a YAML file or INI is under 50ms:

```bash
$ time python -c 'import yaml; f = open(".config/catkin/verb_aliases/00-default-aliases.yaml","r"); print(yaml.load(f)); f.close();' 
{'bt': 'b --this', 'p': 'create pkg', 'b': 'build', 'ls': 'list', 'install': 'config --install', 'test': 'build --verbose --make-args test --', 'run_tests': 'build --verbose --catkin-make-args run_tests --'}
python -c   0.04s user 0.02s system 94% cpu 0.064 total
```

Related: #295, #209

ninjaaron:
I wrote a tiny module called `fastentrypoints` that monkey patches the mechanism behind `entry_points` to generate scripts that don't import pkg_resources.

https://github.com/ninjaaron/fast-entry_points

mikepurvis:
Unless I'm misunderstanding something, the fast entry points script will only improve things for the initial bootstrap from script into module— we _also_ use pkg_resources here for handling plugins, so really get the benefit, we'd need to do a couple things:

- fast entry points for the console scripts
- import pkg_resources on a background daemon thread, and only block on needing it under the following conditions:
    * displaying `catkin -h`, in order to list available verbs.
    * invoking a non-built in verb.
[...]

PR #483, opened 2017-08-16
Improved zsh completion time from ~0.5s to near instant

racko:
The "catkin locate -s" takes around 0.5s. It gets called whenever the cache is checked.

### DIFF
--- a/completion/_catkin
+++ b/completion/_catkin
@@ -102,7 +102,15 @@ _catkin_packages_caching_policy() {
   if [[ -z "$_workspace_source_space" ]]; then
     _debug_log "Searching for source space..."
 
-    _workspace_source_space="$(catkin locate -s --quiet)"
+    _catkin_get_enclosing_workspace $_workspace_root_hint
+
+    # We try this special case first since it is the recommended layout
+    # and calling "catkin locate" is much slower
+    _workspace_source_space="$_workspace_root/src"
+
+    if [[ ! -d "$_workspace_source_space" ]]; then
+        _workspace_source_space="$(catkin locate -s --quiet)"
+    fi
 
     # If the source space can't be found, regenerate cache
     if [[ "$?" -ne "0" ]]; then


completion/_catkin
```diff
@@ -102,7 +102,8 @@ _catkin_packages_caching_policy() {
   if [[ -z "$_workspace_source_space" ]]; then
     _debug_log "Searching for source space..."
 
-    _workspace_source_space="$(catkin locate -s --quiet)"
+    _catkin_get_enclosing_workspace $_workspace_root_hint
+    _workspace_source_space="$_workspace_root/src"
```
wjwwood: This is a flawed assumption, the source directory is not always `<workspace dir>/src`. Since it can be any folder, that's why `catkin locate` is required to look up the setting.

racko:
**tl;dr:** No idea how to solve this "properly". What about adding a special case for "$_workspace_root/src" since it's what works for 99% of all users? It is the recommended layout after all ...

---

Well ... 
```
import catkin_tools.context
ctx = catkin_tools.context.Context.load('/home/racko/catkin_ws', None, type('Options', (object,), dict(space='src'))(), load_env=False)
print ctx.source_space_abs
```
takes around 250ms.

```
from catkin_tools.metadata import get_metadata, get_active_profile
print get_metadata('/home/racko/catkin_ws', get_active_profile('/home/racko/catkin_ws'), 'config')['source_space']
```
takes around 175ms. Calling `get_metadata` or `get_active_profile` alone also takes around 175ms, so among other things, we can conclude that calling `get_active_profile` and then reading the config.yaml with another tool will not be good enough. But still it is interesting to measure.

```
import yaml
print yaml.load(open('/home/racko/catkin_ws/.catkin_tools/profiles/default/config.yaml'))['source_space']
```
takes around 30ms.

```
yaml2json - < /home/racko/catkin_ws/.catkin_tools/profiles/default/config.yaml | jq '.source_space'
```
takes around 20ms.

Also caching the source space path is not an option: To check if the cache is invalid, we need to check the date of the config.yaml, and getting the path takes around 175ms again ...

Reimplementing the get_active_profile logic in zsh (or some compiled language) would be trivial and fast enough, but duplicating the logic is of course not an option. And I'm certain that implementing the logic in C++ and using it in python (similar to tf) was decided against long before I came along, besides it being to big of a change for this pull request :)

wjwwood:
If the issue is only due to asking for the version number, perhaps we could delay acquiring that information until it is asked for.

However, I think in general the entry points performance is a bigger issue, see: https://github.com/catkin/catkin_tools/issues/297

racko:
I close this pull request because the proposed solution isn't actually working:


> What about adding a special case for "$_workspace_root/src" since it's what works for 99% of all users? It is the recommended layout after all ...
> 
> Not even this special case works: Finding "$_workspace_root/src" does not mean that it is the source space :(

Also my suggestion that replacing `pkg_resources.require("catkin_tools")[0].version` would fix the problem seems to be false: While it certainly speeds up migrate_metadata, I couldn't observe a speed-up for `catkin locate -s` or the zsh auto-complete. I guess that `pkg_resources.require("catkin_tools")` or something similar gets called anyhow.

PR #487, opened 2017-10-13
Fixed 'catkin profile set ...' error message

racko:
Calling 'catkin profile set random_garbage', users where being presented
with 'catkin profile: error: Profile `r` does not exist ...'


racko:
Here's the CI output. I guess that's not my fault, right?

```
The path /usr/local/bin/python3 (from --python=/usr/local/bin/python3) does not exist
.travis.before_install.bash: line 12: venv/bin/activate: No such file or directory
The command "source .travis.before_install.bash" failed and exited with 1 during .
Your build has been stopped.
/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found
```

mikepurvis:
Looks like a good change, thanks!

PR #497, opened 2018-01-05
Add bionic.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -3,5 +3,5 @@ Depends: python-argparse, python-setuptools, python-catkin-pkg (> 0.2.9), python
 Depends3: python3-setuptools, python3-catkin-pkg (> 0.2.9), python3-yaml
 Conflicts: python3-catkin-tools
 Conflicts3: python-catkin-tools
-Suite: trusty vivid wily xenial wheezy yakkety zesty artful jessie stretch
+Suite: trusty vivid wily xenial wheezy yakkety zesty artful bionic jessie stretch
 X-Python3-Version: >= 3.2


tests/test_code_format.py
```diff
@@ -6,9 +6,7 @@ def test_flake8():
     """Test source code for pyFlakes and PEP8 conformance"""
     this_dir = os.path.dirname(os.path.abspath(__file__))
     source_dir = os.path.join(this_dir, '..', 'catkin_tools')
-    cmd = ['flake8', source_dir, '--count', '--max-line-length=120']
-    # work around for https://gitlab.com/pycqa/flake8/issues/179
-    cmd.extend(['--jobs', '1'])
+    cmd = ['flake8', source_dir, '--count', '--max-line-length=120', '--ignore=E722']
```
dirk-thomas: I would suggest to rather fix the code then ignore this valuable error.

wjwwood: @clalancette just revert this commit and I'll merge it without green CI. I'll fix up the travis stuff separately.

racko:
Would you mind having a look at other pull requests? :sweat_smile:

tests/test_code_format.py
clalancette: > @clalancette just revert this commit and I'll merge it without green CI. I'll fix up the travis stuff separately.

Sounds good, will do.

wjwwood:
> Would you mind having a look at other pull requests? 😅

Not today. When I have time I will, but I don't know when that will be.

Please don't ask on unrelated issues in the future. If you want to ask, please do so on individual issues and pull requests on which you've been involved.

PR #509, opened 2018-03-12
Silence E722 flake8 warnings, fix OSX builds.

mikepurvis:
Should unbreak travis testing the pending PRs.

### DIFF
--- a/.travis.before_install.bash
+++ b/.travis.before_install.bash
@@ -4,10 +4,14 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
   sudo apt-get install enchant -y
 elif [ "$TRAVIS_OS_NAME" == "osx" ]; then
   if [ "$PYTHON" == "/usr/local/bin/python3" ]; then
-    brew install python3
+    # Brewed Python 3.
+    brew upgrade python
+  elif [ "$PYTHON" == "/usr/local/bin/python2" ]; then
+    # Brewed Python 2.
+    brew install python@2
   fi
+  $PYTHON -m pip install virtualenv
+  $PYTHON -m virtualenv -p $PYTHON venv
   brew install enchant
-  sudo pip install virtualenv
-  virtualenv -p $PYTHON venv
   source venv/bin/activate
 fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
       env: PYTHON=/usr/bin/python3.4 CATKIN_VERSION=indigo-devel
     - os: osx
       python: 2
-      env: PYTHON=/usr/bin/python CATKIN_VERSION=indigo-devel
+      env: PYTHON=/usr/local/bin/python2 CATKIN_VERSION=indigo-devel
     - os: osx
       python: 3
       env: PYTHON=/usr/local/bin/python3 CATKIN_VERSION=indigo-devel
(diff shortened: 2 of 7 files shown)


catkin_tools/common.py
```diff
@@ -640,7 +640,7 @@ def find_enclosing_package(search_start_path=None, ws_path=None, warnings=None,
         # Find packages under the search path
         try:
             pkgs = find_packages(search_path, warnings=warnings)
-        except:
+        except RuntimeError:
```
mikepurvis: This is based on inspection and the doc comments on the function in question:

http://docs.ros.org/independent/api/catkin_pkg/_modules/catkin_pkg/packages.html#find_packages

dirk-thomas:
Do these cases really want to capture something like `KeyboardInterrupt` or only subclasses of `Exception`?

mikepurvis:
@dirk-thomas That's an interesting question— in the case of subprocess (CommandStage), I believe it would never actually see a SIGINT unless sent one from the main process. I don't see any code directly in catkin_tools which does that, but I don't completely have a handle on how the asyncio side of things works. In any case, it certainly seems that when I Ctrl-C a running catkin_tools build, in-progress Make invocations do get signalled, and there would be little value in capturing stack traces from those into the build log.

For FunctionStage, things are a little murkier, since those are executing in-process (and most are trivial things like "make a directory"). I suppose it boils down to the hypothetical case where a FunctionStage is stuck and you Ctrl-C the whole job. Maybe you would actually want a stack trace in that scenario?

.travis.before_install.bash
```diff
@@ -4,10 +4,14 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
   sudo apt-get install enchant -y
 elif [ "$TRAVIS_OS_NAME" == "osx" ]; then
   if [ "$PYTHON" == "/usr/local/bin/python3" ]; then
-    brew install python3
+    # Brewed Python 3.
+    brew upgrade python
```
mikepurvis: Python 3 is now the default Python on homebrew so `python3` aliases to `python`, but you have to `brew upgrade` to actually get 3 if you already have 2, which is what's in the current Travis VM images.

.travis.yml
```diff
@@ -16,7 +16,7 @@ matrix:
       env: PYTHON=/usr/bin/python3.4 CATKIN_VERSION=indigo-devel
     - os: osx
       python: 2
-      env: PYTHON=/usr/bin/python CATKIN_VERSION=indigo-devel
+      env: PYTHON=/usr/local/bin/python2 CATKIN_VERSION=indigo-devel
```
mikepurvis: I gave up trying to get any of this working with system Python on Travis— it was easiest to just focus on brewed python2 and python3, and explicitly reference the exact binaries until we're safely in a virtualenv.

mikepurvis:
Will wait for a verdict from @wjwwood on the following:

- Whether we want to switch to `except Exception:` as suggested by Dirk, or save functional modifications for a future change.
- Whether switching the OS X CI to cover brewed `python2` and brewed `python3` is acceptable.

wjwwood:
> Whether we want to switch to except Exception: as suggested by Dirk, or save functional modifications for a future change.

Sounds like capturing ctrl-c tracebacks in Function might be useful, so I'd error on the side of putting them in the build log. If someone complains about it being misleading or too verbose we can revaluate it.

> Whether switching the OS X CI to cover brewed python2 and brewed python3 is acceptable.

That's fine by me.

PR #513, opened 2018-04-07
add opt --active to profile list

artivis:
Add `--active` option to `profile list` in order to print only the active profile with or without formatting (`--unformatted`).  

```sh
$ catkin profile list --active 
- default (active)
$ catkin profile list --active -u
default
```

### DIFF
--- a/catkin_tools/verbs/catkin_profile/cli.py
+++ b/catkin_tools/verbs/catkin_profile/cli.py
@@ -46,7 +46,7 @@ def prepare_arguments(parser):
         help='Print profile list without punctuation and additional details.')
 
     add = parser_list.add_argument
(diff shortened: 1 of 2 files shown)


NikolausDemmel:
LGTM.

I have a slight preference to not use the shortcut `-a` here. This doens't seem to me something for regular use from the command line but rather from scripting.

artivis:
Hi, thanks for the feedback.  
Indeed my first use case is scripting, shall I remove the `-a` option ??

wjwwood:
@artivis please do remove the `-a` flag. I think this is good to go otherwise.

artivis:
Shortcut removed, doc updated.

PR #527, opened 2018-11-13
Fix notification typo

sdeleersnijder:
When a build is successful you get a notification that says "succeded" instead of "succeeded".

image


mikepurvis:
Looks great, thanks! CI failures appear to be unrelated.

Issue #500, opened 2018-01-12
Catkin tool build issues with symlinking


tiusty:
### System Info

* Operating System: Linux alexDev-VirtualBox 4.13.0-26-generic #29~16.04.2-Ubuntu SMP Tue Jan 9 22:00:44 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 2.7.12
* Version of catkin_tools: catkin_tools 0.4.4 (C) 2014-2018 Open Source Robotics Foundation
catkin_tools is released under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
---
Using Python 2.7.12 (default, Nov 20 2017, 18:23:56) [GCC 5.4.0 20160609]
* ROS Distro: kinetic

### Build / Run Issue

* [x] Works with `catkin_make`
* [x] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [x] Works with `catkin build -v`
* [x] Works with `VERBOSE=1 catkin build -p1 -v`

### Expected Behavior

The expected behavior is for all the packages to build successfully and create a tools directory that contains all the executables that we built. 

### Actual Behavior

When the build is run with either:
catkin build or catkin build -p1 the build doesn't work 
but when the build is run with 
catkin build -v then the build successfully builds or others as noted above.

Therefore my main question is, what is the difference between catkin build and catkin build -v where the verbose version of the build works (can build the software successfully) but the non-verbose version doesn't. 

Some background on the error. 
1. We are building 4 executables
2. The 4 source files and the make files are copied into the build space
3. Then the CMakeLists.txt will run 'add_custom_target' to executable the makefiles pointing to the source in the build directory.
4. The Makefiles will then build the executables to the devel directory.

With the working version: The executables are built in the devel/.private directory and then afterwards it is symlinked to the merged devel space devel/.

With the broken version (catkin build or catkin build -p1): The executables are built in devel/.private directory but fails to symlink them to the merged devel workspace. 
Instead it throws an error:
No handlers could be found for logger "trollius"ed] [fly_tools:make (50%) - 2.3] [shared_interfaces:cmake - 2.2]

And then later it gives the traceback: 
[build] An internal error occurred!                                                                                                                                                                        
Traceback (most recent call last):
[...]

PR #529, opened 2018-11-27
Dirty fix for symlinking issue

floriantschopp:
I encountered the same issue as in #500 on melodic.
For sure my fix is far from being optimal as it does not actually solve the problem but it may help some of you to still build your packages.


mikepurvis:
Given that this project is on life support, my inclination is to merge this, despite being a hack. Can you confirm that this eliminates the issue described in the original ticket for you?

floriantschopp:
I can confirm that this eliminates the issue for me.

mikepurvis:
Great, thanks!

PR #533, opened 2019-01-23
fix issue when empty install prefix specified

### LINKED ISSUE #532: broken pip --user install
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

### System Info

* Operating System: `uname -a`
`Linux 4.15.0-43-generic #46-Ubuntu SMP Thu Dec 6 14:45:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
* Python Version: `python --version`
`Python 2.7.15rc1`
* Version of catkin_tools: `catkin --version` OR *Git revision*
`Not relevant`
* ROS Distro: `echo $ROS_DISTRO`
`N/A (using outside of ros)`

### Build / Run Issue

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [ ] Works with `catkin_make`
* [ ] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [ ] I did  not `read this`
* [x] Not applicable

### Expected Behavior

[//]: # (Please describe what you expected or think should happen.)
I should be able to do a user install of catkin_tools.

### Actual Behavior

[//]: # (Please describe what actually happens.)
It fails to install

[//]: # (Please put long CLI output into Gists: http://gist.github.com)
[//]: # (Use `VERBOSE=1 catkin build -p1 -v` for verbose output)
[//]: # (Please post `logs` directory if available.)

```
Installing collected packages: catkin-tools
  Running setup.py install for catkin-tools ... error
    Complete output from command /usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-cE992M/catkin-tools/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-record-8xEZNz/install-r
ecord.txt --single-version-externally-managed --compile --user --prefix=:
    usage: -c [--user | --prefix PREFIX]
    -c: error: argument --prefix: not allowed with argument --user/--home

    ----------------------------------------
Command "/usr/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-install-cE992M/catkin-tools/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-record-8xEZNz/install-record.txt --single-versi
on-externally-managed --compile --user --prefix=" failed with error code 2 in /tmp/pip-install-cE992M/catkin-tools/
```

### Steps to Reproduce the Issue

[//]: # (Please give a series of command-line commands which cause the problem.)
[//]: # (Include a checkout URI for the workspace so that we can reproduce the issue.)
[//]: # (If we can't reproduce it on our own machines, we can't fix it!)

```bash
pip install --user --no-cache-dir catkin_tools==0.4.4
```

ibtaylor:
resolves #532 

This patch allows for `pip --user` installs. `pip --user` runs `setup.py` with `--user` and `--prefix=` but `setup.py` currently requires only one be set. This patch allows `prefix` to be unspecified or an empty string when `--user` is specified.

see similar issue https://github.com/vcstools/wstool/issues/126

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,13 +98,20 @@ To enable tab completion, add the following to your '~/.bashrc':
                         'catkin_tools-completion.bash')))
 
 parser = argparse.ArgumentParser(add_help=False)
-prefix_group = parser.add_mutually_exclusive_group()
-prefix_group.add_argument('--user', '--home', action='store_true')
-prefix_group.add_argument('--prefix', default=None)
+parser.add_argument('--user', '--home', action='store_true')
+parser.add_argument('--prefix', default=None)
 
 opts, _ = parser.parse_known_args(sys.argv)
-userbase = site.getuserbase() if opts.user else None
-prefix = userbase or opts.prefix or sys.prefix
+if opts.user and not (opts.prefix == None or opts.prefix == ""):
+    raise Exception("error: argument --prefix: must be unspecified or empty if given with argument --user/--home")
+
+prefix = None
+if opts.user:
+    prefix = site.getuserbase()
+elif opts.prefix != None:
+    prefix = opts.prefix
+else:
+    prefix = sys.prefix
 
 setup(
     name='catkin_tools',


mikepurvis:
Looks reasonable, thanks for the fix to `pip --user install`.

PR #536, opened 2019-02-21
Fix bug when workspace contains special characters

### LINKED ISSUE #521: filepath with @ and . in it causes catkin config to fail
### System Info

* Operating System: Redhat Enterprise Linux (please don't panic/stop reading, i'm 99.9% confident my issue has nothing to do with this)
* Python Version: `Python 2.7.5`
* Version of catkin_tools:   catkin --version
catkin_tools 0.4.4 (C) 2014-2018 Open Source Robotics Foundation
catkin_tools is released under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
Using Python 2.7.5 (default, May 31 2018, 09:41:32) [GCC 4.8.5 20150623 (Red Hat 4.8.5-28)]

* ROS Distro: Kinetic

### Build / Run Issue

* [x] Works with `catkin_make`
* [x] Works with `catkin_make_isolated --merge`
* [] Works with `catkin build`
* [] Works with `catkin build -p1`
* [] I did  not `read this`

### Expected Behavior
Project builds.

### Actual Behavior
Project does not build. Indeed, it does not even configure:

```bash
catkin config --init
```
results in the following:

<an unspeakably long stack trace>...
  File "/homead/<my_user_name>.local/lib/python2.7/site-packages/catkin_tools/terminal_color.py", line 135, in fmt
    return t.substitute(_ansi) + (ansi('reset') if reset else '')
  File "/usr/lib64/python2.7/string.py", line 172, in substitute
    return self.pattern.sub(convert, self.template)
  File "/usr/lib64/python2.7/string.py", line 162, in convert
    val = mapping[named]
KeyError: 'se'

For company policy compliance reasons, my actual username on this system has been redacted and replaced with <my_user_name>, what's important is that has the following format:

somecharacters @ se.somemoreletters.somemoreletters.boeing.com

but without the spaces.

### Steps to Reproduce the Issue

create a catkin workspace (with some trivially compilable project in it) under a directory like my home directory on my Red Hat system:

/path/to/somewhere @ se.test.directory.com/

again, without the spaces.

Attempt to build the catkin project with:
```bash
catkin config --init
catkin build
```
It's worth noting that catkin_make is totally fine with this.

It's also worth noting that if I run this in a system directory (namely /opt/ros/kinetic), catkin tools works totally fine.  This is how I built ROS from source to run on Red Hat.

Please note that the obvious solution of "don't use a path with @ or . in it" will not work for me.  Boeing's standard Red Hat installer assigns home directories this way and is largely unchangeable in the short run.

Thanks!
Trent

haudren:
Simply sanitize the workspace name when printing it out

Would fix #521

### DIFF
--- a/catkin_tools/verbs/catkin_config/cli.py
+++ b/catkin_tools/verbs/catkin_config/cli.py
@@ -21,7 +21,7 @@ from catkin_tools.argument_parsing import add_context_args
 
 from catkin_tools.context import Context
 
-from catkin_tools.terminal_color import ColorMapper
+from catkin_tools.terminal_color import ColorMapper, sanitize
 
 color_mapper = ColorMapper()
 clr = color_mapper.clr
@@ -141,7 +141,7 @@ def main(opts):
         summary_notes = []
 
         if not context.initialized() and do_init:
-            summary_notes.append(clr('@!@{cf}Initialized new catkin workspace in `%s`@|' % context.workspace))
+            summary_notes.append(clr('@!@{cf}Initialized new catkin workspace in `%s`@|' % sanitize(context.workspace)))
 
         if context.initialized() or do_init:
             Context.save(context)


mikepurvis:
Looks like a great fix, thanks!

PR #539, opened 2019-03-19
add local_setup.* to DEVEL_LINK_BLACKLIST

### LINKED ISSUE #538: issues sym-linking catkin's new local_setup.sh
I get the following error since `local_setup.sh` was recently introduced in https://github.com/ros/catkin/pull/993
```
Cannot symlink from /homes/rhaschke/src/ros/devel/.private/rviz/local_setup.sh to existing file /homes/rhaschke/src/ros/devel/./local_setup.sh
Warning: Source hash: 1b5f4fec545a06a4cc4723db5116e3f2
Warning: Dest hash: 51a457f5bad6a3827efef461047f77e8
```
As pointed out in https://github.com/ros/catkin/issues/994#issuecomment-470201871, this looks like catkin-tools needs some adaptation too.

I see the same issues in Travis, building MoveIt, but interestingly [only in this build job](https://travis-ci.org/ros-planning/moveit/jobs/502614343#L652) using clang as a compiler but not in [this one]https://travis-ci.org/ros-planning/moveit/jobs/502614339#L543) using gcc.

rhaschke:
Attempt to resolve #538, adding `local_setup.*` to `DEVEL_LINK_BLACKLIST` as suggested in https://github.com/catkin/catkin_tools/issues/538#issuecomment-470315837.
This resolves the warnings but also doesn't install `local_setup.*` shell files.
As I don't know the purpose of adding those files to catkin, I cannot judge whether this is fine or not.

### DIFF
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -639,6 +639,9 @@ DEVEL_LINK_BLACKLIST = DEVEL_LINK_PREBUILD_BLACKLIST + [
     'setup.bash',
     'setup.zsh',
     'setup.sh',
+    'local_setup.bash',
+    'local_setup.zsh',
+    'local_setup.sh',
     '_setup_util.py',
 ]
 


mikepurvis:
This looks reasonable, thanks!

PR #541, opened 2019-03-26
Fix travis failures on Ubuntu

mikepurvis:
Travis was failing on Ubuntu on account of not being able to install the `enchant` package. Seems the base images were configured with apt having only `${dist}-security` plus a ton of PPAs, and no default repos. This is baffling to me, but also easy to correct at build time.

I've also moved the builds from Trusty to Xenial, given that Trusty is EOL next month anyway.

### DIFF
--- a/.travis.before_install.bash
+++ b/.travis.before_install.bash
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 
 if [ "$TRAVIS_OS_NAME" == "linux" ]; then
-  sudo apt-get install enchant -y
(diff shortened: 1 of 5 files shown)


mikepurvis:
Going to disable the OS X builds, as they are failing for infrastructure reasons (too many open files).

mikepurvis:
@wjwwood Does anything here offend you?

My intent is to (personally) run the current head from source for a few days, and then propose that 0.4.5 be released.

wjwwood:
When you're ready to release, send me a direct email please, and I will help you.

PR #548, opened 2019-04-15
add authors/maintainers/licenses to context, which are the default values on package creation

furushchev:
This PR adds `--authors`, `--maintainers` and `--licenses` options to set the default values on creating new packages.

By using this PR, we can use short command line arguments to create new packages:

BEFORE:

`catkin create pkg awesome_pkg -a 'John Doe' 'foo@bar.com' -a ... -m 'John Doe' 'foo@bar.com' -m ... -l MIT -c <deps>`

AFTER:

`catkin create pkg awesome_pkg -c <deps>`

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -66,6 +66,9 @@ class Context(object):
         'catkin_make_args',
         'whitelist',
         'blacklist',
+        'authors',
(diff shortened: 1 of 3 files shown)


mikepurvis:
Works great, thanks!

Issue #545, opened 2019-04-09
`-x` space_suffix not working anymore in latest version.  


daenny:
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

Setting a `space_suffix` is broken in the latest version. When adding a `space_suffix` to the the catkin profile, nothing is changed in the configuration.   

### System Info

* Operating System: `Linux f717aae6a030 4.15.0-47-generic #50~16.04.1-Ubuntu SMP Fri Mar 15 16:06:21 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux` (ros-kinetic-base docker-image)
* Python Version: `Ptython 2.7.12`
* Version of catkin_tools: `catkin --version` OR *Git revision*
```
catkin_tools 0.4.5 (C) 2014-2019 Open Source Robotics Foundation
catkin_tools is released under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
```
* ROS Distro: `echo $ROS_DISTRO` melodic

### Build / Run Issue

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [x] Not applicable
* [ ] Works with `catkin_make`
* [ ] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [ ] I did  not `read this`

### Expected Behavior

[//]: # (Please describe what you expected or think should happen.)
Adding a space_suffix should add the suffix to all standard spaces. 
```
catkin config -x _debug
-----------------------------------------------------------------
Profile:                     default
Extending:             [env] /opt/ros/kinetic
Workspace:                   /root/ros/kinetic/test
-----------------------------------------------------------------
Build Space:       [missing] /root/ros/kinetic/test/build_debug
Devel Space:       [missing] /root/ros/kinetic/test/devel_debug
[...]

PR #560, opened 2019-07-19
Fix space-suffix option

Matthew-Reynolds:
Addresses https://github.com/catkin/catkin_tools/issues/545

From the docs, the `catkin config --space-suffix` argument should set the "Suffix for build, devel, and install space if they are not otherwise explicitly set." However, currently, the default values for each of these spaces are considered "explicitly set". This means that, except for an extremely narrow window when first initializing a workspace, the `--space-suffix` argument is always useless.

This PR allows `--space-suffix` to affect spaces whose values are either not set **or are set to the default values**. This brings the tool back in line with how it used to behave.

------
Aside: The documentation says that this affects the build, devel, and install spaces, but the code is implemented such that it affects the build, devel, install, and log spaces. I assume the code is correct and the documentation is out of date, but I'd like to get confirmation on which is prefered so that I can update the code/documentation accordingly.

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -306,9 +306,10 @@ class Context(object):
         for space, space_dict in Context.SPACES.items():
             key_name = space + '_space'
             default = space_dict['default']
-            if space_suffix and space != 'source':
-                default += space_suffix
-            setattr(self, key_name, kwargs.pop(key_name, default))
+            value = kwargs.pop(key_name, default)
+            if value == default and space_suffix and space != 'source':
+                value += space_suffix
+            setattr(self, key_name, value)
 
         # Check for unhandled context options
         if len(kwargs) > 0:


Matthew-Reynolds:
Some more details:

Whenever a context is saved, it writes all of its space values to the profile's `config.yaml` file. This includes all space values, even if the space values are default. Therefore, the only time when the `--space-suffix` tool can have any effect is immediately after initializing a workspace before the context is saved.

Importantly, contexts are saved in a lot of different cases (whenever `catkin profile add`, `catkin profile rename`, `catkin build --save-config`, or any `catkin config` commands are run), meaning the `--space-suffix` tool will almost immediately be useless once you start using the workspace. Even just running `catkin config` to read back the config summary or adding a new profile will save the context and write the default values to the config file, rendering the `--space-suffix` argument useless.

mikepurvis:
Thanks for figuring this out and addressing it.

Issue #559, opened 2019-07-19
Sphinx 2.0.0 doesn't support Py2.7 or Py3.4


Matthew-Reynolds:
The Travis script currently attempts to `pip install sphinxcontrib-spelling`. However, as of version 4.3.0, `sphinxcontrib-spelling` requires `Sphinx>=2.0.0`, which dropped support for both py2.7 and 3.4 (See the Sphinx changelog for [v2.0.0b1](http://www.sphinx-doc.org/en/master/changes.html#release-2-0-0-released-mar-29-2019)). This means Travis will always fail immediately since it cannot install `sphinxcontrib-spelling`.

Matthew-Reynolds:
Closed in #561

PR #561, opened 2019-07-20
Force sphinxcontrib-spelling < 4.3.0

Matthew-Reynolds:
Addresses #559.

`sphinxcontrib-spelling` 4.3.0 requires at least `sphinx` 2.0.0, which in turn requires at least Py3.5. Both Py2.7 and 3.4 are no longer supported as of `sphinx` 2.0.0, which causes the Travis build to fail during before_install.


mikepurvis:
Thanks very much for sorting this out!

PR #562, opened 2019-08-20
Bash completion for run_tests verb

ct2034:
I thought it is useful to be able to autocomplete the very handy `run_tests` verb


mikepurvis:
Looks reasonable, thanks!

PR #563, opened 2019-08-20
Fix permissions of created setup files

meyerj:
`catkin_tools` should respect the default permissions assigned by the OS to new files, similar to `chmod +X <file>` at the command line.

The default permissions can be influenced by the active [umask](https://en.wikipedia.org/wiki/Umask). For example, on a shared file system (network mount or a workspace mounted in a docker container) the group of the user should have write access, too (by setting the umask to `0002` or `0000` instead of the default `0022`). Because each build command tries to overwrite the setup files, building the same workspace as a different user fails because of permission errors.

The `setup.*` files created by `build.py` for an isolated devel-space do not need to be executable. The default permissions should be fine, no?

[catkin](https://github.com/ros/catkin) also creates its setup and other files by forcefully overwriting the permissions. I will open another PR there.

### DIFF
--- a/catkin_tools/jobs/cmake.py
+++ b/catkin_tools/jobs/cmake.py
@@ -149,9 +149,14 @@ def generate_env_file(logger, event_queue, context, install_target):
     os.write(tmp_dst_handle, data.encode('utf-8'))
     os.close(tmp_dst_handle)
 
+    # Make the file executable without overwriting the permissions for
+    # the group and others (copy r flags to x)
+    mode = os.stat(tmp_dst_path).st_mode
+    mode |= (mode & 0o444) >> 2
+    os.chmod(tmp_dst_path, mode)
+
     # Do an atomic rename with os.rename
     os.rename(tmp_dst_path, env_file_path)
-    os.chmod(env_file_path, 0o755)
 
     return 0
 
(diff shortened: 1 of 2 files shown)


meyerj:
Unfortunately the outlined approach of sharing workspaces with changing user ids will not work without better support in CMake: https://cmake.org/pipermail/cmake-developers/2015-August/026032.html

`catkin` creates a lot of files using CMake, mostly with [`configure_file()`](https://cmake.org/cmake/help/latest/command/configure_file.html). Unfortunately those will not respect the umask either, but copy the permissions from the source templates. Typically the template files are owned by root and installed in `/opt/ros/...` or `/usr` with default permissions, so no chance to fix it for all cases without extensive CMake hacking...

Still, leaving the PR open, because I think it is a small step in the right direction.

mikepurvis:
@meyerj Should this be merged in its current state, or is it best to hold off?

It seems like a reasonable improvement as written, and I don't see a lot of potential for it to blow up.

meyerj:
> @meyerj Should this be merged in its current state, or is it best to hold off?
> 
> It seems like a reasonable improvement as written, and I don't see a lot of potential for it to blow up.

Agreed. Just not very useful on its own. Maybe a future CMake version will provide an easy way to use `configure_file()` with permissions or to set the executable bits on a file without modifying the other bits set by the OS.

The most significant change is that the generated `setup.*` files for isolated devel-spaces will not be executable anymore, but they never have if created through other means.

mikepurvis:
Sounds good.

PR #564, opened 2019-08-22
Update release suites.

nuclearsandwich:
Cosmic, Disco, and Eoan are now live on the ROS bootstrap repository and future releases should be pushed to them as well.

The second commit removes suites older than Ubuntu Xenial and Debian Jessie.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -3,5 +3,5 @@ Depends: python-argparse, python-setuptools, python-catkin-pkg (> 0.2.9), python
 Depends3: python3-setuptools, python3-catkin-pkg (> 0.2.9), python3-yaml
 Conflicts: python3-catkin-tools
 Conflicts3: python-catkin-tools
-Suite: trusty vivid wily xenial wheezy yakkety zesty artful bionic cosmic disco eoan jessie stretch buster
+Suite: xenial yakkety zesty artful bionic cosmic disco eoan jessie stretch buster
 X-Python3-Version: >= 3.2
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -3,5 +3,5 @@ Depends: python-argparse, python-setuptools, python-catkin-pkg (> 0.2.9), python
 Depends3: python3-setuptools, python3-catkin-pkg (> 0.2.9), python3-yaml
 Conflicts: python3-catkin-tools
 Conflicts3: python-catkin-tools
-Suite: xenial yakkety zesty artful bionic cosmic disco eoan stretch buster
+Suite: xenial yakkety zesty artful bionic cosmic disco eoan focal stretch buster
 X-Python3-Version: >= 3.5
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -3,5 +3,5 @@ Depends: python-argparse, python-setuptools, python-catkin-pkg (> 0.2.9), python
 Depends3: python3-setuptools, python3-catkin-pkg (> 0.2.9), python3-yaml
 Conflicts: python3-catkin-tools
 Conflicts3: python-catkin-tools
-Suite: xenial yakkety zesty artful bionic cosmic disco eoan jessie stretch buster
+Suite: xenial yakkety zesty artful bionic cosmic disco eoan stretch buster
 X-Python3-Version: >= 3.5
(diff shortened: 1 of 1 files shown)


mikepurvis:
Yakkety, Zesty, Artful, and Cosmic are all EOL as well, per: https://wiki.ubuntu.com/Releases

But I definitely agree about adding disco and eoan.

@wjwwood What's the policy on supporting EOL'd releases in these cases?

nuclearsandwich:
I opened the PR in bulk as I was spinning up the ROS bootstrap repositories which I had to do for all new distributions regardless of release status. If there is no demand for it I'd much prefer not releasing into distros that are already unsupported at the time of release.

mikepurvis:
@wjwwood Can we get `osrf-pycommon` released to `disco` and `eoan`?

wjwwood:
Testing in `eoan` I think this should be fixed already. Sorry for the delay in getting back to you. I think @sloretz fixed this when preparing for Noetic and focal (20.04).

wjwwood:
@mikepurvis I think this could just be merged.

mikepurvis:
Sounds good thanks. I'll push one extra commit bumping the supported Python version to 3.5, as that's the Xenial baseline.

dirk-thomas:
Jessie which is still listen in the suites comes with Python 3.4 afaik.

mikepurvis:
Huh. I'm not sure what my commitment to pre-3.5 testing and validation, particularly in service of supporting oldoldstable. Perhaps `jessie` should just be dropped as well?

dirk-thomas:
> Perhaps `jessie` should just be dropped as well?

Kinetic is still an actively supported distro and targets Jessie (see https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021).

mikepurvis:
Jessie users would still have `python-catkin-tools` v0.4.5, and could also pip install other versions. And catkin_tools has never really been an officially supported or recommended method of building ROS in the wiki tutorials (and has always had a per-1.0 version number).

I don't feel strongly one way or another on this. I can bump back the Python version number to 3.4, here and in the travis testing config, but there's unlikely to be much in terms of actual support or attention to issues with it.

What is the OSRF opinion on this?

dirk-thomas:
> What is the OSRF opinion on this?

I don't have an opinion regarding `catkin_tools`. I leave that up to @wjwwood.

(I would rather like to see the effort spent on adding what is missing to `colcon` than continue keeping `catkin_tools` on life support. But that is just my personal opinion.)

mikepurvis:
Colcon doesn't support Xenial or Jessie either (https://github.com/colcon/colcon-core/issues/138#issuecomment-549105854), which is one of the barriers for us pursuing a migration to it. :)

dirk-thomas:
> Colcon doesn't support Xenial or Jessie either

`colcon` does certainly support Xenial (see https://github.com/colcon/colcon-core/blob/5b849001429cb138f3768c7fa93a8b30dd67dd25/stdeb.cfg#L4). Since it requires Python 3.5 (for all the `asyncio` stuff) it can't support Jessie though (assuming you don't want to install a newer Python version).

wjwwood:
I don't have an opinion on what to do one way or the other. I doubt there are many people using `catkin_tools` on Debian Jessie at this point, but you could simply not release new versions of it for Jessie right? (i.e. remove jessie from the supported suites at the same time as raising the require Python version?)

I really don't see what `colcon` has to do with this at all.

mikepurvis:
Ah. The current master branch of catkin tools also uses `asyncio`. So in that case we can't support Python 3.4 regardless. So yeah, I'll drop `jessie` as a target.

wjwwood:
> Ah. The current master branch of catkin tools also uses asyncio. So in that case we can't support Python 3.4 regardless. So yeah, I'll drop jessie as a target.

Not that it changes what you should do imo, but `asyncio` was added in Python 3.3. I was stabilized in Python 3.5, but it did exist and `catkin_tools` supported Python 3.3 at some point.

mikepurvis:
@wjwwood The work by @jbohren to switch catkin_tools over to asyncio/trollius happened for the 0.4.0 release, in 2016. Not saying that the current source wouldn't still work on Python 3.3 or 3.4, but it's very much possible that the one that did back then was the Popen-based version.

wjwwood:
AFAIR, `catkin_tools` always used `asyncio` for subprocess communication. That's why I wrote `osrf_pycommon` and that's why they depend on trollius, in order to support Python2 and older versions of Python3.

Again, doesn't really matter, I just wanted to point out that you can use `asyncio` before Python 3.5, because it was characterized that you could not, which is simply not true.

mikepurvis:
Cool, thanks for the clarification. I'll merge this later today if there are no objections.

PR #565, opened 2019-08-28
fix(controllers): python3 fix for the controllers.py file

tdejager:
Fixes for controllers.py file, should be python2 compatible


mikepurvis:
Great, thanks!

Issue #406, opened 2016-10-26
catkin profile commands fail with non-existent extend directory


jonfink:
### System Info
- Operating System: `Darwin desiderius 15.6.0 Darwin Kernel Version 15.6.0: Mon Aug 29 20:21:34 PDT 2016; root:xnu-3248.60.11~1/RELEASE_X86_64 x86_64`
- Python Version: `Python 2.7.10`
- Version of catkin_tools: `catkin_tools 0.4.2`
- ROS Distro: `jade`
### Expected Behavior

In a workspace where several catkin profiles have been defined to work with different ROS distributions (i.e., a _jade_ profile extends `/opt/ros/jade` while a _kinetic_ profile extends `/opt/ros/kinetic`) but some of those distributions are not installed, one should still be able to freely list and set available catkin profiles.

For example (on a system with `/opt/ros/jade` but not '/opt/ros/kinetic'):

``` bash
$ catkin profile list
[profile] Available profiles:
- default (active)
- jade
- kinetic
```

``` bash
$ catkin profile set kinetic
[profile] Activated catkin metadata profile: kinetic
[profile] Available profiles:
- default 
- jade
- kinetic (active)
```

``` bash
$ catkin profile set jade
[profile] Activated catkin metadata profile: kinetic
[profile] Available profiles:
- default 
- jade (active)
- kinetic
```
### Actual Behavior

Again, on a system with `/opt/ros/jade` but not '/opt/ros/kinetic':

``` bash
$ catkin profile list
[profile] Available profiles:
- default (active)
- jade
- kinetic
```

_Note, default profile extends nothing_

``` bash
$ catkin profile set kinetic
[profile] Activated catkin metadata profile: kinetic
[profile] Available profiles:
- default 
- jade
- kinetic (active)
```

``` bash
$ catkin profile set jade
Traceback (most recent call last):
  File "/usr/local/bin/catkin", line 9, in <module>
    load_entry_point('catkin-tools==0.4.2', 'console_scripts', 'catkin')()
  File "/Library/Python/2.7/site-packages/catkin_tools/commands/catkin.py", line 258, in main
    catkin_main(sysargs)
  File "/Library/Python/2.7/site-packages/catkin_tools/commands/catkin.py", line 253, in catkin_main
    sys.exit(args.main(args) or 0)
  File "/Library/Python/2.7/site-packages/catkin_tools/verbs/catkin_profile/cli.py", line 106, in main
    ctx = Context.load(opts.workspace)
[...]

PR #566, opened 2019-09-02
Allow to switch profiles even if the current profile is misconfigured

AlexisTM:
This PR follows #406 

**Current behavior:**
If the current profile is misconfigured, we cannot switch to any profile.

**After the PR:**
If the current profile is misconfigured, we **can** switch to any profile.

I do not know if there are adverse problems due to not loading the environment but that makes sense for the `profile` verb not to crash due to profile problems.


mikepurvis:
Thanks!

PR #568, opened 2019-09-16
Handle empty file in `get_metadata` correctly

Mofef:
If the `metadata_file_path` exists but the `metadata_file` is empty, `yaml.safe_load(metadata_file)` will return None. But the caller expects a dictionary which can lead to a crash. (e.g. in https://github.com/catkin/catkin_tools/blob/master/catkin_tools/verbs/catkin_build/cli.py#L371)

I'm not sure how I ended up with this file being empty (I can't remember touching it)...


mikepurvis:
Fine with this change for now, though perhaps a better approach if others hit this would be to catch yaml errors and emit a warning.

PR #569, opened 2019-09-24
respect setup.cfg during install

LeroyR:
Installation fails if prefix is set with:
`setup.py setopt -c "install" -o "prefix" -s "PREFIX"`
as the prefix for the bash_completion is only read from command line.

This fixes that by using distutils

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -98,21 +98,12 @@ To enable tab completion, add the following to your '~/.bashrc':
                         'etc/bash_completion.d',
                         'catkin_tools-completion.bash')))
 
-parser = argparse.ArgumentParser(add_help=False)
-parser.add_argument('--user', '--home', action='store_true')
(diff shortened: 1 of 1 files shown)


mikepurvis:
This seems like a reasonable change, but I don't know know enough about Python packaging to really feel confident that I understand the implications— in any case, I'm pretty sure the old slogans about "`setuptools` good, `distutils` bad" don't apply in this scenario.

@wjwwood, do you mind weighing in here?

wjwwood:
> @wjwwood, do you mind weighing in here?

I'm afraid I cannot say if this makes sense or not. I don't understand the use case or the actual technical problem.

I'd say if it method that was failing now works and the related, "normal" method still works, then it's ok.

mikepurvis:
Okay, merging. If there are howls of rage, it can be reverted.

Issue #571, opened 2019-10-18
Commit 12872fd7d573e1 disabled ability to pass directories to `list` verb 


jarvisschultz:
### System Info

* Operating System: `4.15.0-65-generic #74~16.04.1-Ubuntu SMP Wed Sep 18 09:51:44 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
* Python Version: `Python 2.7.12`
* Version of catkin_tools: `catkin_tools 0.4.5`
* ROS Distro: `kinetic`

### Build / Run Issue

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [ ] Works with `catkin_make`
* [ ] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [ ] I did  not `read this`

### Expected Behavior

Numerous  examples in the documentation reference the ability to get a list of packages in a directory (e.g. `catkin list -u path/to/directory`). For example, the [cheat sheet has several examples](https://catkin-tools.readthedocs.io/en/latest/cheat_sheet.html#building-packages), and so does the [header for the list verb](https://catkin-tools.readthedocs.io/en/latest/verbs/catkin_list.html). This functionality was dropped in [this commit](https://github.com/catkin/catkin_tools/commit/12872fd7d573e138a061e3d7b342ac395dab5ec8) that added the ability to run `catkin list --this`. This seems like quite useful functionality. Why was it dropped? If we wanted to add it back should we look to a PR, a custom verb, something else?

mikepurvis:
@jarvisschultz Did you ever do any work on updating the sphinx docs per https://github.com/catkin/catkin_tools/pull/572#issuecomment-560523277?

jarvisschultz:
Finally found a few minutes to deal with this. See PR #604

PR #572, opened 2019-10-18
Added option to past list of directories to `list` verb

jarvisschultz:
Potential simple fix to address #571


mikepurvis:
Looks reasonable. I'm assuming this going in means we also need to update the documentation referenced in the original ticket?

jarvisschultz:
Yes, documentation needs to be updated. Regardless of if this PR is accepted, there are a few instances where the documentation utilizes the `list -u` functionality that no longer works. If this is accepted, docs should reflect new feature, otherwise, remove references.

Would it be helpful if I updated the docs on this PR?

mikepurvis:
I'm fine to merge it now, and you can update the docs in a followup, particularly since the docs are _already_ broken anyway— this change is not making anything worse.

Issue #555, opened 2019-06-18
Blacklisting a package that contains "jn" in its name gets "jn" added to make args


haudren:
### System Info

* Operating System: Ubuntu Bionic Beaver (18.04) x86_64
* Python Version: Python 2.7.15+
* Version of catkin_tools: 0.4.5
* ROS Distro: melodic

### Expected Behavior

Configuring a workspace by blacklisting a package, whatever its name, should not affect make-args. 

### Actual Behavior

Configuring a workspace by blacklisting a package starting by jn adds it to the make args, which potentially slows down the build.

### Steps to Reproduce the Issue

```bash
mkdir -p test_ws/src
cd test_ws/src
catkin_create_pkg j2pkg
cd ..
catkin build
catkin config --blacklist j2pkg
```

Observe the result:
```
-------------------------------------------------
Additional CMake Args:       None
Additional Make Args:        j2
Additional catkin Make Args: None
Internal Make Job Server:    True
Cache Job Environments:      False
-------------------------------------------------
```

gavanderhoorn:
Does `--blacklist 'j2pkg'` change how the args are interpreted?

haudren:
Hey sorry for the late reply: no it doesn't. Using both `'j2pkg'` and `"j2pkg"` results in the same behaviour.

jarvisschultz:
Probably a closely-related issue is trying to build a single package or list of packages that contain `jn` in their names. The `jn` is automatically consumed by the `make-args`, and stripped from the package list:

```
catkin build j2example_msgs
-----------------------------------------------------------
Profile:                     default
Extending:        [explicit] /opt/ros/kinetic
Workspace:                   /home/jarvis/examplews
-----------------------------------------------------------
Build Space:        [exists] /home/jarvis/examplews/build
Devel Space:        [exists] /home/jarvis/examplews/devel
Install Space:      [unused] /home/jarvis/examplews/install
Log Space:          [exists] /home/jarvis/examplews/logs
Source Space:       [exists] /home/jarvis/examplews/src
DESTDIR:            [unused] None
-----------------------------------------------------------
Devel Space Layout:          linked
Install Space Layout:        None
-----------------------------------------------------------
Additional CMake Args:       None
Additional Make Args:        j2
Additional catkin Make Args: None
Internal Make Job Server:    True
Cache Job Environments:      False
-----------------------------------------------------------
Whitelisted Packages:        None
Blacklisted Packages:        None
-----------------------------------------------------------
[...]

Issue #516, opened 2018-05-21
catkin build argument parsing error on leading "l" 


astumpf:
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

### System Info

* Operating System: Linux ThinkPad-T440p 4.13.0-41-generic #46~16.04.1-Ubuntu SMP Thu May 3 10:06:43 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 2.7.12
* Version of catkin_tools: catkin_tools 0.4.4 (C) 2014-2018 Open Source Robotics Foundation
catkin_tools is released under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
---
Using Python 2.7.12 (default, Dec  4 2017, 14:50:18) [GCC 5.4.0 20160609]
* ROS Distro: Kinetic Kame

### Build / Run Issue

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [ ] Works with `catkin_make`
* [ ] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`

### Expected Behavior

[//]: # (Please describe what you expected or think should happen.)
catkin build should complete with a regular build process.

### Actual Behavior

[//]: # (Please describe what actually happens.)

[//]: # (Please put long CLI output into Gists: http://gist.github.com)
[//]: # (Use `VERBOSE=1 catkin build -p1 -v` for verbose output)
[//]: # (Please post `logs` directory if available.)
Argument parsing went wrong. Instead of taking the full name package name it parses a preceding "l" directly followed by a number as an argument leading to a wrong package name. 

### Steps to Reproduce the Issue

[//]: # (Please give a series of command-line commands which cause the problem.)
[...]

PR #573, opened 2019-10-21
Updated regex for extracting '-l' and '-j' args for make

jarvisschultz:
Addresses #555 

Summary of improvements:
  - No longer consuming args that happen to begin `jX` or `lX` (where `X` is a
  sequence if integers.
  - Fixed bug where an empty 'j' arg would throw a `ValueError` when trying to
  convert an empty string (`''`) to an int
  - Both `-l/--load-average` and `-j/--jobs` can now be empty (which is
  supported by GNU Make
  - The `-l/--load-average` arg will now accept a float value (e.g. `catkin
  build -l4.0`) which is supported by GNU Make

### DIFF
--- a/catkin_tools/argument_parsing.py
+++ b/catkin_tools/argument_parsing.py
@@ -226,9 +226,9 @@ def extract_jobs_flags_values(mflags):
     :rtype: dict
     """
 
-    regex = r'(?:^|\s)(?:-(j|l)(\s*\d+(?:\.\d*)?|\s|$))' + \
+    regex = r'(?:^|\s)+?(?:-)(j|l)(?:=|(?:\s+?)|$)?(\d*(?:\.\d*)?)?(?!-)' + \
             r'|' + \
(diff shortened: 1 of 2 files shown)


timonegk:
This also addresses #516.

mikepurvis:
This seems like a good improvement, but I'll be honest those regexes kind of terrify me. Would you be willing to contribute along with this a small test that demonstrates expected inputs/outputs?

jarvisschultz:
They terrify me too! They aren't that different from what was already there, so I figured just modifying the regex (instead of a complete re-write) would be less challenging to get accepted.

Happy to contribute a few tests. I'll update the PR in a few days.

jarvisschultz:
It took over a month instead of a few days, but I finally pushed a few tests. Wouldn't you know it, the unit tests revealed a few additional minor flaws in the regular expressions that were in use. The current version seems to pass all corner cases that I could think of.

mikepurvis:
Thank you so much for this contribution and your patience in awaiting a merge.

PR #582, opened 2019-11-14
Revert #561

Matthew-Reynolds:
Now that the project is only targetting Python 3.5+, #561 is no longer required and we can use the most up-to-date version of `sphinx` and `sphinxcontrib-spelling`.


mikepurvis:
Excellent, thanks!

PR #587, opened 2019-11-27
fix line breaks in CMakeIOBufferProtocol.colorize_cmake

ipa-mdl:
Just stumbled upon this.
This is the only line, which uses `\r\n`, everywhere else `\ņ` is used.
(with the exception of some `\r` in https://github.com/catkin/catkin_tools/blob/master/catkin_tools/execution/controllers.py)

Did this slip through in https://github.com/catkin/catkin_tools/pull/276/commits/cc842c8775ec8a7a0a13f3bc8a280ea0fa373c14 or was it intended?

### DIFF
--- a/catkin_tools/jobs/commands/cmake.py
+++ b/catkin_tools/jobs/commands/cmake.py
@@ -150,7 +150,7 @@ class CMakeIOBufferProtocol(IOBufferProtocol):
                         cline = cline.format(*match.groups())
                     break
 
-        return cline + '\r\n'
+        return cline + '\n'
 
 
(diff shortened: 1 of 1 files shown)


mikepurvis:
@ipa-mdl This seems like it should go in?

ipa-mdl:
> This seems like it should go in?

I am not sure.
The old version kept the original line ending.
The new version appends `\r\n` unconditionally.

For now, I just can tell that our GitHub Actions logs contain many empty lines: https://github.com/ros-industrial/industrial_ci/runs/359655683#step:3:167

timonegk:
Replaced by #666.

PR #590, opened 2019-12-30
Also run tests with Python 3.7 and 3.8

sloretz:
Debian Buster uses Python 3.7 and Ubuntu Focal will use Python 3.8


mikepurvis:
Looks good, thanks.

PR #592, opened 2020-01-13
Allow extending profiles and merging their contents.

janstrohbeck:
This allows profiles to contain the following:
```yaml
extends: <profile name>
```
The `Context` class then loads that base profile and merges the contents of the profiles (last value for a given key wins). When saving the context, the changes are saved in the correct files by keeping a lookup table which stores the file that each parameter was loaded from.

The use case I have for this is that we have many users, each with a separate profile. Most users just want to set a custom `whitelist` or `blacklist` and use default values from another profile for the `cmake_args` etc. With this PR, a lot of copied and pasted YAML code can be removed.

I do not know if this is something that is wanted in this project. If yes, I can add unit tests for the added functionality or refactor the changes if there is a nicer way to implement it (the `key_origins` feels a bit hacky).

Edit: `extends` might not be a good name as it is similar to `extend_path`. Also, I couldn't get the unit tests to work on my machine, so the CI build might fail.

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -207,7 +207,8 @@ class Context(object):
                     if base_profile is None:
                         return config_metadata
                     if base_profile in visited_profiles:
-                        raise IOError("Profile dependency circle detected at dependency from '%s' to '%s'!" % (profile, base_profile))
+                        raise IOError("Profile dependency circle detected at dependency from '%s' to '%s'!"
+                                      % (profile, base_profile))
                     base = get_metadata_recursive(base_profile)
                     base.update(config_metadata)
                     config_metadata = base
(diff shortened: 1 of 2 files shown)


mikepurvis:
This looks like a reasonable and useful contribution. I don't think I have much quibble with the name— `extends`, `parent`, `base`, they're all basically the same thing. We do use "extend" to refer especially to the business of _extending a workspace_, but that's pretty separate from config profiles.

I'm happy to see tests come in a follow-on. Is there anything you want to add to this before it goes in?

janstrohbeck:
We've used it at my organization for a couple of weeks and have not had any issues with it yet. It works even with 4 levels of `extends` :-)

So I think it can be merged as is. I probably won't have time to add tests in the next few weeks.

mikepurvis:
Awesome, thanks for the contribution.

PR #593, opened 2020-01-21
Add an option for catkin build to skip the installation step

meyerj:
With this pull request I propose to add a new command line option for the `build` verb, that allows to skip the installation step temporarily, only for the current invocation:

```
  --skip-install        Skips the installation step. Useful in combination
                        with --make-args or for running unit tests in a pre-
                        installed workspace.
```
(renamed from `--no-install`, see https://github.com/catkin/catkin_tools/pull/593#issuecomment-595981685)

Other than `catkin config --no-install`, this flag does not modify the workspace configuration.

It is useful in combination with certain `--make-args` and `--catkin-make-args` targets, first of all `--catkin-make-args run_tests` or `--make-args test` to execute unit tests. Typically you do not want to call `make install` afterwards in this case and **only** run the tests. I also added `--skip-install` to the predefined aliases `catkin test` and `catkin run_tests`.

Obviously, if packages depend on each other, the workspace has to be built once before without `--skip-install` (maybe this is not the case with `--merge-devel`?). But if that has been done before, adding `--skip-install` in a subsequent `run_tests` run significantly reduces the build (or test) time. This is especially true if only a few packages actually have unit tests or if the reinstallation of one package triggers a rebuild of one or more downstream packages.

For one particular workspace with 158 packages the difference in catkin_tools/cmake/make overhead (*without actually running unit tests!*) is

```cpp
$ catkin build
...
[build] Summary: All 158 packages succeeded!                                                                                                                                                                                        
[build]   Ignored:   2 packages were skipped or are blacklisted.                                                                                                                                                                    
[build]   Warnings:  1 packages succeeded with warnings.                                                                                                                                                                            
[build]   Abandoned: None.                                                                                                                                                                                                          
[build]   Failed:    None.                                                                                                                                                                                                          
[build] Runtime: 2 minutes and 45.2 seconds total.

$ catkin build --skip-install
...
[build] Summary: All 158 packages succeeded!                                                                                                                                                                                        
[build]   Ignored:   2 packages were skipped or are blacklisted.                                                                                                                                                                    
[build]   Warnings:  None.                                                                                                                                                                                                          
[build]   Abandoned: None.                                                                                                                                                                                                          
[build]   Failed:    None.                                                                                                                                                                                                          
[build] Runtime: 39.0 seconds total. 
```

In both cases the workspace was already built and installed successfully before.


catkin_tools/jobs/cmake.py
```diff
@@ -286,8 +286,7 @@ def create_cmake_build_job(context, package, package_path, dependencies, force_c
 
     # Pre-clean command
     if pre_clean:
-        make_args = handle_make_arguments(
-            context.make_args + context.catkin_make_args)
```
meyerj: Adding `context.catkin_make_args` here for `cmake` type jobs like in [catkin.py:464](https://github.com/catkin/catkin_tools/blob/613771f330c5b6ed86f3fec89f6e3214e21558e5/catkin_tools/jobs/catkin.py#L464-L467) was probably a bug.

But in which cases would it be useful to add more arguments to `make clean` at all? At least this breaks use cases like
```sh
catkin build --pre-clean --make-args tests
```

mikepurvis:
I understand the motivation for this functionality, but it feels like a tough one to explain to a newcomer, especially the subtle difference between `catkin config --no-install` and `catkin build --no-install`.

I think my criteria for accepting this would be a) changing the name to something like `--skip-install` to clearly differentiate it, and b) putting up at least something in the way of guardrails to prevent accidental misuse. For example, keeping track of past builds and ensuring that `--skip-install` is only used when all packages in question have already been built and installed at least once.

At the end of the day, though, my true preference for how to fix the underlying issue here would be to build a real, first class `catkin test` verb which does the right thing in terms of knowing that installation is not a prerequisite to testing, etc. In the meantime, I generally resolve the issue here like so:

```
catkin build
cd build/my_fancy_package
make run_tests
```

meyerj:
> I think my criteria for accepting this would be a) changing the name to something like `--skip-install` to clearly differentiate it, and b) putting up at least something in the way of guardrails to prevent accidental misuse. For example, keeping track of past builds and ensuring that `--skip-install` is only used when all packages in question have already been built and installed at least once.

Understood. I addressed a) in https://github.com/catkin/catkin_tools/pull/593/commits/9d8f66de56cc2e4f13fc94835eb06d0fd3db53fd, but currently do not have time to look into b). I can imagine that it is not straight-forward, because even if all packages have been built before there is no guarantee that the installed version is the latest version and that `make run_tests` is never executed against an older version of a dependency installed before.

I wonder whether the current behavior is actually fail-safe and consistent:

- As far as I know, the CMake target `test` does not implicitly depend on the targets that actually builds the test executables it will run. `make test` only runs all commands that have been registered through [ctest](https://cmake.org/cmake/help/latest/manual/ctest.1.html). In general, any custom command can be registered as a test using the [add_test()](https://cmake.org/cmake/help/latest/command/add_test.html) command in cmake.
  - So for running `catkin test` (aliased to `build --verbose [--skip-install] --make-args test --`) all test targets need to be built before.
  - There is no automatic rebuild if a compiled unit test or a library it depends on has been updated after it has been built originally.

- The `run_tests` target created by catkin in [tests.cmake](https://github.com/ros/catkin/blob/86439ec5d2010d5c47c30b815aa97e525037930f/cmake/test/tests.cmake) has a dependency on target `tests`, which indirectly depends on all test targets that have been added with one of the [catkin test macros](http://docs.ros.org/api/catkin/html/howto/format2/gtest_configuration.html), e.g. [catkin_add_gtest()](https://github.com/ros/catkin/blob/noetic-devel/cmake/test/gtest.cmake#L84). But other targets in that package, for example executables or libraries not involved in any unit test will not be built by `make run_tests`.
  - In a pure devel-space setup (`--no-install`) this could lead to the same problem as with `make test` again, namely that downstream packages fail to build because libraries are missing. Or missing executables could break downstream unit tests because a rostest depends on them.
  - In an install-space setup (`--install`) at least each package gets installed after it has been tested, so there should be no difference for downstream packages. Some targets might only get built during the `make install` step though.

Except for the last case, which only works for catkin packages, there is no reliable way to run all unit tests other than
```sh
catkin build
catkin <test|run_tests>
```

without touching anything in between. In this case `--skip-install` as a default for the second step cannot do any harm in installed workspaces and is a no-op for devel-space workspaces.

Given that the analysis above is correct, maybe an option `--additional-make-args` would be useful, such that
```sh
catkin build --additional-make-args <test|run_tests>
```
executes the standard `make`, `make test` and optional `make install` sequence for each package, and
```sh
catkin build --make-args <test|run_tests> --skip-install
```
could be used to only run the unit tests in a pre-built workspace, without an additional `install` step.

I am not sure whether `make install` always implies `make all`, e.g. for general, non-catkin cmake packages. If not, then `make install` should never be called without a previous `make` call that actually builds all targets to be installed, which is not guaranteed by something like
```sh
catkin build --this --no-deps --make-args single_library_target
```

If installation is enabled, then `catkin build` always runs `make single_library && make install`, and `--make-args` is pretty useless without a way to skip the install step. I would even suggest that passing `--make-args` should implicitly skip the install step. Then `--additional-make-args` and `--skip-install` would not be required either. You could still build and install in a single invocation by adding target `install` explicitly to `--make-args`, e.g. `--make-args all test install`.

meyerj:
I guess this is obsolete with the new `test` verb added in https://github.com/catkin/catkin_tools/pull/676, which also solves the problem of duplicate install steps with the previous `catkin build --make-args test` approach.

At least for me that was the only use case for `--skip-install` so far.

PR #595, opened 2020-01-29
Update executor.py to have proper asyncio support

ProfFan:
This is a PR to make catkin_tools support newer python (python 3).


mikepurvis:
Looks like no functional change here, just eliminating superfluous parentheses?

ProfFan:
Oh never mind, just realized it is already fixed in the master branch :) the release version for melodic is still lagging so I patched them...

Issue #543, opened 2019-04-02
Packages ignored. What package and why? 


roman-tub:
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

Hi guys,
this might be stupid, but I spent quite a while trying to find the information I'm looking for. 
Maybe this is a feature request or maybe I'm just not seeing the obvious.

### System Info

* Operating System: `Ubuntu 4.15.0-46-generic`
* Python Version: `Python 2.7.15rc1`
* Version of catkin_tools: `catkin_tools 0.4.4`
* ROS Distro: `melodic`

### Build / Run Issue

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [ ] Works with `catkin_make`
* [ ] Works with `catkin_make_isolated --merge`
* [x] Works with `catkin build`
* [x] Works with `catkin build -p1`
* [ ] I did  not `read this`

### Expected Behavior

[//]: # (Please describe what you expected or think should happen.)

I just want to know which package was ignored and why. 

### Actual Behavior

[//]: # (Please describe what actually happens.)

[//]: # (Please put long CLI output into Gists: http://gist.github.com)
[//]: # (Use `VERBOSE=1 catkin build -p1 -v` for verbose output)
[//]: # (Please post `logs` directory if available.)

### Steps to Reproduce the Issue

[//]: # (Please give a series of command-line commands which cause the problem.)
[//]: # (Include a checkout URI for the workspace so that we can reproduce the issue.)
[//]: # (If we can't reproduce it on our own machines, we can't fix it!)

Whitelist and blacklist are None.

```bash
VERBOSE=1 catkin build -p1 -v | grep -i ignor
```

gives: 
```bash
[build] Ignored: 1 packages were skipped or are blacklisted. 
```

mikepurvis:
Show the packages in your workspace:

```
catkin list
```

[...]

Issue #418, opened 2016-12-29
Generating metapackage artifacts with catkin_tools


meyerj:
### System Info

* Operating System: `Linux im-laptop-002 4.4.0-57-generic #78~14.04.1-Ubuntu SMP Sat Dec 10 00:14:47 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`
* Python Version: `Python 2.7.6`
* Version of catkin_tools: `catkin_tools 0.4.2 (C) 2014-2016 Open Source Robotics Foundation`
* ROS Distro: `indigo`

### Build / Run Issue

This issue was already reported on ROS answers: http://answers.ros.org/question/241107/generating-metapackage-artifacts-with-catkin_tools/

catkin_tools is not able to build metapackages. In particular, I tried with [orocos_toolchain](https://github.com/orocos-toolchain/orocos_toolchain/tree/toolchain-2.9/orocos_toolchain) and with both, install mode enabled and disabled.

[//]: # (If you are migrating from catkin_make, please follow the migration guide:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/migration.html)

[//]: # (Please also check for solved issues here:)
[//]: # (http://catkin-tools.readthedocs.org/en/latest/troubleshooting.html)

[//]: # (And check for open issues here:)
[//]: # (https://github.com/catkin/catkin_tools/issues?q=is%3Aopen+is%3Aissue+label%3Abug)

[//]: # (If you have an issue building and running code, please verify the following:)

* [x] Works with `catkin_make`
* [x] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [ ] I did  not `read this`

### Expected Behavior

[//]: # (Please describe what you expected or think should happen.)
`catkin build` should build (and install) all packages and metapackages in the workspace.
`catkin build orocos_toolchain --no-deps` should build (and install) the metapackage only because it has no build dependencies. Actually I am not sure whether `--no-deps` relates to build or run-time dependencies. I think it should consider all kind of dependencies, but in any case none of them should be built with `--no-deps`.

### Actual Behavior

[//]: # (Please describe what actually happens.)
`catkin build` ignores the metapackage:
```bash
$ VERBOSE=1 catkin build -p1 -j2 --summary
--------------------------------------------------------------
Profile:                     default
Extending:        [explicit] /opt/ros/indigo
Workspace:                   /home/johannes/orocos_ws
[...]

Issue #366, opened 2016-04-30
Clarify metapackage dependency behavior.


jbohren:
Currently, if a user's package depends on a metapackage (such as `fanuc`), `catkin build` displays an error similar to the following:

```
Abandoned <<< fanuc_motion             [ Depends on unknown jobs: fanuc ]
```

This is in accordance with [REP-140](http://www.ros.org/reps/rep-0140.html#metapackage), which says the following with respect to metapackages:

> They cannot be used for catkin builds and compile nothing themselves. 

Except apparently `catkin_make_isolated` builds metapackages and packages which are linked only by `<exec_depend>` dependencies.

So we should either allow or disallow depending on metapackages for building. If we allow it, we need to patch the `build` verb, and if we continue to disallow it, we need to make this error message clearer so people know it's because they're depending on metapackages.

Related discussions: #364 #224 #148

jack-oquin:
Does ``catkin_make`` allow ``<exec_depend>`` on metapackages, too?

If so, maybe REP-0140 should be amended to allow that behavior, and ``catkin build`` should support it, too. It does make intuitive sense to allow such dependencies. I believe it was ruled out only due to difficulties with building.

gavanderhoorn:
>  It does make intuitive sense to allow such dependencies. I believe it was ruled out only due to difficulties with building.

I can imagine it was perhaps also disallowed to discourage ppl from using transitive dependency declarations. If the 'contents' of a metapackage changes (ie: it's `run_depend`s or `exec_depend`s are changed), dependants could suddenly find themselves in a situation where there dependencies are no longer present, even though they didn't change anything.

jbohren:
On looking at this a little closer, it looks like `catkin_make` and `catkin_make_isolated` both *allow* `<exec_depend>` on metapackages, but do not actually use them to define the build order, they just ignore them. The fact that this has been working for people is just lucky (or they've actually been building against other versions installed to their systems).

Since there's nothing in `catkin_make` or `catkin_make_isolated` preventing people from doing this, I propose we include metapackages in dependency resolution so that it does what the user is _trying_ to do, but warn the user loudly if we do.

@wjwwood What do you think about that?

gavanderhoorn:
[...]

PR #596, opened 2020-02-03
Do not ignore metapackages

meyerj:
Simple fix for #418. Another related issue is https://github.com/catkin/catkin_tools/issues/543.

Metapackages must not be ignored by catkin_tools. The fact that other non-metapackages cannot depend on them according to [REP-140](http://www.ros.org/reps/rep-0140.html#metapackage) does not imply that they do not need to be built and installed at all.

The question on which packages are selected if a list of package names is given at the command line (or as a whitelist) is a different story in my opinion. This is more what issue https://github.com/catkin/catkin_tools/issues/366 is about, but not directly connected to this pull request. I will open a separate PR for that.

### DIFF
--- a/catkin_tools/verbs/catkin_build/build.py
+++ b/catkin_tools/verbs/catkin_build/build.py
@@ -167,9 +167,8 @@ def get_built_unbuilt_packages(context, workspace_packages):
     # Get names of all unbuilt packages
     unbuilt_pkgs = set()
     for path, pkg in workspace_packages.items():
-        if 'metapackage' not in [e.tagname for e in pkg.exports]:
-            if pkg.name not in built_packages:
-                unbuilt_pkgs.add(pkg.name)
+        if pkg.name not in built_packages:
+            unbuilt_pkgs.add(pkg.name)
 
     return built_packages, unbuilt_pkgs
 
@@ -480,11 +479,7 @@ def build_isolated_workspace(
         if pkg.name not in packages_to_be_built_names:
             continue
 
-        # Ignore metapackages
-        if 'metapackage' in [e.tagname for e in pkg.exports]:
-            continue
-
-        # Get actual execution deps
+        # Get actual build deps
         deps = [
             p.name for _, p
             in get_cached_recursive_build_depends_in_workspace(pkg, packages_to_be_built)


gavanderhoorn:
Can you give a use-case for building and installing them?

The contents of metapackages is strictly regulated, their purpose is fixed to allowing group-installs of other packages, other non-metapackages are not allowed to depend on them and when installed, only a `package.xml` is placed in `share/$pkg_name`.

Is it the latter that you are interested in?

meyerj:
> Can you give a use-case for building and installing them?
> 
> The contents of metapackages is strictly regulated, their purpose is fixed to allowing group-installs of other packages, other non-metapackages are not allowed to depend on them and when installed, only a `package.xml` is placed in `share/$pkg_name`.

> Is it the latter that you are interested in?

Yes, that is the primary intention. Because only metapackages are allowed to depend on metapackages it has no consequences if `catkin build` ignores them in most cases, but there is also no good reason to do so and it is also not an implication of REP-140. As far as I know, `catkin_make` and `catkin_make_isolated` do not treat metapackages in any special way, both for finding recursive dependencies nor for actually building them. The only difference is that they create a default `CMakeLists.txt` for metapackages from a template if the package does not provide one (https://github.com/ros/catkin/pull/349).

One valid scenario, where the current behavior of ignoring metapackages is broken, is with two workspaces, where the second one has a metapackage that depends on a metapackage in the first one:
```
workspace_a
\-- src
    |-- package_a1
    |-- package_a2
    \-- metapackage_a    (exec_depend on package_a1 and package_a2)

workspace_b    (extends workspace_a/devel or workspace_a/install)
\-- src
   |-- package_b1
   \-- metapackage_b   (exec_depend on package_b1 and metapackage_a)
```

There are some existing examples of metapackages depending on other metapackages, e.g. [desktop/package.xml](https://github.com/ros/metapackages/blob/20c6fe68aaca88813fa63821d3224865f28c9a10/desktop/package.xml).

meyerj:
> The question on which packages are selected if a list of package names is given at the command line (or as a whitelist) is a different story in my opinion. This is more what issue #366 is about, but not directly connected to this pull request. I will open a separate PR for that.

I found that catkin_tools does already consider all types of dependencies when selecting packages to be built, which is what I would expect. Only the name of functions [get_cached_recursive_build_depends_in_workspace()](https://github.com/catkin/catkin_tools/blob/af9db1ed8057eca9089a5347afdad8ae471b31ca/catkin_tools/common.py#L166) and [get_recursive_build_depends_in_workspace()](https://github.com/catkin/catkin_tools/blob/af9db1ed8057eca9089a5347afdad8ae471b31ca/catkin_tools/common.py#L264) is confusing, because they actually return all recursive `build_depends`, `buildtool_depends`, `test_depends` and `run_depends` since https://github.com/catkin/catkin_tools/commit/6bc286af0e41a7f9bce79e981413bed72c712f1c.

mikepurvis:
Looks great, thanks for the contribution.

PR #600, opened 2020-03-06
Revert "jobs: Fixing environment required to run catkin test targets on pre-indigo catkin"

meyerj:
This pull requests suggest to revert #265 (_[PR196+] jobs: Fixing environment required to run catkin test targets on pre-indigo catkin_), which is supposed to be obsolete. According to the description and inline comments it was only required for ROS Hydro and earlier.

If complete removal is too disruptive, I suggest to only install the env-hook in ROS Hydro and earlier (not sure what would be a reliable way to detect this, before any workspace is sourced).

Rationale:
- The `CATKIN_TEST_RESULTS_DIR` is not used anymore by catkin since https://github.com/ros/catkin/pull/728. Not sure about `ROS_TEST_RESULTS_DIR`, I did not check.
- The additional env-hook get installed unconditionally and "pollutes" the devel- and install-spaces, although normally it should not do any harm.
- In some cases it does harm, e.g. when building Debians using catkin_tools. catkin itself has flags like `CATKIN_BUILD_BINARY_PACKAGE` that prevents installation of common files in a shared install-space. `06-ctr-nuke.sh` did not respect this and required manual removal from the staged `DESTDIR` to avoid that the file is added to multiple binary packages (see also related issue https://github.com/catkin/catkin_tools/issues/280).

### DIFF
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -345,18 +345,6 @@ def link_devel_products(
     return 0
 
 
-def ctr_nuke(logger, event_queue, prefix):
-    """Adds an env-hook which clears the catkin and ros test results dir."""
-
-    ctr_nuke_path = os.path.join(prefix, 'etc', 'catkin', 'profile.d')
-    ctr_nuke_filename = os.path.join(ctr_nuke_path, '06-ctr-nuke.sh')
-    mkdir_p(ctr_nuke_path)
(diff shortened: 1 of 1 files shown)


mikepurvis:
This looks reasonable; there's no reason to be supporting pre-Indigo distros at this point, heck the master branch doesn't even work with Python 2 any more. :)

PR #601, opened 2020-03-20
Fix get_python_install_dir issues

### LINKED ISSUE #597: get_python_install_dir issues
This was noticed by @tspicer01 as part of Clearpath's Python 3 migration effort. The crux of the issue is that catkin's version of this functionality ([implemented in CMake](https://github.com/ros/catkin/blob/noetic-devel/cmake/python.cmake)) is based on checking `${PYTHON_VERSION_MAJOR}` as populated by the `PythonInterp` CMake module, which I believe actually _calls_ `PYTHON_EXECUTABLE` to check what it is.

In contrast, the version in catkin_tools is based on checking `sys.version_info`, that is, it assumes that the Python that's invoking catkin_tools will be the same one as `PYTHON_EXECUTABLE`. This is definitely _not_ a valid assumption, particularly after catkin_tools 0.5.0 is released (#594 ) and we may have users running Python 3 catkin_tools to build Python 2 ROS Melodic workspaces.

The ideal would be to replace this logic with just shelling out to a `cmake -P` invocation that just includes and calls the function from `catkin`. The difficulty with that is that we don't know for certain that `catkin` is in the workspace or even in an underlay— catkin_tools needs this functionality even in the scenario where it's just building a bunch of CMake packages with no `catkin` in sight. So it seems like the right answer then is to extract `PYTHON_EXECUTABLE` (if present) from the context.cmake_args, and also invoke it ourselves to determine what version it is (falling back to `env python`), and otherwise continuing to duplicate the logic from `catkin`.

And because shelling out to a new `python` process is rather more expensive than just looking up `sys.version_info`, the result after the change should be returned from a cache after the first call.

timonegk:
This pull request attempts to close #597 by shelling out to check `cmake -P`, as suggested in the issue. The executed script, however, does not depend on catkin and should therefore work fine for workspaces without catkin. 
While it might have been possible to replicate the behavior of cmake manually, I decided against it because this solution is easier to maintain and less error-prone, especially since the `find_program` cmake builtin would have been rather annoying to reimplement.

I marked this pull request as a draft because
* there is still a todo in the code that I do not really know how to implement. I think it would be right to abort the build process when conflicting cmake variables have been set (e.g. python version > version of python executable). 
* no caching of the result has been implemented yet (I am not even sure if it is needed because the result of `get_python_install_dir` is only used to write `setup.sh` once).

In the second commit I fixed the install path for python3 on debian systems to reflect catkin's behavior.


mikepurvis:
Hi @timonegk— sorry for the long delay, but thanks for taking a stab at this! I've been busy but would like to get back to trying to get this release out and am just taking a look at your work here.

Regarding the business of conflicting variables, I think if we were able to re-implement less logic here this wouldn't matter as much. That is, if we simply vendored-in (and clearly marked, in order to keep it in sync) a copy of [python.cmake](https://github.com/ros/catkin/blob/noetic-devel/cmake/python.cmake), and called it from a wrapper that printed out whatever `PYTHON_INSTALL_DIR` it lands on. Then it wouldn't be up to us to determine conflicts because even if they exist, we'd still be consistent with what was being done by catkin (it uses its version of that logic to determine where to install generated python message headers, for example).

Regarding caching, yeah it can be left off for now. It's not important in the base case of catkin/cmake packages. Internally at Clearpath, we call the function each time we build a plain Python package (see https://github.com/mikepurvis/catkin_tools_python/blob/master/catkin_tools_python/job.py#L215), but if that becomes a concern for us, we can handle adding the caching or changing that logic.

I think the litmus test for getting this in is probably to add a test that does the following:

- Clone ros/catkin (which is a plain cmake package and therefore uses this logic)
- Build the workspace
- Source the workspace and check that PYTHONPATH is pointing to the correct Python.

This should be run 4x, once each for the workspace being Python 2 or Python 3 (catkin_tools itself is now always Python 3), and once each for the develspace and installspace, per this logic here:

https://github.com/catkin/catkin_tools/blob/74e9d2b3c2db5b460cd84e9bdcb3f946c6eea2ef/catkin_tools/jobs/cmake.py#L169-L175

timonegk:
Thank you for the feedback. I added the tests that you proposed. Regarding the `PYTHON_INSTALL_DIR`, you are probably right and it makes more sense not to mix cmake and python code as I did, I will update the pull request accordingly.

timonegk:
I do not really understand why the CI is failing. It seems to be related to https://github.com/ros/catkin/issues/863 and therefore seems to be a problem of the CI itself. I also cannot reproduce the problem in a bionic container. If you don't know how to fix the issue, I would suggest to remove the install tests (since the same `get_python_install_dir` function is used for devel and install anyway).

.travis.yml
```diff
@@ -33,15 +33,18 @@ before_script:
   # Install catkin_tools test harness dependencies
   - ./.travis.before_script.bash
   - pip install empy sphinx_rtd_theme nose coverage flake8 mock --upgrade
+  - python2 -m pip install --user catkin-pkg
```
mikepurvis, line 36: Is this necessary? Pretty sure we should always just call `python` and let the PATH set by travis steer that to the right place.

timonegk: Building catkin with python 2 requires the catkin_pkg package for python 2. This is not related to the python version for catkin_tools which is managed by travis

timonegk:
Are you sure that setting the `PYTHON` environment variable actually changes the used python version? The travis log suggests otherwise, e.g. in [the current build](https://travis-ci.org/github/catkin/catkin_tools/jobs/690219684) in line 597, the travis virtualenv is used.

tests/system/verbs/catkin_build/test_pythonpath.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+
+def test_python2_devel():
+    with workspace_factory() as wf:
+        os.mkdir(os.path.join(wf.workspace, 'src'))
+        subprocess.call(['git', 'clone', '-q', 'https://github.com/ros/catkin', 'src/catkin'])
+        assert catkin_success(['build', '-DPYTHON_VERSION=2'])
+        pythonpaths = subprocess.check_output(
+            ['bash', '-c', 'source ' + wf.workspace + '/devel/setup.bash && echo $PYTHONPATH']).decode().split(':')
+        ws_pythonpath = [p for p in pythonpaths if p.startswith(wf.workspace)][0]
```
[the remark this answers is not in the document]
timonegk: I do not think so because we have to check if the PYTHONPATH points to the correct directory within the workspace.

timonegk:
Well, I just realized that the new `get_python_install_dir` function is never called in the tests. I adapted the tests to use a CMake package from the resources folder and made sure that the function is actually called this time. For that it was necessary to set the workspace layout to something other than "linked" to avoid using `catkin_tools_prebuild` which uses catkin's own logic to generate the setup files. (When building the `catkin` package as we did before, catkin generates the setup files too.)

catkin_tools/jobs/cmake.py
```diff
@@ -88,15 +88,17 @@ def get_python_install_dir():
...
-    python_install_dir = 'lib'
-    if os.name != 'nt':
-        python_version_xdoty = str(sys.version_info[0]) + '.' + str(sys.version_info[1])
-        python_install_dir = os.path.join(python_install_dir, 'python' + python_version_xdoty)
-
-    python_use_debian_layout = os.path.exists('/etc/debian_version')
-    python_packages_dir = 'dist-packages' if python_use_debian_layout else 'site-packages'
-    python_install_dir = os.path.join(python_install_dir, python_packages_dir)
-    return python_install_dir
+    cmake_command = ['cmake']
```
mikepurvis, line 91: Please use `CMAKE_EXEC` here.

mikepurvis:
This looks terrific. Really appreciate this contribution and your patience in iterating on it with me.

PR #615, opened 2020-06-09
Fix installation of new cmake files

timonegk:
Follow-up on #601: The cmake files were not installed, the fix only worked when installing in development mode. This PR also updates the tests to test on the properly installed, non-development, installation.


timonegk:
Hey, I just wanted to quickly ping this PR, it is a very small fix for the current master where determining the correct python version is broken because I forgot a change in `setup.py` in #601.

mikepurvis:
Looks good, thanks.

PR #622, opened 2020-07-13
Remove blacklist/whitelist Terms

allenh1:
Given the recent wave of renaming terms to be more inclusive, I thought it might be worth removing the terms "blacklist" and "whitelist".

This PR replaces the terms with "denylist" and "allowlist" respectively (one of the proposed replacements for these terms for Linux kernel), but I would be happy to change this patch to use other terms upon request.

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -64,7 +64,7 @@ class Context(object):
         'use_internal_make_jobserver',
         'use_env_cache',
         'catkin_make_args',
-        'whitelist',
+        'allowlist',
         'denylist',
         'authors',
         'maintainers',
@@ -293,7 +293,7 @@ class Context(object):
         use_env_cache=False,
         catkin_make_args=None,
         space_suffix=None,
-        whitelist=None,
+        allowlist=None,
         denylist=None,
         authors=None,
         maintainers=None,
@@ -340,8 +340,8 @@ class Context(object):
         :type catkin_make_args: list
         :param space_suffix: suffix for build, devel, and install spaces which are not explicitly set.
         :type space_suffix: str
-        :param whitelist: a list of packages to build by default
-        :type whitelist: list
+        :param allowlist: a list of packages to build by default
+        :type allowlist: list
         :param denylist: a list of packages to ignore by default
         :type denylist: list
         :raises: ValueError if workspace or source space does not exist
@@ -379,8 +379,8 @@ class Context(object):
 
         self.destdir = os.environ['DESTDIR'] if 'DESTDIR' in os.environ else None
 
-        # Handle package whitelist/denylist
-        self.whitelist = whitelist or []
+        # Handle package allowlist/denylist
+        self.allowlist = allowlist or []
         self.denylist = denylist or []
 
         # Handle default authors/maintainers
@@ -546,7 +546,7 @@ class Context(object):
                 clr("@{cf}Cache Job Environments:@|      @{yf}{_Context__use_env_cache}@|"),
             ],
             [
-                clr("@{cf}Whitelisted Packages:@|        @{yf}{whitelisted_packages}@|"),
+                clr("@{cf}Whitelisted Packages:@|        @{yf}{allowlisted_packages}@|"),
                 clr("@{cf}Blacklisted Packages:@|        @{yf}{denylisted_packages}@|"),
             ]
         ]
@@ -592,7 +592,7 @@ class Context(object):
             'devel_missing': existence_str(self.devel_space_abs),
             'install_missing': existence_str(self.install_space_abs, used=self.__install),
             'destdir_missing': existence_str(self.destdir, used=self.destdir),
-            'whitelisted_packages': ' '.join(self.__whitelist or ['None']),
+            'allowlisted_packages': ' '.join(self.__allowlist or ['None']),
             'denylisted_packages': ' '.join(self.__denylist or ['None']),
         }
         subs.update(**self.__dict__)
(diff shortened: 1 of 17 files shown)


gavanderhoorn:
This doesn't sound like a good change: the packages on the blacklist are not denied anything. They are *ignored* by the tool during topological ordering, which is something else.

allenh1:
> This doesn't sound like a good change: the packages on the blacklist are not denied anything.

Which term would you suggest instead? Maybe "skip list"?

wjwwood:
Perhaps `include_list` and `exclude_list`?

dirk-thomas:
In `colcon` there is a semantic different between *ignoring* packages (as if they didn't exist in the file system) and *skipping* packages (not processing them in the current invocation but considering there effect on e.g. the dependency graph). I don't know the behavior of the existing options in `catkin_tools` but if they match one of these two it might make sense to align the naming.

catkin_tools/context.py
```diff
@@ -546,8 +546,8 @@ def summary(self, notes=[]):
                 clr("@{cf}Cache Job Environments:@|      @{yf}{_Context__use_env_cache}@|"),
             ],
             [
-                clr("@{cf}Whitelisted Packages:@|        @{yf}{whitelisted_packages}@|"),
-                clr("@{cf}Blacklisted Packages:@|        @{yf}{blacklisted_packages}@|"),
+                clr("@{cf}Whitelisted Packages:@|        @{yf}{allowlisted_packages}@|"),
+                clr("@{cf}Blacklisted Packages:@|        @{yf}{denylisted_packages}@|"),
```
wjwwood, line 550: The labels would need to be updated as well.

catkin_tools/jobs/catkin.py
```diff
@@ -234,8 +234,8 @@ def link_devel_products(
     # List of files that collide
     files_that_collide = []
 
-    # Select the blacklist
-    blacklist = DEVEL_LINK_PREBUILD_BLACKLIST if prebuild else DEVEL_LINK_BLACKLIST
+    # Select the denylist
+    denylist = DEVEL_LINK_PREBUILD_BLACKLIST if prebuild else DEVEL_LINK_BLACKLIST
```
wjwwood, line 238: Missed one here too.

wjwwood:
These options are most closely related to the ignoring behavior in colcon. It dictates which packages are built or not by default when you don't specify anything else. It's part of the persistent `catkin config` command. Each invocation can override these though.

docs/verbs/catkin_config.rst
```diff
@@ -138,43 +138,43 @@ This can be done with the ``--extend`` option like so:
...
+    catkin config --no-allowlist
 
-If the *blacklist* is non-empty, it will filter the packages to be built in all cases except where a given package is named explicitly.
-This means that blacklisted packages will not be built even if another package in the workspace depends on them.
+If the *denylist* is non-empty, it will filter the packages to be built in all cases except where a given package is named explicitly.
+This means that denylisted packages will not be built even if another package in the workspace depends on them.
 
 .. note::
 
     Blacklisting a package does not remove it's build directory or build
```
wjwwood, line 162: And here.

allenh1:
@wjwwood thanks for the review, I'll catch those in my next pass.

@dirk-thomas / @wjwwood `ignore_list` certainly makes a lot of sense given this context, but I'm not sure what the opposite verb would be? `focus_list`? `build_list`?

Glad to hear other suggestions for that one, as I'm not quite sure (I don't really use the `whitelist` feature).

dirk-thomas:
> I'm not sure what the opposite verb would be? `focus_list`? `build_list`?

`colcon` uses *select* in that context.

mintar:
Also make sure to properly update the verbs. "blacklisted"/"whitelisted" are verbs, but "allowlisted/denylisted"? If you're going with "ignorelist" and "selectlist", use the proper verb forms like "ignored" and "selected".

@wjwwood / @dirk-thomas : Is this change going to be rolled out in released distros? Will there be some kind of backwards compatibility / deprecation warnings / user config migration?

allenh1:
> "blacklisted"/"whitelisted" are verbs, but "allowlisted/denylisted"? If you're going with "ignorelist" and "selectlist", use the proper verb forms like "ignored" and "selected".

@mintar Yeah, that's a good point. For context, I just looked up what Linux kernel is using to replace "blacklist/whitelist" and applied it in this patch, but it certainly isn't proper English and doesn't really sound all that great (or as @gavanderhoorn pointed out earlier make all that much sense).

```
catkin config --packages-select [x] [y] [z]
```

and

```
catkin config --packages-ignore
```

seem to be the most in-line with the way `colcon` is doing things, and the most sensible replacement to me. Sound good?

wjwwood:
> @wjwwood / @dirk-thomas : Is this change going to be rolled out in released distros? Will there be some kind of backwards compatibility / deprecation warnings / user config migration?

I don't know how we'd land the change just yet. I think it would be best to perhaps allow the old arguments to continue to work as aliases to the new arguments, but not document them and perhaps have a deprecation warning.

If we released it, it would likely only impact the Python3 version of catkin_tools I still need to release and so it would limited to some of the newer Ubuntu distributions and only if you use the python3 version in older distros (as it is available). It has no correlation with ROS distributions, since it is a pure Python tool.

NikolausDemmel:
> If we released it, it would likely only impact the Python3 version of catkin_tools I still need to release and so it would limited to some of the newer Ubuntu distributions and only if you use the python3 version in older distros (as it is available). It has no correlation with ROS distributions, since it is a pure Python tool.

Also would affect pip installs, but those a probably a smaller minority.

I agree some config migration / backwards compatibility / deprecation would be good.

wjwwood:
I looked at this again for a while this afternoon. I want to get it in a better state in terms of backwards compatibility before I push some new changes. I just wanted to let you know @allenh1 that I was making fixups and stuff so you didn't do the same.

allenh1:
@wjwwood thanks for letting me know, sorry for leaving this one get stale. I will have some time this weekend if you need me to do any modifications.

wjwwood:
It's ok, I'll let you know.

PR #623, opened 2020-07-14
Add `--this` option to `clean` verb

spurnvoj:


Rayman:
Thank you, I'm always typing this but it was not yet implemented

mikepurvis:
It would be great to have a test for this behaviour as well, please consider submitting one.

PR #624, opened 2020-07-15
fix 'catkin locate' for symlinked pkgs inside workspace

spurnvoj:


mikepurvis:
Look straightforward.

timonegk:
I think in line 180, `os.getcwd()` should be changed to `getcwd()` to correctly locate relative paths in symlinked packages.

spurnvoj:
Hi @timonegk, you are right. I also modified line 109  according to your pull request. Your is a nicer solution :)

Issue #625, opened 2020-07-29
[noetic] AttributeError: 'str' object has no attribute 'decode'


Rayman:
### System Info

* Operating System: Linux user-VirtualBox 5.4.0-37-generic #41-Ubuntu SMP Wed Jun 3 18:57:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
* Python Version: Python 3.8.2
* Version of catkin_tools: master 9a3cc3172b1f72e3e1d364d7dfd000abb5712e53
* ROS Distro: noetic

### Build / Run Issue
`catkin_tools` prints inbetween logs the commands it executes. I was trying to debug a build, copy pasted the line, but I got a stacktrace.

Original command:
```bash
cd /home/user/ros/common/build/somepackage; catkin build --get-env somepackage | catkin env -si  /usr/bin/make test --jobserver-auth=3,4; cd -
```
Minimum reproducable example
```bash
echo | catkin env -s echo test
```

### Expected Behavior
The minimum working example should print `test`.

### Actual Behavior
```
Traceback (most recent call last):
  File "/home/user/.local/bin/catkin", line 11, in <module>
    load_entry_point('catkin-tools', 'console_scripts', 'catkin')()
  File "/home/user/source/catkin_tools/catkin_tools/commands/catkin.py", line 272, in main
    catkin_main(sysargs)
  File "/home/user/source/catkin_tools/catkin_tools/commands/catkin.py", line 267, in catkin_main
    sys.exit(args.main(args) or 0)
  File "/home/user/source/catkin_tools/catkin_tools/verbs/catkin_env/cli.py", line 109, in main
    environ.update(parse_env_str(input_env_str))
  File "/home/user/source/catkin_tools/catkin_tools/common.py", line 675, in parse_env_str
    split_envs = [e.split('=', 1) for e in cmd_split(environ_str)]
  File "/home/user/source/catkin_tools/catkin_tools/common.py", line 31, in cmd_split
    return _cmd_split(s.decode('utf-8'))
AttributeError: 'str' object has no attribute 'decode'
```

### Steps to Reproduce the Issue
```bash
echo | catkin env -s echo test
```

timonegk:
Fixed in #627 (and follow-up in #649).

PR #627, opened 2020-08-15
Fix byte decoding for python 3 (Issue #625)

g-gemignani:
Hello, by fixing this line, the package is, at least for the small set of features tried by me, usable in noetic. 
Definitely more testing is needed and I know that no one is currently maintaining it, but by merging this PR, at least we could get the unmaintained version running in noetic.


mikepurvis:
Looks great, thanks.

PR #634, opened 2020-12-01
bump cmake min ver to 2.8.12

acxz:
resolves #633

PR #635, opened 2020-12-04
Fix for python3.9

ckurtz22:
(This breaks support for Python 3.4, if 3.4 was even supported)

Functionality for asyncio locks was changed and the current usage was deprecated, and finally removed in 3.9.
My changes are functional, as in catkin build works for me, but I'm not certain if I implemented all of the changes correctly so someone should double check my changes.

PR #637, opened 2020-12-09
remove older py35+xenial config and add py39+focal

acxz:
Addresses https://github.com/catkin/catkin_tools/pull/634#issuecomment-741904218

Decided to also add a focal configuration with the latest python. I can remove if it is unnecessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@
 language: generic
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


mikepurvis:
Looks like 3.9 fails, but should be fixable by #635, so will merge on spec.

PR #639, opened 2020-12-10
Install python2 before travis runs.

mikepurvis:
Small follow-up to #634, #635, #637.

This is needed only on Focal+ where python2 is not installed by default (and we run tests which depend on it being available), but because a `python2` package doesn't exist in the earlier versions, the install of it needs to be allowed to fail.


mikepurvis:
FYI @ckurtz22, @acxz.

PR #640, opened 2020-12-18
Add note for ROS environments that target Python3

methylDragon:
The standard install command for `catkin_tools` does not work on distributions like Noetic, since it relies on Python2 dependencies.

This PR adds a note in the setup instructions to cover that case.

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -27,6 +27,15 @@ Once you have added that repository, run these commands to install ``catkin_tool
 
     $ sudo apt-get update
     $ sudo apt-get install python-catkin-tools
+    
+.. note::
+
+    If you are on Noetic or any ROS distribution that targets Python3, instead use:
+    
+    .. code-block:: bash
+
+        $ sudo apt-get install python3-osrf-pycommon python3-catkin-tools
+
 
 Installing on other platforms with pip
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


mikepurvis:
Note that in principle, there should be nothing which prevents Python 2 catkin_tools from building a Python 3 workspace and vice-versa, particularly if catkin_tools is run from a venv. The issue historically has been in conflicts between py2/3 versions of common dependencies like catkin_pkg and rosdep. I believe most of this has how been fixed at this point.

I think my preference would be to suggest that all users use Python 3 catkin_tools and if system package conflicts prevent installing it with apt on your system, then install it from pip3 instead.

mintar:
I believe the PR is correct. On Noetic, installing the Python2 version of `catkin_tools` *will* lead to package conflicts, so why let users figure that out themselves? I believe people who try using the Python2 version on a Python3 ROS system are not doing that intentionally in 99% of cases.

This PR also includes the recommended workaround for #594, which is probably never going to be fixed.

I believe it's important that the installation instructions are correct, clear and succinct, and this PR achieves this. I think clarity is more important than covering all nuances.

It will save lots of new users a lot of time. It's really bad that the installation instructions have been incorrect for over a year now. Yet another colleague of mine was just bitten by this, so I was going to submit an identical PR when I discovered this one.

Can we just get this merged?

timonegk:
I agree with you and I understand your frustration. I am currently in contact with the previous maintainers of the repository and trying to get access to the documentation on readthedocs. At the moment, I could merge the pull request, but that would not really change anything as long as I cannot deploy it, since nobody will look at the unbuilt documentation in this repository, I think. Therefore, I would like to keep it open for now as a reminder for me to get access to the documentation. I'm sorry for your inconveniences :/
(By the way, I am also in contact with @wjwwood regarding the release of #594, so hopefully we will be able to get this done soon…)

mintar:
Thanks for the clarification, and thank you for taking care of this!

wjwwood:
> I am currently in contact with the previous maintainers of the repository and trying to get access to the documentation on readthedocs.

@timonegk you have access to RTD now, let me know if you need more permissions or have other issues.

timonegk:
Now that the new version is released, I updated the install documentation accordingly (see #687). There, the installation of the python 3 version is recommended for all ROS versions regardless of the ROS python version, because the python version used by catkin_tools is now independent of it.
I am therefore closing this PR.

Issue #523, opened 2018-09-21
Circular exec_depend not possible


luator:
[//]: # (Thank you for reporting an issue for catkin_tools!)

[//]: # (Please fill out the details below so that your issue can be resolved quickly!)
[//]: # (If you want to suggest a feature or start a discussion, feel free to delete them.)

### System Info

* Operating System: `Linux SNCN0051 4.4.0-135-generic #161-Ubuntu SMP Mon Aug 27 10:45:01 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
* Python Version: Python 2.7.12
* Version of catkin_tools:

      catkin_tools 0.4.4 (C) 2014-2018 Open Source Robotics Foundation
      catkin_tools is released under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
      ---
      Using Python 2.7.12 (default, Dec  4 2017, 14:50:18) [GCC 5.4.0 20160609]

* ROS Distro: kinetic

### Build / Run Issue

`catkin build` fails when I have two packages `foo` and `bar` with a circular `exec_depend`.

[//]: # (If you have an issue building and running code, please verify the following:)

* [x] Works with `catkin_make`
* [x] Works with `catkin_make_isolated --merge`
* [ ] Works with `catkin build`
* [ ] Works with `catkin build -p1`
* [ ] I did  not `read this`

### Expected Behavior

[//]: # (Please describe what you expected or think should happen.)
According to my understanding, `exec_depend` means, the dependency is only needed when executing something of the package. So during build, `exec_depend` should be completely irrelevant and thus a circular `exec_depend` should not be a problem for `catkin build`.

### Actual Behavior

[//]: # (Please describe what actually happens.)
```
❱ VERBOSE=1 catkin build -p1 -v
...                                                                                                                                                   
Traceback (most recent call last):
  File "/home/fwidmaier/src/catkin_tools/catkin_tools/verbs/catkin_build/build.py", line 549, in build_isolated_workspace
    continue_without_deps=False))
  File "/home/fwidmaier/src/catkin_tools/catkin_tools/execution/executor.py", line 374, in run_until_complete
    return loop.run_until_complete(coroutine)
  File "/usr/lib/python2.7/dist-packages/trollius/base_events.py", line 350, in run_until_complete
    return future.result()
  File "/usr/lib/python2.7/dist-packages/trollius/futures.py", line 285, in result
[...]

Issue #430, opened 2017-02-03
Unhelpful error message with circular dependency (part 2)


timotheecour:
related to https://github.com/catkin/catkin_tools/issues/229 but that was closed despite https://github.com/catkin/catkin_tools/issues/229#issuecomment-245881842 showing the fix didn't work (and that was catkin 0.4.2, I'm also seeing the bug with 0.4.3):

```
[build] Found '68' packages in 0.0 seconds.
Traceback (most recent call last):
  File "/usr/bin/catkin", line 9, in <module>
    load_entry_point('catkin-tools==0.4.3', 'console_scripts', 'catkin')()
  File "/usr/lib/python2.7/dist-packages/catkin_tools/commands/catkin.py", line 267, in main
    catkin_main(sysargs)
  File "/usr/lib/python2.7/dist-packages/catkin_tools/commands/catkin.py", line 262, in catkin_main
    sys.exit(args.main(args) or 0)
  File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/cli.py", line 419, in main
    summarize_build=opts.summarize  # Can be True, False, or None
  File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/build.py", line 305, in build_isolated_workspace
    packages, context, workspace_packages)
  File "/usr/lib/python2.7/dist-packages/catkin_tools/verbs/catkin_build/build.py", line 98, in determine_packages_to_be_built
    workspace_package_names = dict([(pkg.name, (path, pkg)) for path, pkg in ordered_packages])
AttributeError: 'str' object has no attribute 'name'
```

possibly also related to https://github.com/catkin/catkin_tools/issues/429 but not sure since it's not hanging

timotheecour:
workaround:

* build catkin_tools from source (https://github.com/catkin/catkin_tools/blob/master/docs/installing.rst#installing-from-source)
* edit brew/lib/python2.7/site-packages/catkin_tools/verbs/catkin_build/build.py:
```
for path, pkg in ordered_packages:
  print(('path, pkg', path, pkg))
```
right above
```
workspace_package_names = dict([(pkg.name, (path, pkg)) for path, pkg in ordered_packages])
```

this'll print suspicious items

kopp:
When you found the right place: Do you think that you can make this a change with pull request?

130s:
I opened a possible solution https://github.com/ros-infrastructure/catkin_pkg/pull/208

[...]

PR #641, opened 2021-01-05
Abort with error message on circular dependency.

iwanders:
Currently, when building a workspace with circular dependencies `catkin build` crashes with a backtrace and provides no useful information. Given three packages that are circular;
```
packageA, depends on packageB
packageB, depends on packageC
packageC, depends on packageA
```

Output from the current version is:
```
[build] Found '3' packages in 0.0 seconds.                                     
<snip>
  File "<snip>verbs/catkin_build/build.py", line 94, in determine_packages_to_be_built
    workspace_package_names = dict([(pkg.name, (path, pkg)) for path, pkg in ordered_packages])
  File "<snip>verbs/catkin_build/build.py", line 94, in <listcomp>
    workspace_package_names = dict([(pkg.name, (path, pkg)) for path, pkg in ordered_packages])
AttributeError: 'str' object has no attribute 'name'
```

Catkin does have all the necessary information to help the user resolve the issue. The `topological_order_packages` function provides the [info](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/topological_order.py#L166-L167) we need, stating that if a circular depedency exist, the last tuple in the list starts with `None` and contains a superset of the guilty packages. In the three package example the `ordered_packages` variable will hold `[(None, 'packageA, packageB, packageC')]
`.

With the change in this PR, the output instead becomes:
```
[build] Found '3' packages in 0.0 seconds.                                                                                                                                                                        
[build] Circular dependency detected in the following packages: packageA, packageB, packageC
```
and the `catkin build` command exits with `1`. This provides a clear message that helps the user resolve the issue.

FYI @mikepurvis , @jasonimercer

### DIFF
--- a/catkin_tools/verbs/catkin_build/build.py
+++ b/catkin_tools/verbs/catkin_build/build.py
@@ -86,9 +86,15 @@ def determine_packages_to_be_built(packages, context, workspace_packages):
     ordered_packages = topological_order_packages(workspace_packages)
     # Set the packages in the workspace for the context
     context.packages = ordered_packages
-    # Determin the packages which should be built
+    # Determine the packages which should be built
     packages_to_be_built = []
     packages_to_be_built_deps = []
+
+    # Check if topological_order_packages determined any circular dependencies, if so print an error and fail.
+    # If this is the case, the last entry of ordered packages is a tuple that starts with nil.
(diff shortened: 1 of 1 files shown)


mikepurvis:
Thanks for this! Looks like this is definitely the fix for #430, and possibly for #523 as well.

PR #644, opened 2021-01-26
Fail build if jobs were abandoned

### LINKED ISSUE #371: consider abandoning packages as an error
catkin build should return an error code when packages are abandoned. See #370.
If catkin build is used in CI, abandoning packages should be signaled as an error.

simonschmeisser:
fixes #371 

If there are valid use cases for continuing with the installation after jobs were abandoned, I can also turn this into an option. If so, please indicate if that should be on by default or off


mikepurvis:
LGTM.

PR #647, opened 2021-02-06
fix `catkin locate` for symlinked package src directories

timonegk:
In some places in `catkin locate`, `os.getcwd()` was used instead of `catkin_tools.common.getcwd()`. Therefore, `catkin locate` did not work for workspaces where packages are symlinked into the `src` directory.


timonegk:
Sorry, duplicate of #624.

PR #648, opened 2021-02-10
Fix stdeb for python 3

### LINKED ISSUE #594: Python 3 release - will be 0.7.0

timonegk:
This is my attempt to fix #594. The following changes are done:

- replace dependency catkin-pkg with catkin-pkg-modules (`> 0.2.9` can be dropped because catkin-pkg-modules was introduced with 0.3.0)
- `Depends` and `Conflicts` (i. e., the python 2 settings) were removed from `stdeb.cfg` because no further python 2 packages will be built.
- `python3-osrf-pycommon` was added as a deb dependency (I don't know why it was missing)

I tested the changes on five different systems:

- Debian 9 Stretch (python 3.5.3)
- Debian 10 Buster (python 3.7.3)
- Ubuntu 16.04 Xenial Xerus (python 3.5.2)
- Ubuntu 18.04 Bionic Beaver (python 3.6.9)
- Ubuntu 20.04 Focal Fossa (python 3.8.5)

Once all build dependencies were installed, I used the command `python3 setup.py --command-packages=stdeb.command sdist_dsc --with-python2=False bdist_deb` to create the deb file. It could be installed on all systems and catkin commands could be executed without error.
The mentioned systems cover all the systems in `Suite`, assuming that the non-LTS Ubuntu releases behave like either the predecessor or the successor.

Before the release with these changes, the trivial fix in https://github.com/osrf/osrf_pycommon/pull/71 should be merged. Currently, `python3-osrf-pycommon` and `python-osrf-pycommon` don't conflict in all cases, resulting in annoying apt errors on installing  `python3-osrf-pycommon` when `python-osrf-pycommon` was already installed.

Before the release, the version in `setup.py` an the changelog should obviously be updated.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -1,7 +1,5 @@
 [DEFAULT]
-Depends: python-argparse, python-setuptools, python-catkin-pkg (> 0.2.9), python-yaml
-Depends3: python3-setuptools, python3-catkin-pkg (> 0.2.9), python3-yaml
-Conflicts: python3-catkin-tools
+Depends3: python3-setuptools, python3-catkin-pkg-modules, python3-yaml, python3-osrf-pycommon
 Conflicts3: python-catkin-tools
 Suite: xenial yakkety zesty artful bionic cosmic disco eoan focal stretch buster
 X-Python3-Version: >= 3.5
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -4,3 +4,4 @@ Conflicts3: python-catkin-tools
 Suite: xenial yakkety zesty artful bionic cosmic disco eoan focal stretch buster
 X-Python3-Version: >= 3.5
 Setup-Env-Vars: DEB_BUILD=1
+No-Python2:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog for package catkin_tools
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+0.6.0 (2020-06-03)
+------------------
+* This release restores the 0.4.5 state due to an accident with the 0.5.0 release where we pushed it to Python2 users which it doesn't support.
+
 0.5.0 (2020-06-02)
 ------------------
 * Revert "jobs: Fixing environment required to run catkin test targets on pre-indigo catkin" (`#600 <https://github.com/catkin/catkin_tools/issues/600>`_)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,45 @@
 Changelog for package catkin_tools
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+0.7.0 (2021-03-01)
+------------------
+* Fix placeholders for cmake warning and error coloring (#678)
+* Fix catkin clean --all-profiles when not at workspace root (#673)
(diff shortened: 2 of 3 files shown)


timonegk:
There is still a problem, we cannot require `catkin-pkg-modules` in `setup.py` because that will break catkin_tools on systems where no apt packages for ros are available (`catkin-pkg-modules` is not a pip package). Therefore we have to use `catkin-pkg >= 0.3.0` in `setup.py`. This will now break on systems where only `python3-catkin-pkg-modules` is installed, because `catikn-pkg` is not available there. But `python3-catkin-pkg` can not be a requirement because it conflicts with ros. The best option might be to remove `catkin-pkg(-modules)` completely from `setup.py` – or find a way to depend on one of the two depending on pypi/deb installation...

timonegk:
The cleanest solution (in my opinion) would be to create a pypi package `catkin-pkg-modules` that contains no code but depends on `catkin-pkg`, essentially a meta package or wrapper around `catkin-pkg`. Then, `catkin-pkg-modules` could be the dependency for deb and pypi/pip.

timonegk:
Okay, I found another way, as `ros_buildfarm` is doing it.

timonegk:
Sorry for the spam, now everything works with pip and with stdeb. Yay!

gavanderhoorn:
`testing` doesn't appear to have a `ros-melodic-catkin` at the moment ([status_page/ros_melodic_default.html?q=catkin](http://repositories.ros.org/status_page/ros_melodic_default.html?q=catkin)). Could that be related to this PR?

Noetic seems OK: [status_page/ros_noetic_default.html?q=catkin](http://repositories.ros.org/status_page/ros_noetic_default.html?q=catkin).

timonegk:
As you found out, the build failure for `ros-melodic-catkin` is completely unrelated, this repository has no connection to `ros-melodic-catkin`.

mikepurvis:
Looks sane to me; thanks for the thorough effort at testing this!

timonegk:
I just added the changelog. Maybe you'll manage to take a look at it in the next few days.

stdeb.cfg
```diff
@@ -1,8 +1,7 @@
 [DEFAULT]
-Depends: python-argparse, python-setuptools, python-catkin-pkg (> 0.2.9), python-yaml
-Depends3: python3-setuptools, python3-catkin-pkg (> 0.2.9), python3-yaml
-Conflicts: python3-catkin-tools
+Depends3: python3-setuptools, python3-catkin-pkg-modules, python3-yaml, python3-osrf-pycommon
 Conflicts3: python-catkin-tools
-Suite: xenial yakkety zesty artful bionic cosmic disco eoan focal stretch buster
+Suite3: xenial yakkety zesty artful bionic cosmic disco eoan focal stretch buster
```
nuclearsandwich, line 4: I don't think this line change is necessary. The separate Suite3 config field is only needed when the release suites differ between Python 2 and Python 3.

I'm pretty certain that stdeb defaults to `unstable` if no suites are configured although I haven't confirmed that. Leaving this line as it was while adding the `No-Python2` field below should be sufficient to halt all further python2 releases of this package.

timonegk: Yes, you seem to be right. Thank you, I changed it back.

Issue #300, opened 2016-03-03
bash completion files install to the wrong location


wjwwood:
I just noticed that they are installed to `/etc` when I think they ought to be installed to `/usr/local/etc`. I think this is because we fallback to `sys.prefix` which isn't necessarily where files are installed to.

timonegk:
Closed in #652.

PR #652, opened 2021-02-17
Fix shell completion install locations

timonegk:
This pull request cleans up the shell completion locations. Bash completion will no longer be written to the appropriate directory depending on pip/deb install and not to `/etc`, zsh completion will be written to a directory that is actually on the zsh function search path.

The following changes were made:
- When installing via `pip`, bash completions will be installed to `<prefix>/share/bash-completion/completions` where prefix is `/usr/local` for `sudo pip` and `$HOME/.local` for `pip install --user`. Both locations are used by `bash-completion` (see https://github.com/scop/bash-completion/blob/master/bash_completion#L2261), therefore I removed the bash-completion instructions. For the deb, the prefix is `/usr`, and the resulting bash-completions directory is also searched by default. The file had to be renamed to one of `catkin`, `catkin.bash` or `_catkin` to be recognized. I chose `catkin.bash` to make it clear that the file belongs to `bash`.
- zsh completion is normally installed to `<prefix>/share/zsh/site-functions`. This works out of the box for `/usr/local`, but not for `$HOME/.local`. However, there is no default non-root completion directory for zsh, therefore I did not change that directory. When building the deb, the completion gets installed to `/usr/share/zsh/vendor-completions` since `/usr/share/zsh/site-functions` is not on the default `fpath` (the path for zsh functions), which is also the reason why the current zsh completion does not work out of the box.

I unfortunately don't have a OSX system to test, but I did not change the zsh completion target and assume that when the `/etc/bash_completion.d` directory works there, `/usr/share/bash-completion` will too.


timonegk:
I also removed the invocation of distutils introduced in #569 and instead change the default path to `''`. This lets setuptools deal with the prefix.

timonegk:
This would also close #300.

caioaamaral:
Any updates on this? Just notice that completion is not working for the xenial release

timonegk:
I just tested this on xenial and it works, so I am going to merge it now.

PR #657, opened 2021-02-25
escape whitespaces in `catkin config` printout

pseyfert:
In the `catkin config` printout, the cmake-args line changes from
```
Additional CMake Args:       -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_CXX_FLAGS=-g -fno-omit-frame-pointer -O3 -DNDEBUG -DENABLE_TIMING=0
```
to
```
Additional CMake Args:       -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_CXX_FLAGS=-g\ -fno-omit-frame-pointer\ -O3\ -DNDEBUG -DENABLE_TIMING=0
```
thereby the printout can be copy and pasted to `catkin config
--cmake-args` without changing the configuration. In the above example
the last element `-DENABLE_TIMING` is a cmake option and not a
gcc preprocessor definition. Before this patch, this is
indistinguishable from the printout.

I did not thoroughly check which of the other entries in the printout might have whitespaces that deserve escaping - cmake-args is the only one where it occurs to me.

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -578,6 +578,9 @@ class Context(object):
             install_layout = 'merged' if not self.__isolate_install else 'isolated'
 
         def quote(argument):
+            # Distinguish in the printout if space separates two arguments or if we
+            # print an argument with a space.
+            # e.g. -DCMAKE_C_FLAGS="-g -O3" -DCMAKE_C_COMPILER=clang
             if ' ' in argument:
                 if "=" in argument:
                     key, value = argument.split("=", 1)
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -579,8 +579,11 @@ class Context(object):
 
         def quote(argument):
             if ' ' in argument:
-                key, value = argument.split("=")
-                return key + '="' + value + '"'
+                if "=" in argument:
+                    key, value = argument.split("=", 1)
+                    if ' ' not in key:
+                        return key + '="' + value + '"'
+                return '"' + argument + '"'
             return argument
 
         subs = {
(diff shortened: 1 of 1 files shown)


timonegk:
I think I would prefer to enclose the argument in quotes if it contains spaces, rather than using backslashes, as I think it looks more understandable. Would this work for you as well?

pseyfert:
(shortening the above example)
Using quotes will work, too. And I agree this looks better. I think like
```
-DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_CXX_FLAGS="-g -O3"
```
would be nicest to read, but right now I don't see how to get this done at the same complexity on the generation side (here I put quotes only if there actually is a whitespace occuring, and then I quote the part after `=`).

At the same complexity to generate while still grouping the cmake arguments and being valid copy&pastes would be
```
"-DCMAKE_EXPORT_COMPILE_COMMANDS=ON" "-DCMAKE_CXX_FLAGS=-g -O3"
```
(here I put every cmake argument in quotes, needed or not)

timonegk:
I would imagine something like this

```python
def quote(argument):
    if ' ' in argument:
        key, value = argument.split('=')
        return key + '="' + value + '"'
    return argument
```

pseyfert:
indeed that works. i used `split("=", 1)` to make sure only the first = gets split off for arguments like `-DCMAKE_CXX_FLAGS=-DPREPROCESSOR_VARIABLE=VALUE`.

timonegk:
Ah yes, that's better. You don't seem to have pushed it, though?

pseyfert:
:see_no_evil: whoops. also added handling a few more edge cases (absence of `=`, first space before the first `=`). I don't have examples for those but i think it's important `catkin config` doesn't crash should somebody run into such a scenario.

timonegk:
Very nice, thank you. Finally please add a comment to the function describing why we need it and then we're good to go.

PR #668, opened 2021-03-23
Use loadavg over the last 1 minute

mobangjack:
Make the build process more sensitive to the current load.
We've observed that using loadavg over the last 5 minutes can be a little slow in response. 1 minute seems to be more reasonable.


mobangjack:
Hi @timonegk , would you please take a look at this PR?

timonegk:
Makes sense, thanks for the PR.

PR #673, opened 2021-04-07
Fix catkin clean --all-profiles when not at workspace root

zig-for:
Currently it silently fails because it tries to read cwd as a workspace, which may or may not contain profiles.


asm (approved):

timonegk (approved):
Thank you!

PR #688, opened 2021-08-19
WIP: Use only CATKIN_IGNORE to ignore packages, not COLCON_IGNORE/AMENT_IGNORE

janstrohbeck:
My use case: I have a shared source space in which both ROS1 and ROS2 packages reside (the migration to ROS2 is ongoing). Now I want to build the ROS1 packages using catkin_tools, same as before, and the ROS2 packages using colcon. To do this, I place CATKIN_IGNORE files in the ROS2 packages, and COLCON_IGNORE files in the ROS1 packages.

Problem: catkin_tools/catkin_pkg ignores the ROS1 packages due to an existing COLCON_IGNORE.

I think catkin_tools should ignore ROS1 packages only when there is a CATKIN_IGNORE file present, and not when COLCON_IGNORE/AMENT_IGNORE is present.

This PR depends on a PR in catkin_pkg: https://github.com/ros-infrastructure/catkin_pkg/pull/307

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -22,7 +22,7 @@ import sys
 from fnmatch import fnmatch
 from itertools import chain
 
-from catkin_pkg.packages import find_packages
+from catkin_pkg.packages import find_packages as _find_packages
 
 from .terminal_color import ColorMapper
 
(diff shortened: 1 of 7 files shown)


timonegk:
By the way, because I just noticed that I never responded to this PR: I like your suggestion, but we have to wait for the merge of the other pull request, of course.

rhaschke (approved):

rhaschke:
> we have to wait for the merge of the other pull request, of course.

@timonegk: https://github.com/ros-infrastructure/catkin_pkg/pull/307 was merged meanwhile.

timonegk:
@rhaschke thanks for the update, I will merge it in the evening and create a new release.

timonegk:
@wjwwood I just tagged 0.9.5 and pushed it to PyPI. I don't know if you still do the apt releases, but if you do, could you please pull the trigger there?

PR #691, opened 2021-10-06
Some space fixes

MatthijsBurgh:
Fix summary for spaces
Handle non-default spaces in `catkin locate`

### DIFF
--- a/catkin_tools/verbs/catkin_clean/cli.py
+++ b/catkin_tools/verbs/catkin_clean/cli.py
@@ -203,7 +203,7 @@ def clean_profile(opts, profile):
             if getattr(ctx, '{}_space_exists'.format(space))():
                 space_name = Context.SPACES[space]['space']
                 space_abs = getattr(ctx, '{}_space_abs'.format(space))
-                spaces_to_clean_msgs.append(clr("[clean] {}:     @{yf}{}").format(space_name, space_abs))
+                spaces_to_clean_msgs.append(clr("[clean] {:14} @{yf}{}").format(space_name + ':', space_abs))
 
         if len(spaces_to_clean_msgs) == 0 and not opts.deinit:
             log("[clean] Nothing to be cleaned for profile:  `{}`".format(profile))
--- a/catkin_tools/verbs/catkin_clean/cli.py
+++ b/catkin_tools/verbs/catkin_clean/cli.py
@@ -104,18 +104,18 @@ def prepare_arguments(parser):
         ' the workspace.')
 
     # Basic group
-    basic_group = parser.add_argument_group(
+    spaces_group = parser.add_argument_group(
         'Spaces',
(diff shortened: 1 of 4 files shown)


mikepurvis:
Looks reasonable. Are there any other cases like this in the source, or is this the only one?

MatthijsBurgh:
I checked the rest of `context`, which has most spaces stuff in it. Context has `package_SPACE_NAME_space` functions, but for devel and install space, this really different, so these are hard to generate dynamically.
For the other spaces it is just a path join. So we could skip if the attr already exists(devel and install), otherwise add the path join function.

As extending the space descriptions to add such custom functions is to complex IMO.

I also found that `catkin locate` can't handle other spaces.

MatthijsBurgh:
Friendly ping ;)

MatthijsBurgh:
@mikepurvis friendly ping ;)

MatthijsBurgh:
Friendly ping @mikepurvis @timonegk ;)

catkin_tools/verbs/catkin_clean/cli.py
```diff
@@ -191,21 +193,17 @@ def clean_profile(opts, profile):
...
-        if opts.devel and devel_exists:
-            spaces_to_clean_msgs.append(clr("[clean] Devel Space:   @{yf}{}").format(ctx.devel_space_abs))
-        if opts.install and install_exists:
-            spaces_to_clean_msgs.append(clr("[clean] Install Space: @{yf}{}").format(install_path))
+    if opts.spaces and not (opts.yes or opts.dry_run):
+        for space in opts.spaces:
+            if getattr(ctx, '{}_space_exists'.format(space))():
+                space_name = Context.SPACES[space]['space']
+                space_abs = getattr(ctx, '{}_space_abs'.format(space))
+                spaces_to_clean_msgs.append(clr("[clean] {}:     @{yf}{}").format(space_name, space_abs))
```
timonegk, line 206:
```suggestion
                spaces_to_clean_msgs.append(clr("[clean] {:14} @{yf}{}").format(space_name + ':', space_abs))
```
I would suggest this change to align the output (as it was the case before).

timonegk (commented):
Apart from a small cosmetic suggestion I think this looks fine.

MatthijsBurgh:
When this PR is merged, could a new release be scheduled? As the current releases don't accept any non-default pluggable spaces during execution.

timonegk:
Sure. I just released a new version (0.8.0) to PyPI. @wjwwood could you release the new version to the package repositories? This time I released from master and also tagged the version ;)

wjwwood:
@timonegk sure thing, and thanks!

wjwwood:
Ok, it is available from the ROS 1 repository and soon from the ROS 2 repository (waiting on build farm queue).

PR #692, opened 2021-10-14
More cleanup

MatthijsBurgh:
I wasn't planning, but then I did do this extra cleaning up.


timonegk:
This is great, thank you very much!

PR #702, opened 2021-12-01
Add back flags that were removed in #691

### LINKED ISSUE #701: --src option and stability of the CLI
### System Info

* Operating System: Ubuntu
* Python Version: 3.8.10
* Version of catkin_tools: 0.8.0
* ROS Distro: noetic

### Build / Run Issue

In #691 the `--src` option was replaced by `--source-space`. This change broke our build server as it used `catkin locate ... --src PACKAGE` to locate the source directory.
I would suggest keeping the old name as an alias to stay backward compatible.
That is of course unless the CLI is not intended to be used for scripting, in that case, an appropriate notice should probably be added to the docs.

timonegk:
Closes #701.

### DIFF
--- a/catkin_tools/verbs/catkin_clean/cli.py
+++ b/catkin_tools/verbs/catkin_clean/cli.py
@@ -112,9 +112,9 @@ def prepare_arguments(parser):
     for space, space_dict in Context.SPACES.items():
         if space == 'source':
             continue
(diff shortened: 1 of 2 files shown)


timonegk:
@wjwwood Sorry for all the releases. But could you do one more? It's 0.8.2.

wjwwood:
It's done. 👍

PR #709, opened 2022-01-07
Grammar correction to catkin_config.rst

DanielDowling1:
"Its" does not adopt an apostrophe of ownership. The apostrophic variant only denotes the contraction for "it is".


janbernloehr (approved):

jbohren:
It's better to make this change here: https://github.com/catkin/catkin_tools/pull/622/files#r455205583

PR #710, opened 2022-01-13
Fix catkin clean for single packages

MichaelGrupp:
The clean command looks up the installed files from the
install_manifest.txt of the package. But this file was not copied to
the install space.

This patch adds that copy in the same way it is already done in the
plain CMake build job (see jobs/cmake.py).

In addition to that, installed Python modules are also removed if
their package name is the same as the catkin package.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,34 @@ Changelog for package catkin_tools
 Upcoming
 --------
 
+0.8.2 (2021-12-01)
+------------------
+
+* Add back flags that were removed in #691 (#702)
+* Documentation: Update CLI documentation
+
+0.8.1 (2021-11-30)
+------------------
+
+* catkin test: add --limit-status-rate option (#699)
+
+0.8.0 (2021-11-30)
+------------------
+
+* Add catkin test verb (#676)
+* Code cleanup (#692)
+* Documentation: https in docs, properties over vars, sphinx cleanup (#690)
+* Documentation: Improve wording (#694)
+* Documentation: Add custom build, devel, install dirs (#697)
+* Contributors: Balint, Jim Aldon D'Souza, Matthijs van der Burgh, mobangjack, pseyfert, Timon Engelke
+
+0.7.2 (2021-11-18)
+------------------
+
+* Update installation docs for python 3 (#687)
+* Fix environment variable parsing (#686)
+* Switch from Travis CI to GitHub actions (#684)
+* Regenerate setup files when the install space was cleaned (#682)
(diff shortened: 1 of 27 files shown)


timonegk:
I agree that it would be good to remove the installed files with catkin clean and that the current stage `cleaninstall` in `create_catkin_clean_job` does not really make sense as there are never any packages to clean for catkin packages.

Currently, there are two possible setups for the install space: the merged layout and the isolated layout. For the isolated layout, your approach works fine, but for the merged layout, it must be adapted. The install_manifest.txt also contains common files (e.g. the `setup.bash` file). Consequently, they would also be removed by catkin clean, rendering the install space unusable.

Therefore it would be necessary to add a way of distinguishing the package specific files from the common files added by catkin to avoid removing too many files.

MichaelGrupp:
@timonegk I added a patch that filters the files in the shared install base directory when deleting: https://github.com/catkin/catkin_tools/pull/710/commits/798a24235f52583de483401351ce9762f82a3b12
Seems to work, but I don't know if there are some corner case where this rather cheap trick could be broken?

timonegk:
This looks fine to me. The solution might not be perfect because in theory, any package could install to the install directory root. But that would only lead to these files not being removed in this particular case, which is still better than not removing any files.

MichaelGrupp:
One more thing that should be resolved: when cleaning a single package with Python code, the modules in `install/lib/python3/dist-packages/` are not removed. This would lead to unexpected behavior because module imports would still use this path after cleaning.

MichaelGrupp:
Added another commit to handle Python packages that follow the [standard ROS practice](https://docs.ros.org/en/api/catkin/html/howto/format2/installing_python.html#modules) for package naming. I also changed the name of the PR since the scope is now a bit more than just the install_manifest.

Now everything works in my local test environment with mixed C++/Python Catkin packages.

timonegk:
Yes, that is probably the best way to do it currently. In theory, we should be able to get the exact list of files from `setup.py` via the `--record` parameter. But this parameter would have to be passed through the catkin / cmake pipeline, and there is currently no way to do that. (If you wanted to look into that, the call to `setup.py` is done in [python_distutils_install.sh.in](https://github.com/ros/catkin/blob/a31d8899b2567fe89173c2d0f49b7c66da803f91/cmake/templates/python_distutils_install.sh.in). But it is probably not worth it.)
I think your solution looks fine, I will test it a bit more to make sure it doesn't break anything. If you have the resources, do you mind adding a test?

timonegk:
I added a test and everything works nicely, so I will merge it now. Thanks again for the patch!

PR #712, opened 2022-02-14
adding better terminology for selecting which packages to build

jbohren:
Overview:
- Changes all instances of "whitelist" to "buildlist"
- Changes all instances of "blacklist" to "skiplist"
- Permits usage of old cli arguments
- Warns the user when old cli arguments are used

@allenh1 I appreciate you putting the time in to address the poorly chosen terminology present in catkin_tools. This PR adds terminology which I think is clearer and better-suited for this application than "allow" and "deny". Since you made the initial PR, I added you as a coauthor to the main commit in this PR. Let me know if you want me to remove you from that commit.

This should supersede:
- https://github.com/catkin/catkin_tools/pull/622
- https://github.com/catkin/catkin_tools/pull/709


allenh1:
Great -- thanks for following up on this.

wjwwood:
I support this change, but I don't have time to take this pull request on, sorry. Maybe @timonegk can give it a review and merge.

catkin_tools/verbs/catkin_config/cli.py
```diff
@@ -70,18 +70,19 @@ def prepare_arguments(parser):
...
     add = lists_group.add_mutually_exclusive_group().add_argument
-    add('--blacklist', metavar="PKG", dest='blacklist', nargs="+", required=False, type=str, default=None,
-        help='Set the packages on the blacklist. Packages on the blacklist are '
+    add('--skiplist', '--blacklist', metavar="PKG", dest='skiplist', nargs="+", required=False, type=str, default=None,
+        help='Set the packages on the skiplist. Packages on the skiplist are '
         'not built with a bare call to `catkin build`.')
-    add('--no-blacklist', dest='blacklist', action='store_const', const=[], default=None,
-        help='Clear all packages from the blacklist.')
+    add('--no-skiplist', '--no-blacklist', dest='skiplist', action='store_const', const=[], default=None,
+        help='Clear all packages from the skiplist.')
```
timonegk, line 85: I would prefer if `--blacklist` / `--whitelist` were no longer visible on the help page. You can achieve that by adding the old terms separately to the mutually exclusive group and setting `help=argparse.SUPPRESS`.

timonegk (changes requested):
Thank you very much for the pull request, I really appreciate the changes!
Currently, the configuration yaml file also contains the blacklist/whitelist terms. In `metadata.py`, there is a function to migrate metadata. The migration to the new terms should be added there.
Also, the CI is currently failing. Could you look into that?

catkin_tools/verbs/catkin_config/cli.py
jbohren: Yeah, that would be better, but I won't be adding it.

catkin_tools/commands/catkin.py
```diff
@@ -194,12 +194,32 @@ def catkin_main(sysargs):
...
+        print('')
+        print(fmt('@!@{rf}WARNING:@| Some arguments are deprecated and will be'
+                  ' removed in a future release.\n'))
+        print('Please switch to using their replacements as follows:')
+        for old_arg, new_arg in used_deprecated_args:
+            print(" - note: '{}' is deprecated, use '{}' instead".format(
+                    old_arg,
+                    new_arg))
+        print('')
+
```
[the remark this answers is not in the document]
jbohren: Moving it to config will be better, but I won't be able to do that.

The list of `note:` lines is a pretty common pattern (e.g. gcc) when there's a high-level issue and several specific notes for fixing the issue. I'm fine if you want to remove that though.

timonegk:
Thanks!

timonegk:
On the same note, I also changed the name of the base branch from `master` to `main`.

timonegk:
@wjwwood I just tagged a new release (0.8.3)

jbohren:
@wjwwood @timonegk thanks for following through on this!

timonegk:
Oh yes, I forgot about PyPi. Thanks for both!

timonegk:
Nothing here is a breaking change.

MatthijsBurgh:
It is for plugins. For example [`catkin_tools_document`](https://github.com/mikepurvis/catkin_tools_document/). As it accesses the `black/whitelist` attributes of `Context`, see https://github.com/mikepurvis/catkin_tools_document/pull/25/

timonegk:
Ah yes, that's true... Sorry 😕

timonegk:
Okay, then I would do the following to mitigate the issues:
- Add a new patch release that adds back the `black/whitlist` attributes of `Context`, migrates config files back to the old terms and can deal with both types of arguments. That should add back all breaking changes.
- Release a new version 1.0.0 with the migrated config and `Context` class.
What do you think of this plan? Did I forget anything?

PR #714, opened 2022-03-14
Fix breaking change in 0.8.3

timonegk:
This adds back backwards compatibility to the changes in 0.8.3 (see #712), namely
- Add back `whitelist` and `blacklist` attributes of `Context`
- Add back the corresponding parameters to the constructors of `Context` and `ConsoleStatusController`
- Add back `DEVEL_LINK_BLACKLIST` and `DEVEL_LINK_PREBUILD_BLACKLIST` variables
- Rename `skiplist` and `buildlist` in the configuration file to the old terms and revert the migration if necessary. This is to make the configuration files backwards compatible as well.

All of these changes will eventually be reverted in version 1.0.0.


timonegk:
Sorry once again to everyone for the inconveniences caused by my lack of foresight while releasing new versions.

timonegk:
@wjwwood Could you release this (0.8.4) to packages.ros.org? I already did the pypi release.

wjwwood:
I pushed the debs, but there was an issue on the ros1 buildfarm. I'll have to wait for someone to help me with that, but it should be available from the ros2 package repositories already.

PR #716, opened 2022-03-18
Correct usage check of build/skiplist

MatthijsBurgh:
In case an user uses `--no-buildlist` or `--no-skiplist` the arguments are set to an empty list, this resolves to `False` in the old situation. While it should check for it not being set at all, which is equal to `None`.

P.S. these `assert_in_config` checks need to be updated to buildlist/skiplist for the breaking release.

### DIFF
--- a/catkin_tools/verbs/catkin_config/cli.py
+++ b/catkin_tools/verbs/catkin_config/cli.py
@@ -168,10 +168,10 @@ def main(opts):
         no_action = not any(actions)
 
         # Handle old argument names necessary for Context.load
-        if opts.buildlist:
+        if opts.buildlist is not None:
             opts.whitelist = opts.buildlist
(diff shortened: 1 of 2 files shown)


MatthijsBurgh:
See the GH action run on my repo for it failing on adding the test: https://github.com/MatthijsBurgh/catkin_tools/actions/runs/2003199792
But it succeeding when fixing the code:
https://github.com/MatthijsBurgh/catkin_tools/actions/runs/2003201973 (This should of course be the same as the GH actions PR run on this repo)

timonegk:
Thank you!

MatthijsBurgh:
@timonegk could you please create a new release including this fix?

wjwwood:
Should be done. 👍

PR #719, opened 2022-03-27
Close logger in finally

### LINKED ISSUE #700: Logger closing
I am not experiencing any issues. But I was scrolling a bit through the code and found that in `async_job` the logger isn't closed, when there is any interruption, Exception, break, continue.

https://github.com/catkin/catkin_tools/blob/6d9f6fdc86c56850e521e6a665b89b033612a2f8/catkin_tools/execution/executor.py#L172-L179

I think this is incorrect. The logger should always be closed.

A `finally` will also be executed on a `return`, `break`, `continue`. See https://docs.python.org/3/reference/compound_stmts.html#finally

So by just moving the `logger.close()` to the finally should do the job.

MatthijsBurgh:
Fixes #700

### DIFF
--- a/catkin_tools/execution/executor.py
+++ b/catkin_tools/execution/executor.py
@@ -174,7 +174,8 @@ async def async_job(verb, job, threadpool, locks, event_queue, log_path):
                 break
 
         finally:
-            logger.close()
+            if logger is not None:
+                logger.close()
 
             lock.release()
 
--- a/catkin_tools/execution/executor.py
+++ b/catkin_tools/execution/executor.py
@@ -173,9 +173,9 @@ async def async_job(verb, job, threadpool, locks, event_queue, log_path):
             if retcode == stage.early_termination_retcode:
                 break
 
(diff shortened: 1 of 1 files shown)


catkin_tools/execution/executor.py
```diff
@@ -173,9 +173,9 @@ async def async_job(verb, job, threadpool, locks, event_queue, log_path):
             if retcode == stage.early_termination_retcode:
                 break
 
-            # Close logger
-            logger.close()
         finally:
+            logger.close()
```
meyerj, line 177:
```suggestion
            if logger: logger.close()
```
?

`logger` is only assigned within the `try` block and may still be `None` if an exception is raised early.

MatthijsBurgh:
```suggestion
            if logger is not None:
                logger.close()
```

MatthijsBurgh:
@timonegk friendly ping ;)

PR #729, opened 2022-05-16
remove leading and trailing ':' in CMAKE_PREFIX_PATH

lucasw:
I had some CMAKE_PREFIX_PATH modifications like the following that resulted in leading `:`

```
export CMAKE_PREFIX_PATH=$CMAKE_PREFIX_PATH:/my/path/...
```

and the result is that `::` gets baked into the CMAKE_PREFIX_PATH later, but then a mismatch occurs:

```
WARNING: Your current environment's CMAKE_PREFIX_PATH is different from the cached CMAKE_PREFIX_PATH used the last
time this workspace was built.

If you want to use a different CMAKE_PREFIX_PATH you should call `catkin clean` to remove all references to the previous
CMAKE_PREFIX_PATH.

Cached CMAKE_PREFIX_PATH:
	/home/lucasw/base_catkin_ws/devel:/home/lucasw/other/install:/home/lucasw/other/install/lib/cmake
Current CMAKE_PREFIX_PATH:
	/home/lucasw/base_catkin_ws/devel::/home/lucasw/other/install:/home/lucasw/other/install/lib/cmake
```

so added `strip(':')` everywhere CMAKE_PREFIX_PATH is loaded from the environment

### DIFF
--- a/catkin_tools/context.py
+++ b/catkin_tools/context.py
@@ -449,7 +449,7 @@ class Context(object):
 
         self.cached_cmake_prefix_path = ''
         if 'CMAKE_PREFIX_PATH' in sticky_env:
-            split_result_cmake_prefix_path = sticky_env.get('CMAKE_PREFIX_PATH', '').split(':')
+            split_result_cmake_prefix_path = sticky_env.get('CMAKE_PREFIX_PATH', '').strip(':').split(':')
             if len(split_result_cmake_prefix_path) > 1:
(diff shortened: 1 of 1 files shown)


timonegk (changes requested):
I would prefer to use `.replace('::', ':').strip(':')` everywhere so you don't need strip-replace-strip and trailing double colons are removed everywhere.

PR #741, opened 2022-12-18
Fix deprecated asyncio.wait() usage

roehling:
The explicit passing of coroutine objects to asyncio.wait() has been deprecated since Python 3.8 and breaks with Python 3.11.

A very simple fix is to wrap the coroutine in a task.


v4hn:
@timonegk the patch is [in the debian package](https://salsa.debian.org/science-team/ros-catkin-tools) and Debian testing started rolling python 3.11 now.
Could you please merge & release with the fix?

timonegk:
Sorry for the delay, thanks for the reminder and the fix.

PR #753, opened 2023-05-08
Add bookworm as a python3 target

tfoote:
This will need a release or manual flood to deploy.


timonegk (approved):
We can probably also remove all the EOL ubuntu distributions and add jammy and lunar, right?

PR #761, opened 2023-08-14
chore: fix cmake min version deprecation warning

rickstaa:
Not sure if this is desirable given back-compatibility, but this pull request fixed the cmake min version < 3.5 deprecation warning:

```bash
Warnings   << catkin_pkg_make_err:cmake /home/ricks/development/work/catkin_tools_ws/logs/catkin_pkg_make_err/build.cmake.000.log                                                                          
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):                                                                                                                                    
  Compatibility with CMake < 3.5 will be removed from a future version of                                                                                                                                  
  CMake.                                                                                                                                                                                                   
                                                                                                                                                                                                           
  Update the VERSION argument <min> value or use a ...<max> suffix to tell                                                                                                                                 
  CMake that the project does not need compatibility with older versions.
```

### DIFF
--- a/catkin_tools/jobs/catkin.py
+++ b/catkin_tools/jobs/catkin.py
@@ -732,7 +732,7 @@ DEVEL_LINK_BLACKLIST = DEVEL_LINK_SKIPLIST
 
 # CMakeLists.txt for prebuild package
 SETUP_PREBUILD_CMAKELISTS_TEMPLATE = """\
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.10)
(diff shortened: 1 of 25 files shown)


rickstaa:
We can also update to version 3.10 since this is what is suggested by [the migration guide](https://wiki.ros.org/noetic/Migration#Increase_required_CMake_version_to_avoid_author_warning).

rickstaa:
> Yes, let's do 3.10 then. ROS Noetic targets Cmake > 3.10 anyway.

Cool. Done!

PR #766, opened 2023-09-02
Replace pkg_resources by importlib.metadata

MatthijsBurgh:
`pkg_resources` is deprecated and was causing issues in finding dependencies of `catkin_tools_document`.

I replaced it by `importlib.metadata`. I am not 100% sure the changes in `catkin_tools/metadata.py` are correct.

The call on `entry_points()` is expensive, therefore I added a private function in combination with a global variable. I only applied this in the `commands/catkin.py`. Ideally this is shared between files. Please let me know when you have suggestion to it a bit more cleanly and share it between the files.

### DIFF
--- a/catkin_tools/commands/catkin.py
+++ b/catkin_tools/commands/catkin.py
@@ -19,9 +19,9 @@ from datetime import date
 from shlex import quote as cmd_quote
 
 try:
-    from importlib.metadata import distribution, entry_points
+    from importlib.metadata import entry_points, version
 except ImportError:
-    from importlib_metadata import distribution, entry_points
+    from importlib_metadata import entry_points, version
 
 from catkin_tools.common import is_tty
 from catkin_tools.config import get_verb_aliases
@@ -196,7 +196,7 @@ def catkin_main(sysargs):
     # Check for version
     if '--version' in sysargs:
         print('catkin_tools {} (C) 2014-{} Open Source Robotics Foundation'.format(
-            distribution('catkin_tools').version,
+            version('catkin_tools'),
             date.today().year)
         )
         print('catkin_tools is released under the Apache License,'
--- a/catkin_tools/commands/catkin.py
+++ b/catkin_tools/commands/catkin.py
@@ -18,7 +18,10 @@ import sys
 from datetime import date
 from shlex import quote as cmd_quote
 
-import pkg_resources
+try:
+    from importlib.metadata import distribution, entry_points
+except ImportError:
+    from importlib_metadata import distribution, entry_points
 
 from catkin_tools.common import is_tty
 from catkin_tools.config import get_verb_aliases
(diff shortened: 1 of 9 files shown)


catkin_tools/commands/catkin.py
```diff
@@ -18,7 +18,10 @@
 from datetime import date
 from shlex import quote as cmd_quote
 
-import pkg_resources
+try:
+    from importlib.metadata import distribution, entry_points
+except ImportError:
+    from importlib_metadata import distribution, entry_points
```
graingert, line 24: importlib.metadata is always available because catkin_tools only supports 3.7+ so this branch is never hit

catkin_tools/commands/catkin.py
```diff
@@ -29,16 +32,25 @@
...
+    global ENTRY_POINTS
+    if ENTRY_POINTS is None:
+        ENTRY_POINTS = entry_points()
+    return ENTRY_POINTS
+
 
 def list_verbs():
     verbs = []
-    for entry_point in pkg_resources.iter_entry_points(group=CATKIN_COMMAND_VERB_GROUP):
+    for entry_point in _get_entry_points().get(CATKIN_COMMAND_VERB_GROUP, []):
```
graingert, line 47: this is deprecated DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.

catkin_tools/commands/catkin.py
```diff
@@ -184,7 +196,7 @@ def catkin_main(sysargs):
     # Check for version
     if '--version' in sysargs:
         print('catkin_tools {} (C) 2014-{} Open Source Robotics Foundation'.format(
-            pkg_resources.get_distribution('catkin_tools').version,
+            distribution('catkin_tools').version,
```
graingert, line 199: there's importlib.metadata.version as a shortcut for this

catkin_tools/commands/catkin.py
MatthijsBurgh: This is only available from 3.8,https://docs.python.org/3/library/importlib.metadata.html

graingert: I mean 3.8+ https://github.com/catkin/catkin_tools/blob/main/.github/workflows/workflow.yml#L12

graingert:
I have a PR that fixes the deprecation warnings: https://github.com/catkin/catkin_tools/pull/767

graingert:
> But it lacks the single call to entry_points for now.

It uses functools.lru_cache to do a single call

timonegk:
> > But it lacks the single call to entry_points for now.
> 
> It uses functools.lru_cache to do a single call

But only for the verb entry points, in this PR there is a single call to entry_points for all groups combined

MatthijsBurgh:
This PR also does multiple calls for now. But we should combine the code regarding it of this PR and #767

PR #774, opened 2025-10-27
Add CATKIN_IGNORE_ROS2 marker support for mixed ROS1/ROS2 workspaces

iory:
This PR adds support for a new marker file `CATKIN_IGNORE_ROS2` to enable better coexistence of ROS1 and ROS2 packages in the same workspace.

## Problem
Currently, when using `CATKIN_IGNORE` in a mixed ROS1/ROS2 workspace:
- Both catkin_tools and colcon recognize and ignore packages with `CATKIN_IGNORE`
- This makes it impossible to have catkin_tools ignore ROS2 packages while allowing colcon to build them

## Solution
Introduce a new marker file `CATKIN_IGNORE_ROS2` that:
- Is recognized only by catkin_tools (colcon does not recognize it)
- Allows catkin_tools to skip ROS2 packages without affecting colcon's behavior
- Maintains backward compatibility with existing `CATKIN_IGNORE` functionality

## Changes
- Added `CATKIN_IGNORE_ROS2` to the ignore_markers set in `find_packages()`
- Updated help message in catkin clean command to mention the new marker

## Usage
For ROS2 packages that should be ignored by catkin_tools but built by colcon:
```bash
touch ros2_package/CATKIN_IGNORE_ROS2
```

This enables truly mixed ROS1/ROS2 workspaces where each build tool can operate independently.

### DIFF
--- a/catkin_tools/common.py
+++ b/catkin_tools/common.py
@@ -522,8 +522,10 @@ def wide_log(msg, **kwargs):
 
 
 def find_packages(*args, **kwargs):
-    """
-    Crawls the filesystem to find package manifest files. Ignores subfolders if CATKIN_IGNORE is present.
+    """Crawls the filesystem to find package manifest files.
+
+    Ignores subfolders if CATKIN_IGNORE or CATKIN_IGNORE_ROS2 is present.
+
     :param basepath: The path to search in, ``str``
     :param exclude_paths: A list of paths which should not be searched, ``list``
     :param exclude_subspaces: The flag is subfolders containing a .catkin file should not be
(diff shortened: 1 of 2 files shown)


mikepurvis:
I've been handling this scenario with `colcon build --packages-up-to my_robot_ros2/1` though that does require having one or more top-level packages that aggregate.

iory:
Thank you for the suggestion! That's indeed a valid approach for scenarios where you only want to build a specific subset of ROS2 packages using `colcon`.

However, the goal of this PR (`CATKIN_IGNORE_ROS2`) is slightly different and aims to simplify the **coexistence** of *both* build systems in a shared workspace, especially for users who actively switch between ROS1 and ROS2 development.

The `--packages-up-to` approach requires:
1.  Maintaining a top-level meta-package that aggregates all target ROS2 packages.
2.  Users to remember and type the specific option (`--packages-up-to [meta-package]`) every time they build with `colcon`.

This PR allows for a more "declarative" approach. By simply adding a `CATKIN_IGNORE_ROS2` file, `catkin_tools` explicitly ignores ROS2 packages.

The primary benefit is **simplicity for the end-user**:
* To build ROS1 packages: run `$ catkin build`
* To build ROS2 packages: run `$ colcon build`

No extra options or meta-packages are needed. This is particularly valuable in **educational settings** (like universities or labs) where students are learning both ROS1 and ROS2, often in the same environment. It lowers the barrier to entry and reduces confusion, allowing them to focus on the robotics concepts rather than complex build tool configurations.

